PR #94, opened 2023-04-18
feature: Multiple conditions

orius123:
## Related Issues

Fixes https://github.com/descope/etc/issues/2073
Fixes https://github.com/descope/etc/issues/2322

## Description

1. Send the version of the flow as we get it in the config.json to the start and next requests, to avoid cases where the tab was stale for a long time and when the end-user is resuming the flow has already been changed by the descoper
1. Support multiple client conditions in the first step and not just a single condition by iterating the conditions until first is met

## Must

- [x] Tests


aviadl (commented):
Looks good to me 
But will like @nirgur review here as well

nirgur (commented):
Looks good added a few comments

packages/web-component/src/lib/helpers/conditions.ts
```diff
@@ -1,6 +1,11 @@
-import { ClientCondition, ConditionsMap, Context } from '../types';
+import {
+  ClientCondition,
+  ClientConditionResult,
+  ConditionsMap,
+  Context,
+} from '../types';
 
-export const conditions: ConditionsMap<Context> = {
+export const conditionsMap: ConditionsMap<Context> = {
```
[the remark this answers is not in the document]
orius123: Just renamed, this is what we had already before
I think we thought that in the future we might have different kind of conditions

packages/web-component/src/lib/helpers/conditions.ts
```diff
@@ -9,14 +11,16 @@ export const conditions: ConditionsMap<Context> = {
     'is-true': (ctx) => !!ctx.code,
     'is-false': (ctx) => !ctx.code,
   },
+  [elseInteractionId]: {
+    // always true
+  },
```
nirgur, line 16: WDYT about returning a function like the other conditions?
I think it will be more consistent

nirgur (commented):
Looks much better 😄 
I think we can try to have the same convention for the else condition and it will make the code a bit more consistent
See my comments below and LMK what you think

packages/web-component/src/lib/helpers/conditions.ts
orius123: function of what operator?

nirgur: Without an operator
something like
```
[elseInteractionId]: () => true,
```

orius123: So we need to change the def of the conditionsMap

packages/web-component/src/lib/helpers/conditions.ts
```diff
@@ -26,3 +30,27 @@ export const calculateCondition = (
...
+    }
+    return checkFunc(ctx);
+  });
+  if (conditionResult) {
+    return {
+      startScreenId: conditionResult.met.screenId,
+      conditionInteractionId: conditionResult.met.interactionId,
+    };
+  }
+  return {};
```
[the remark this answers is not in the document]
orius123: this will override outside and I want it to not to if there is no result

nirgur (approved):

PR #96, opened 2023-04-19
fix: remove oauth redirect

asafshen:
## Description

remove oauth redirect as well


nirgur (approved):
Just for the protocol
I think some users are using this option and I think we should keep it

PR #104, opened 2023-04-24
feat: Add options to merge users in case of phone or email update

aviadl:
Also determine if to add as loginid or not

### DIFF
--- a/packages/core-js-sdk/src/sdk/magicLink/index.ts
+++ b/packages/core-js-sdk/src/sdk/magicLink/index.ts
@@ -85,12 +85,14 @@ const withMagicLink = (httpClient: HttpClient) => ({
         loginId: string,
         email: string,
         URI?: string,
-        token?: string
+        token?: string,
+        addToLoginIDs?: boolean,
+        onMergeUseExisting?: boolean
       ): Promise<SdkResponse<MaskedEmail>> =>
         transformResponse(
           httpClient.post(
             apiPaths.magicLink.update.email,
-            { loginId, email, URI },
+            { loginId, email, URI, addToLoginIDs, onMergeUseExisting },
             { token }
           )
         )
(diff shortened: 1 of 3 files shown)


packages/core-js-sdk/src/sdk/magicLink/index.ts
```diff
@@ -99,11 +101,18 @@ const withMagicLink = (httpClient: HttpClient) => ({
...
         [delivery]: withUpdatePhoneValidations(
-          (loginId: string, phone: string, URI?: string, token?: string) =>
+          (
+            loginId: string,
+            phone: string,
+            URI?: string,
+            token?: string,
+            addToLoginIDs?: boolean,
+            onMergeUseExisting?: boolean
+          ) =>
```
nirgur, line 111: We are casting it to `MagicLink[Routes.updatePhone]` so we need to update the type as well

In addition, it seems like we have several optional parameters, 
I would consider putting it in an object for easier usage
e.g. `...phone.sms(loginId, phone, {URI, token, addToLoginIDs, ...})`

In this way, in case we want to pass only the last param `onMergeUseExisting = true` we don't have to call the fn in this way 
`...phone.sms(loginId, phone, undefined, undefined, undefined, true)`

packages/core-js-sdk/src/sdk/otp/index.ts
```diff
@@ -115,11 +117,17 @@ const withOtp = (httpClient: HttpClient) => ({
       (acc, delivery) => ({
         ...acc,
         [delivery]: withUpdatePhoneValidations(
-          (loginId: string, phone: string, token?: string) =>
+          (
+            loginId: string,
+            phone: string,
+            token?: string,
+            addToLoginIDs?: boolean,
+            onMergeUseExisting?: boolean
```
nirgur, line 125: Same here, we should also the type

nirgur (commented):
Looks good 👍 
We need to update some types, 
And consider changing the fn API
See my comments below

Update: maybe we should also add a few tests 😄

PR #125, opened 2023-05-13
Fix: Check input type before submitting value

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/2682

## Description

We are using input with type `checkbox`, and the WC always sampled the input's `value`, which in `checkbox` type it should sample `checked`.

## Must

- [x] Tests (in progress)


asafshen (approved):
Well done 👍

PR #144, opened 2023-05-23
fix: Skip start optimization fro oidc flows

asafshen:
## Description
When OIDC flow is on - we don't want to apply the start optimization

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -14,6 +14,7 @@ import {
   isConditionalLoginSupported,
   replaceWithScreenState,
   setTOTPVariable,
+  showFirstScreenOnExecutionInit,
   State,
   withMemCache,
 } from '../helpers';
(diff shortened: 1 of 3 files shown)


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -249,7 +251,7 @@ class DescopeWc extends BaseDescopeWc {
     }
 
     // if there is no screen id (probably due to page refresh) we should get it from the server
-    if (!screenId && !startScreenId) {
+    if (!screenId && !startScreenId && !oidcIdpStateId) {
```
nirgur, line 254: Why do we want to change this condition?

nirgur (commented):
Looks good 💪 
Added one small comment

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: oh, its a leftover from previous change, removed 👍

PR #149, opened 2023-05-24
feat: Defer redirect when in background for Android

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/310

## Description

When running flows in native Android, the final redirect cannot happen in the background, like happens with Enchanted links. This PR adds a new `redirect auth` URL parameter to know who the initiator is, and defers the redirect specifically for the case of background and android.

## Must

- [X] Tests

### DIFF
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -52,7 +52,7 @@ class BaseDescopeWc extends HTMLElement {
 
   #init = false;
 
-  #flowState = new State<FlowState>();
+  #flowState = new State<FlowState>({ deferredRedirect: false } as FlowState);
 
   #debugState = new State<DebugState>();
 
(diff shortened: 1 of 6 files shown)


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -182,6 +183,16 @@ class BaseDescopeWc extends HTMLElement {
     this.#flowState.update({ stepId, executionId });
   }
 
+  #syncStateWithVisibility() {
+    if (!document.hidden) {
+      // Defer the update a bit, it won't work otherwise
+      setTimeout(() => {
+        // Trigger state update that will redirect and pending deferred redirection
+        this.#flowState.update({ deferredRedirect: false });
+      }, 300);
```
nirgur, line 192: I know we talked about it, and I don't have a better idea how we can solve it, 
but putting here an arbitrary delay is risky

nirgur (commented):
Looks good
my only concern is about the timeout
do you think we can come up with something creative that will be a bit more robust?

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
itaihanski: It's the best we have at the moment. I'm not aware of a different event we can listen too, I think it's Android OS related. In any case it's only for Android flows.

nirgur (approved):

PR #164, opened 2023-06-05
add optional token validations

asafshen:
## Related Issues
https://github.com/descope/etc/issues/2883

## Description
added optional string validations for top level functions that was missing

### DIFF
--- a/packages/core-js-sdk/src/sdk/index.ts
+++ b/packages/core-js-sdk/src/sdk/index.ts
@@ -16,10 +16,11 @@ import withSaml from './saml';
 import withTotp from './totp';
 import withPassword from './password';
 import { JWTResponse, UserResponse } from './types';
-import { stringNonEmpty, withValidations } from './validations';
+import { stringNonEmpty, withValidations, string } from './validations';
 import withWebauthn from './webauthn';
 
 const withJwtValidations = withValidations(stringNonEmpty('token'));
+const withOptionalTokenValidations = withValidations(string('token'));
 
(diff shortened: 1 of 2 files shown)


packages/core-js-sdk/src/sdk/validations/index.ts
```diff
@@ -19,7 +25,7 @@ export const withValidations =
   };
 
 export const string = (fieldName: string) => [
```
asafshen, line 27: noting that there is only once place that uses `string` which is the webauthn sign in - https://github.com/descope/core-js-sdk/pull/153/files

I prefer to modify this function to also allow `undefined` rather then create 3 types of validations - string, nonempty-string / string-or-undefined

packages/core-js-sdk/test/sdk/webauthn.test.ts
```diff
@@ -173,10 +173,6 @@ describe('webauthn', () => {
 
   describe('signIn', () => {
     describe('start', () => {
-      it('should throw an error when loginId is not a string', () => {
```
asafshen, line 176: refer to the this comment about why this test was removed - https://github.com/descope/descope-js/pull/164/files#r1218491667

packages/core-js-sdk/src/sdk/validations/index.ts
nirgur: What about having an `OR` validator, that will get a list of validators 
and allow us to check if at least one validator is passed?

nirgur (approved):
Well done 👍

PR #172, opened 2023-06-15
add getTenants utility function to fetch tenant list from JWT

chauhanshubham:
## Description

This commit adds a `getTenants` utility function to fetch the list of tenants from the JWT token.
The JWT token consists of tenants as keys, and in order to fetch roles/permissions for a specific tenant, 
we need to parse the tenant keys ourselves. 
The helper function provides an easier way to get the tenant list, which can then be further used
to fetch roles/permissions in a tenant enabled environment.

## Must

- [x] Tests

### DIFF
--- a/packages/core-js-sdk/src/sdk/helpers/index.ts
+++ b/packages/core-js-sdk/src/sdk/helpers/index.ts
@@ -33,6 +33,17 @@ export function isJwtExpired(token: string): boolean {
 }
 
 /**
+ * Returns the list of tenants in the given JWT
+ *
+ * @param token JWT token
+ */
+export function getTenants(token: string): string[] {
+  let claims: any = parseJwt(token);
+  const items = Object.keys(claims?.tenants);
+  return Array.isArray(items) ? items : [];
+}
+
+/**
  * Returns the list of permissions granted in the given JWT but DOES NOT check for signature
  *
  * @param token JWT token
--- a/packages/core-js-sdk/src/sdk/index.ts
+++ b/packages/core-js-sdk/src/sdk/index.ts
@@ -4,6 +4,7 @@ import withAccessKeys from './accesskey';
 import withEnchantedLink from './enchantedLink';
 import withFlow from './flow';
 import {
+  getTenants,
   getJwtPermissions,
   getJwtRoles,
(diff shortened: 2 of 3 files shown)


packages/core-js-sdk/src/sdk/helpers/index.ts
```diff
@@ -32,6 +32,17 @@ export function isJwtExpired(token: string): boolean {
...
 
+/**
+ * Returns the list of tenants in the given JWT
+ *
+ * @param token JWT token
+ */
+export function getTenants(token: string): string[] {
+  let claims: any = parseJwt(token);
+  const items = Object.keys(claims?.tenants);
+  return Array.isArray(items) ? items : [];
```
asafshen, line 43: this logic of extracting the items from the key is already implemented in `getJwtAuthorizationItems` so we can utilize it, so I suggest to reuse it

```suggestion
  return getJwtAuthorizationItems(token, '', 'tenants');
```

asafshen (commented):
looks great!

one comment

also mind that this repo follows [conventional commits](https://www.conventionalcommits.org), so commit message should start with `fix: `chore:`, `feat:`, etc...

packages/core-js-sdk/src/sdk/helpers/index.ts
chauhanshubham: `tenants` is not an array and we need keys in this case. So will have to do some handling in the `getJwtAuthorizationItems` function right?

chauhanshubham: could do something like
```
const items = claims?.[claim];
if (Array.isArray(items)) return items;
if (items instanceof Object) return Object.keys(items);
return []
```
in the `getJwtAuthorizationItems` function

asafshen: good point 👍 

thanks for the explantation

asafshen (approved):

asafshen:
@chauhanshubham Build has been fixed, thanks!

asafshen:
@chauhanshubham thanks for this neat contribition!

PR #173, opened 2023-06-15
feat: add prefer-biometrics option that defaults to true

shilgapira:
## Related Issues
Fixes https://github.com/descope/etc/issues/2880

## Description
- Added a new option called `prefer-biometrics` to the `web-component`
- This option is enabled by default but can be turned off with: `<descope-wc ... prefer-biometrics="false">`
- If the user is running a Chromium-based browser and it seems like they have a platform authenticator connected we'll limit passkeys to only allow using the platform authenticator
- The effect of this is simply that the user will immediately get a prompt to create a passkey rather than going through the multiple selection dialog (see the issue above)
- This has no effect in Firefox or Safari

## Must
- [X] Tests

### DIFF
--- a/packages/web-component/test/descope-wc.test.ts
+++ b/packages/web-component/test/descope-wc.test.ts
@@ -2055,7 +2055,7 @@ describe('web-component', () => {
       await waitFor(() => expect(startMock).toHaveBeenCalled());
 
       await waitFor(() => screen.findByShadowText('It works!'), {
-        timeout: 4000,
+        timeout: 6000,
       });
 
       fireEvent.click(screen.getByShadowText('click'));


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -155,6 +154,11 @@ class BaseDescopeWc extends HTMLElement {
     return res === 'true';
   }
 
+  get preferBiometrics(): boolean {
+    const res = this.getAttribute('prefer-biometrics') ?? 'true';
```
asafshen, line 158: we should probably add it to readme with all other properties 
image

asafshen (commented):
cool, don't forget to expose in react and vue sdks

PR #187, opened 2023-06-28
chore: Improve docs

asafshen:
## Related Issues
https://github.com/descope/etc/issues/3105

## Description
improve readme to note about `sessionTokenViaCookie`


chris4490 (approved):
Looks good to me. I don't think this effects the docs as the behavior is unchanged, so I'm good to go here.

PR #219, opened 2023-07-27
feat: Fetch token from url on start

aviadl:
This way we can have a flow that will start with verify token 
related to https://github.com/descope/etc/issues/2288

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -140,11 +140,14 @@ class DescopeWc extends BaseDescopeWc {
       // As an optimization - we want to show the first screen if it is possible
       if (!showFirstScreenOnExecutionInit(startScreenId, oidcIdpStateId)) {
         const inputs: Record<string, any> = {};
+        let exists = false;
         if (code) {
+          exists = true;
           inputs.exchangeCode = code;
           inputs.idpInitiated = true;
         }
         if (token) {
+          exists = true;
           inputs.token = token;
         }
         const sdkResp = await this.sdk.flow.start(
@@ -158,7 +161,7 @@ class DescopeWc extends BaseDescopeWc {
           },
           conditionInteractionId,
           '',
-          inputs,
+          exists ? inputs : undefined,
           flowConfig.version
         );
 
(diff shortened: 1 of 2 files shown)


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -139,12 +139,14 @@ class DescopeWc extends BaseDescopeWc {
...
-              exchangeCode: code,
-              idpInitiated: true,
-            }
-          : undefined;
+        const inputs = {};
+        if (code) {
+          inputs['exchangeCode'] = code;
+          inputs['idpInitiated'] = true;
+        }
+        if (token) {
```
orius123, line 147: Can you add a test?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -322,8 +324,8 @@ class DescopeWc extends BaseDescopeWc {
     // If there is a start screen id, next action should start the flow
     // But if oidcIdpStateId is not empty, this optimization doesn't happen
     if (showFirstScreenOnExecutionInit(startScreenId, oidcIdpStateId)) {
-      stepStateUpdate.next = (interactionId, inputs) =>
-        this.sdk.flow.start(
+      stepStateUpdate.next = (interactionId, inputs) => {
+        return this.sdk.flow.start(
```
orius123, line 328: In what case is it needed?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen, line 1: We also need to handle cleanup of the token in this case
image

need to examine the code above to see where it make sense to do it

asafshen (commented):
code looks good, we need to ensure cleanup is done as well

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: if we want the invite to the flow to start with a screen - we will need it

aviadl: When flow starts with a screen for some reason
But we still want to take the token from the URL

packages/web-component/src/lib/descope-wc/DescopeWc.ts
aviadl: @asafshen dont understand your comment

asafshen: currently, after we use token (e.i after it is sent to `update` request - we clean the token from the component state

we should do the same to not pass token again in next request

asafshen (dismissed):

asafshen (approved):

PR #230, opened 2023-08-04
feat: locale support

dorsha:
## Related Issues

https://github.com/descope/etc/issues/290

## Description

Add locale support.
First we try to get it from the provided locale, if not provided we will take it from navigator.language.
We check first if the locale is part of the config locales.
If not able to fetch the html with the locale we always fallback to the original html without locale.

## Must

- [x] Tests


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -309,6 +310,17 @@ class DescopeWc extends BaseDescopeWc {
...
+    const filename = `${startScreenId || screenId}.html`;
+
+    let filenameWithLocale: string;
+    const userLocale = (locale || navigator.language || '').toLowerCase(); // use provided locals, otherwise use browser locale
+    const targetLocales = await this.getTargetLocales();
+
+    if (targetLocales.includes(userLocale)) {
+      filenameWithLocale = `${startScreenId || screenId}-${userLocale}.html`;
+    }
+
```
nirgur, line 323: I think it will be better to extract it into a function

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -319,7 +331,10 @@ class DescopeWc extends BaseDescopeWc {
           name: this.sdk.getLastUserDisplayName() || loginId,
         },
       },
-      htmlUrl: getContentUrl(projectId, `${startScreenId || screenId}.html`),
+      htmlUrl: getContentUrl(projectId, filename),
+      htmlLocaleUrl: filenameWithLocale
+        ? getContentUrl(projectId, filenameWithLocale)
+        : undefined,
```
nirgur, line 337:
```suggestion
      htmlLocaleUrl: filenameWithLocale && getContentUrl(projectId, filenameWithLocale),
```

nirgur (commented):
Looks good 👍 

Since the web component code is already pretty complex, I'd suggest pulling out the locale logic into separate functions
That should make it a bit more readable

WDYT?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -504,11 +519,28 @@ class DescopeWc extends BaseDescopeWc {
...
+          `Failed to fetch html from ${htmlLocaleUrl}, error: ${ex}. Fallback to url ${htmlUrl}`
+        );
+      }
+    }
+
+    if (!stepTemplate.innerHTML) {
+      const { body } = await fetchContent(htmlUrl, 'text');
+      stepTemplate.innerHTML = body;
+    }
+
```
nirgur, line 543: I think it will be better to put it in a fn that returns the page content

nirgur (approved):

PR #231, opened 2023-08-06
feat: flow config locales usage

dorsha:
## Related Issues

https://github.com/descope/etc/issues/290

## Description

Flow config locales usage

## Must

- [x] Tests


nirgur (commented):
Looks good
But why did we change it from project level to flow level?
Can we have different flows which translated to different languages?

dorsha:
> Looks good But why did we change it from project level to flow level? Can we have different flows which translated to different languages?

It was always flow level, previous PR was a mistake.

nirgur (approved):

PR #236, opened 2023-08-10
feat: Add logger prop for runner

Bars92:
## Related Issues

related https://github.com/descope/etc/issues/2892

## Related PRs
react sdk - https://github.com/descope/react-sdk/pull/380
app - 

## Description
- add `onNext` event on web component to callback on next action including action details and state
- move logger to function to allow overloading
- add trycatch to screen html file
- add step name to flow responses

### DIFF
--- a/packages/web-component/README.md
+++ b/packages/web-component/README.md
@@ -101,6 +101,28 @@ const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
 descopeWcEle.errorTransformer = translateError;
 ```
 
+### `logger` - An object that defines how to log error, warning and info. Defaults to console.error, console.warn and console.info respectively
+
+Usage example:
+
+```javascript
+const log = {
+  info: (message: string, description: string, state: any) => {
+    console.log(message, description);
+  },
+  warn: (title: string, description: string) => {
+    console.warn(`WARN: ${title}`, description);
+  },
+  error: (title: string, description: string) => {
+    console.error(`ERROR: ${title}`, description);
+  },
+};
+
+const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
+
+descopeWcEle.logger = log;
+```
+
 ## Events
 
 ### `error` - Fired when an error occurs. The event detail contains the error object.
@@ -122,16 +144,5 @@ Usage example:
 const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
 descopeWcEle.addEventListener('success', (e) =>
   alert(`Success! - ${JSON.stringify(e.detail)}`)
-);
-```
-
-### `log` - Fired when the log is called. The event detail contains the message and details and a possible current state.
-
-Usage example:
-
-```javascript
-const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
-descopeWcEle.addEventListener('log', (e) =>
-  alert(`Log! - ${JSON.stringify(e.detail)}`)
 );
 ```
--- a/packages/web-component/README.md
+++ b/packages/web-component/README.md
@@ -106,7 +106,7 @@ descopeWcEle.errorTransformer = translateError;
 Usage example:
 
 ```javascript
-const log = {
+const logger = {
   info: (message: string, description: string, state: any) => {
     console.log(message, description);
   },
@@ -120,7 +120,7 @@ const log = {
 
 const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
 
-descopeWcEle.logger = log;
+descopeWcEle.logger = logger;
 ```
 
 ## Events
--- a/packages/web-component/README.md
+++ b/packages/web-component/README.md
@@ -109,7 +109,9 @@ Usage example:
 
 ```javascript
 const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
-descopeWcEle.addEventListener('error', (e) => alert(`Error! - ${e.detail.errorMessage}`));
+descopeWcEle.addEventListener('error', (e) =>
+  alert(`Error! - ${e.detail.errorMessage}`)
+);
 ```
 
 ### `success` - Fired when the flow is completed successfully. The event detail contains the flow result.
@@ -118,5 +120,18 @@ Usage example:
 
 ```javascript
 const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
-descopeWcEle.addEventListener('success', (e) => alert(`Success! - ${JSON.stringify(e.detail)}`));
+descopeWcEle.addEventListener('success', (e) =>
+  alert(`Success! - ${JSON.stringify(e.detail)}`)
+);
+```
+
+### `log` - Fired when the log is called. The event detail contains the message and details and a possible current state.
+
+Usage example:
+
+```javascript
+const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
+descopeWcEle.addEventListener('log', (e) =>
+  alert(`Log! - ${JSON.stringify(e.detail)}`)
+);
 ```
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -30,9 +30,7 @@ declare const BUILD_VERSION: string;
 
 // this base class is responsible for WC initialization
 class BaseDescopeWc extends HTMLElement {
-  logger: ILogger = {
-    ...console,
-  };
+  logger: ILogger = console;
 
   static get observedAttributes() {
     return [
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -265,7 +265,7 @@ class BaseDescopeWc extends HTMLElement {
         executionContext: { geo: headers['x-geo'] },
       };
     } catch (e) {
-      this.logger.error(
+      this.logger().error(
         'Cannot get config file',
         'make sure that your projectId & flowId are correct'
       );
@@ -293,7 +293,7 @@ class BaseDescopeWc extends HTMLElement {
       const { body } = await fetchContent(themeUrl, 'text');
       styleEle.innerText = body;
     } catch (e) {
-      this.logger.error(
+      this.logger().error(
         'Cannot fetch theme file',
         'make sure that your projectId & flowId are correct'
       );
@@ -363,17 +363,19 @@ class BaseDescopeWc extends HTMLElement {
     );
   }
 
-  logger = {
-    error: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.error(message, description, new Error());
-      this.#updateDebuggerMessages(message, description);
-    },
-    info: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.log(message, description);
-    },
-  };
+  logger() {
+    return {
+      error: (message: string, description = '') => {
+        // eslint-disable-next-line no-console
+        console.error(message, description, new Error());
+        this.#updateDebuggerMessages(message, description);
+      },
+      info: (message: string, description = '') => {
+        // eslint-disable-next-line no-console
+        console.log(message, description);
+      },
+    };
+  }
 
   #handleKeyPress() {
     // we want to simulate submit when the user presses Enter
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -19,6 +19,7 @@ import {
   DebugState,
   FlowState,
   FlowStateUpdateFn,
+  ILogger,
   SdkConfig,
   ThemeOptions,
 } from '../types';
@@ -29,6 +30,10 @@ declare const BUILD_VERSION: string;
 
 // this base class is responsible for WC initialization
 class BaseDescopeWc extends HTMLElement {
+  logger: ILogger = {
+    ...console,
+  };
+
   static get observedAttributes() {
     return [
       'project-id',
@@ -55,19 +60,16 @@ class BaseDescopeWc extends HTMLElement {
 
   #init = false;
 
-  #log = {
+  loggerWrapper = {
     error: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.error(message, description, new Error());
+      this.logger.error(message, description, new Error());
       this.#updateDebuggerMessages(message, description);
     },
     warn: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.warn(message, description);
+      this.logger.warn(message, description);
     },
-    info: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.log(message, description);
+    info: (message: string, description = '', state: any = {}) => {
+      this.logger.info(message, description, state);
     },
   };
 
@@ -281,7 +283,7 @@ class BaseDescopeWc extends HTMLElement {
         executionContext: { geo: headers['x-geo'] },
       };
     } catch (e) {
-      this.logger().error(
+      this.loggerWrapper.error(
         'Cannot get config file',
         'make sure that your projectId & flowId are correct'
       );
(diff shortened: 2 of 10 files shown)


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -95,14 +129,21 @@ class DescopeWc extends BaseDescopeWc {
...
       }
     }
 
-    const { body } = await fetchContent(htmlUrl, 'text');
-    return body;
+    try {
+      const { body } = await fetchContent(htmlUrl, 'text');
+      return body;
+    } catch (ex) {
+      this.logger().error(
```
asafshen: can you explain why this change is needed?

the log addition looks good,
but I think that throwing is better is more suitable here than returning null
it will signal caller that the action failde

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -721,6 +776,10 @@ class DescopeWc extends BaseDescopeWc {
     this.rootElement.classList.add('fade-out');
   }
 
+  #dispatchNext(m: MessageLog) {
+    this.#dispatch('next', m);
```
asafshen, line 779: please add an usage for it in the sample html file, and also add this to README

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -47,6 +49,34 @@ class DescopeWc extends BaseDescopeWc {
 
   flowState: State<FlowState>;
 
+  log = {
+    error: (message: string, description?: string) => {
+      super.logger().error(message, description);
+      this.#dispatchNext({
```
asafshen, line 55: I don't think that we should dispatch next on every log, and if so, I would communicate it differently 

lets if something happened during execution that is not a next response, it should not trigger the `onNext` callback

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -55,6 +55,22 @@ class BaseDescopeWc extends HTMLElement {
 
   #init = false;
 
+  #log = {
```
asafshen, line 58: I didn't understand the inheritance pattern where log is implemented both in `BaseDescopeWc` and in `DescopeWc`

why do we need it? 

I don't think that logger in the base should trigger `next` callback (aka dispatchNext)

might be simpler to make it as before

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -47,6 +49,34 @@ class DescopeWc extends BaseDescopeWc {
...
+        type: MessageLogType.ERROR,
+        title: message,
+        description,
+      });
+    },
+    warn: (message: string, description?: string) => {
+      super.logger().warn(message, description);
+      this.#dispatchNext({
```
asafshen, line 63: following latest comment, this means that descoper will get a next, without `state`

packages/web-component/src/lib/types.ts
```diff
@@ -80,6 +80,17 @@ export type DebuggerMessage = {
...
 
+export enum MessageLogType {
+  INFO = 'info',
+  ERROR = 'error',
+  WARN = 'warning',
+}
+
+export type MessageLog = {
+  type: MessageLogType;
+  state?: any;
```
asafshen, line 91: the term `state` doesn't convey a lot of information with it, we should find out a better naming for it?

asafshen (commented):
looks great!

@nirgur if you have a time to take a look it will be also nice :)

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: maybe i should change the event name? because the intention is to know about everything that the sdk is doing specially any error. its easier to call it all "next" for simplicity. currently, all logs are tied to an action and all logs i would like to send

packages/web-component/src/lib/types.ts
Bars92: i agree, however it could be anything and should represent the current "state" of the log. mostly add some details on what is the current state such as steps details and screen details. i would rather keep it vague so we can add whatever we want here

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: there are other logs occurrences that are not tied to the next:

image

so imo it is not easier/simpler, cuz this name affects the api we expose

I suggest we only trigger `next` when we should (aka after `next` response) or adjust the API to something more appropriate to log

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: since its an object i moved kept it a property and the logger became a function so we can override it in the DescopeWc

packages/web-component/src/lib/types.ts
```diff
@@ -80,6 +80,17 @@ export type DebuggerMessage = {
...
 };
 
+export enum MessageLogType {
+  INFO = 'info',
+  ERROR = 'error',
+  WARN = 'warning',
+}
+
+export type MessageLog = {
+  type: MessageLogType;
```
asafshen, line 90:
```suggestion
  level: MessageLogLevel;
```

asafshen (commented):
cool 👍

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -449,6 +493,16 @@ class DescopeWc extends BaseDescopeWc {
       });
       return;
     }
+
+    this.logger().info(`Step "${stepName || `#${stepId}`}" is ${status}`, '', {
```
asafshen, line 497: this will make the log a bit noisi, by default, to every usage

maybe worth to split debugger and the dispatch? imo those are related, but not the same use cases

wdyt?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: this is the only place that throw an error. i would expect this error to be part of the sdk as in the logs and same as the other errors to be passed as an error, otherwise we and other users need to get the error details from different places

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: i think this is currently very useful when running the flow. i would suggest we keep it and get used to it before we release. if we want before the release we can always remove the `console.log` or disable it

packages/web-component/src/lib/types.ts
```diff
@@ -80,6 +80,17 @@ export type DebuggerMessage = {
   description?: string;
 };
 
+export enum MessageLogType {
+  INFO = 'info',
+  ERROR = 'error',
+  WARN = 'warning',
+}
+
+export type MessageLog = {
```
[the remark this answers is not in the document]
Bars92: thats a nice idea but its currently not used and probably wont be maintained so we should add this once we know how and why

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: how could you disable it?

I think it is useful if you are a Descoper that is debugging Descope flow, but if you are not, it is just noise, you may be a developer that is not interested in the login process, or something like this? 

I think that those two capabilities (a. debugging flow to subscribe to `onNext` and b. internal logging for troubleshooting purposes) are not tightly coupled

Bars92: i dont think we can easily assume it will cause noise and/or hard to work with.. this `info` logs that are normally filtered will only be displayed when actually running the flow and not developing your application. If we find this is too much noise we can address it in a way that fit the actual customer needs.

This pr only refers to `onLog`, the `onNext` is discarded and will be added at later stages to track the actual actions and screens

packages/web-component/test/descope-wc.test.ts
```diff
@@ -216,6 +216,24 @@ describe('web-component', () => {
...
+    );
+
+    document.body.innerHTML = `<h1>Custom element test</h1> <descope-wc flow-id="otpSignInEmail" project-id="1"></descope-wc>`;
+
+    const wcEle = document.getElementsByTagName('descope-wc')[0];
+
+    const onLog = jest.fn();
+    wcEle.addEventListener('log', onLog);
+
+    await waitFor(() => expect(onLog).toHaveBeenCalled(), { timeout: 1000 });
```
asafshen, line 232: please also ensure the argument of `onLog` (aka `toHaveBeenCalledWith`)

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -363,17 +379,9 @@ class BaseDescopeWc extends HTMLElement {
...
-      // eslint-disable-next-line no-console
-      console.error(message, description, new Error());
-      this.#updateDebuggerMessages(message, description);
-    },
-    info: (message: string, description = '') => {
-      // eslint-disable-next-line no-console
-      console.log(message, description);
-    },
-  };
+  logger() {
```
nirgur, line 382: Why do we need to make it a function?
We can override it also when it's a property, no?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: than I somewhat I disagree about the need of `onLog`, I think that a developer would like to get whenever `next` happens and not when `log` happens, so I think those two topics should not be tightly coupled like they are now

>  this info logs that are normally filtered will only be displayed when actually running the flow and not developing your application

I don't think the average developer does this, but I argued enough, approving

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -47,6 +49,34 @@ class DescopeWc extends BaseDescopeWc {
 
   flowState: State<FlowState>;
 
+  log = {
+    error: (message: string, description?: string) => {
+      super.logger().error(message, description);
+      this.#dispatchLog({
```
nirgur, line 55: Do we think the descopers will use this event or it's only for internal use?

asafshen (approved):
👍

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -411,23 +449,21 @@ class DescopeWc extends BaseDescopeWc {
...
-    if (errorText) {
-      this.logger.error(errorText);
-    }
-
     if (sdkResp.data?.error) {
-      this.logger.error(
+      this.logger().error(
         `[${sdkResp.data.error.code}]: ${sdkResp.data.error.description}`,
-        sdkResp.data.error.message
+        `${errorText ? `${errorText} - ` : ''}${sdkResp.data.error.message}`
```
nirgur, line 463: Why do we need to "build" the error string here and we cannot use the existing pattern?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur, line 497: I think we should better understand the usecase here, is it only for internal use or it will be relevant for the descoper, maybe this should be enabled only when in debug mode

In addition, I don't think it's a good practice to build the error string here, let's use a function for it

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -95,14 +129,20 @@ class DescopeWc extends BaseDescopeWc {
...
       }
     }
 
-    const { body } = await fetchContent(htmlUrl, 'text');
-    return body;
+    try {
+      const { body } = await fetchContent(htmlUrl, 'text');
+      return body;
+    } catch (ex) {
+      this.logger().error(`Failed to fetch html page from ${htmlUrl}`, ex);
```
nirgur, line 143: Isn't it redundant?
We are already logging an error on the fetchContent fn

nirgur (commented):
Looks good 👍 
I added some comments

I think we should have a better understanding of the `log` event use cases,
Is it something that the descopers might use (not sure how exactly)
Or it's only for internal use 
if the only use case we can think of is the console debugger, we might think of another implementation that will not "pollute" the API

WDYT?
We can also talk about it tomorrow F2F if you like

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: it is mainly to be able to call base function for the debugger and console

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: yes, this is currently the only option for descopers to know what happening in the sdk, this should be useful.

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: the screen error is a good link between the actual error and the user error. in order to connect between them we concat them instead of producing 2 errors that mean the same.

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: not sure i folllow @nirgur, this will be used by descopers (also using console.log), this is not an error but a "debug" message to callback with the current flow state. The problem with debug mode is that it is tidily coupled with the debugger while to disable/enable only the debugger window is breaking changes. 


We should be able to let descopers decide how to consume the logger, currently we will provide console and log event.

packages/web-component/src/lib/descope-wc/DescopeWc.ts
Bars92: we are not logging, we are throwing. this will log it with context, similar to other fetches.

packages/web-component/README.md
```diff
@@ -101,6 +101,28 @@ const descopeWcEle = document.getElementsByTagName('descope-wc')[0];
 descopeWcEle.errorTransformer = translateError;
 ```
 
+### `logger` - An object that defines how to log error, warning and info. Defaults to console.error, console.warn and console.info respectively
+
+Usage example:
+
+```javascript
+const log = {
```
asafshen, line 109:
```suggestion
const logger = {
```

(And also in line 123)

packages/web-component/src/lib/types.ts
```diff
@@ -125,3 +125,9 @@ export interface Context {
   loginId?: string;
   code?: string;
 }
+
+export interface ILogger {
+  info(title: string, description: string, state: any): void;
+  warn(title: string, description?: string): void;
+  error(title: string, description?: string, params?: any): void;
```
asafshen, line 132: I don't think this is the proper interface, it should be `info(message?: any, ...optionalParams: any[])` , like the console interface (also for warn/error)

maybe even use the `Parameters<console.info>` or  have the entire `ILogger` pick the info,warn and error from `console`

asafshen (commented):
looks good! see small comments about the types

packages/web-component/src/lib/types.ts
Bars92: i do want to differentiate between our logger (that has state with possible types we can add for example) rather then console.info(params) we are actually never used by us. The console still implements the same method signatures

Bars92: i did change the error signature to better match

asafshen: I don't think that this differentiation helps, it will just add redundant constrain to the developer which is not needed
I would also do this for the other errors if possible

asafshen (approved):
👍 

make sure to also push this to Vue in addition to React

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -55,6 +58,19 @@ class BaseDescopeWc extends HTMLElement {
 
   #init = false;
 
+  loggerWrapper = {
+    error: (message: string, description = '') => {
+      this.logger.error(message, description, new Error());
```
nirgur, line 63: Now when we are getting the logger, some fns might not be implemented
I suggest to trigger it only if it exists, e.g.
```
this.logger.error?.(...)
```

PR #249, opened 2023-08-22
fix: Add polling execution flag

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/3844

## Description
a proposal for fixing polling issue - when the action of polling takes more than the interval execution 

there are other, maybe more appropriate ways to fix that, which is changing the mechanism to timeout based rather then interval based, but I think that this is the safest

image

image

@nirgur lmk what you think, and I will continue with the testing


asafshen:
closing in favor of https://github.com/descope/descope-js/pull/257

PR #251, opened 2023-08-23
Components context

orius123:
Related https://github.com/descope/etc/issues/3733

### DIFF
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -82,7 +82,9 @@ class BaseDescopeWc extends HTMLElement {
 
   #debugState = new State<DebugState>();
 
-  componentsContext = {};
+  #componentsContext = {};
+
+  getComponentsContext = () => this.#componentsContext;
 
   nextRequestStatus = new State<{ isLoading: boolean }>({ isLoading: false });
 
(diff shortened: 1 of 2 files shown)


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -82,6 +82,8 @@ class BaseDescopeWc extends HTMLElement {
 
   #debugState = new State<DebugState>();
 
+  componentsContext = {};
```
nirgur, line 85: I think we should making it private by adding a hash and add a getter fn
WDYT?

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -528,6 +535,11 @@ class BaseDescopeWc extends HTMLElement {
       // we want to update the state when user clicks on back in the browser
       window.addEventListener('popstate', this.#eventsCbRefs.popstate);
 
+      window.addEventListener(
```
nirgur, line 538: Worth adding a comment about the mixin that dispatch this event

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -732,10 +732,12 @@ class DescopeWc extends BaseDescopeWc {
 
       const formData = await this.#getFormData();
       const eleDescopeAttrs = getElementDescopeAttributes(submitter);
+      const contextArgs = this.componentsContext;
 
       const actionArgs = {
         ...eleDescopeAttrs,
         ...formData,
+        ...contextArgs,
```
nirgur, line 740: We need to think what is the correct order here

nirgur (commented):
Looks good,
Added a few small comments

orius123:
@nirgur I fixed you comments, lmk if you want to think about the order thing together

nirgur:
Let's wait with it until I fix all tests
Hopefully it will be today

orius123:
@nirgur can we merge this one?

nirgur:
> @nirgur can we merge this one?

👍

PR #255, opened 2023-08-25
Pass components v2 in flow calls

orius123:
Fixes https://github.com/descope/etc/issues/3680

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -151,6 +151,7 @@ class DescopeWc extends BaseDescopeWc {
     let conditionInteractionId: string;
     const loginId = this.sdk.getLastUserLoginId();
     const flowConfig = await this.getFlowConfig();
+    const projectConfig = await this.getProjectConfig();
 
     const redirectAuth =
       redirectAuthCallbackUrl && redirectAuthCodeChallenge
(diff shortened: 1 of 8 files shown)


orius123:
Really hard to know the state when all tests fail like that

packages/web-component/src/lib/types.ts
```diff
@@ -135,3 +131,35 @@ export interface ILogger {
...
+
+export type FlowConfig = {
+  startScreenId?: string;
+  version: number;
+  targetLocales?: string[];
+  conditions?: ClientCondition[];
+  condition?: ClientCondition;
+};
+
+export interface Configuration {
```
asafshen, line 156:
```suggestion
export interface ProjectFlowsConfiguration {
```

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -656,8 +674,15 @@ class DescopeWc extends BaseDescopeWc {
         `[${ELEMENT_TYPE_ATTRIBUTE}="polling"]`
       );
       if (loader) {
+        const flowConfig = await this.getFlowConfig();
+        const projectConfig = await this.getProjectConfig();
```
asafshen, line 678: I would not let the caller set this here

instead, we can pass it where the function is defined 
(also true for the [change bellow](https://github.com/descope/descope-js/pull/255/files#diff-4e98988585b972daa77cdd3c14af366d7021a2ef1666a77d6b95f50fc68e60c9R768)
image)

asafshen (commented):
looks good!

noting that maybe, instead passing version and componentVersion everywhere that we want to call `start/`next`, we can set it once, so the callers won't have to fetch it every time 

this requires a bit of design, but I think it will worth it

packages/web-component/src/lib/descope-wc/DescopeWc.ts
orius123: Talked to @nirgur, we prefer to do the types changes after fixing all the tests in the v2 branch

orius123:
Waiting for tests and will merge after

PR #258, opened 2023-08-28
fix tests

nirgur:
fixes: https://github.com/descope/etc/issues/3900

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request]
+on: [pull_request]
 env:
   NODE_VERSION: 18.2
   PNPM_VERSION: 7.28.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [pull_request]
+on: [push, pull_request]
 env:
   NODE_VERSION: 18.2
   PNPM_VERSION: 7.28.0
--- a/packages/web-component/jest.config.js
+++ b/packages/web-component/jest.config.js
@@ -9,10 +9,10 @@ module.exports = {
   collectCoverageFrom: ['src/lib/**/*.ts'],
   coverageThreshold: {
     global: {
-      branches: 84,
+      branches: 82,
       functions: 91,
-      lines: 95,
-      statements: 95,
+      lines: 94,
+      statements: 94,
     },
   },
   globals: {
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request]
```
orius123, line 2: Y?

packages/web-component/jest.config.js
```diff
@@ -9,10 +9,10 @@ module.exports = {
   collectCoverageFrom: ['src/lib/**/*.ts'],
   coverageThreshold: {
     global: {
-      branches: 84,
+      branches: 82,
       functions: 91,
-      lines: 95,
-      statements: 95,
+      lines: 94,
```
orius123, line 14: Let's try to get it to a 100, so it will be easier to see what are we missing

packages/web-component/test/helpers/templates.test.ts
```diff
@@ -12,26 +12,22 @@ describe('templates', () => {
   it('should handle descope inputs', async () => {
     document.body.innerHTML = `<div>
 			<input class="descope-input" name="email">
-			<textarea class="descope-input" name="description">
```
orius123, line 15: Y?
Also down' we want to test the new components?

packages/web-component/test/descope-wc.test.ts
```diff
@@ -143,13 +144,22 @@ describe('web-component', () => {
...
     (createSdk as jest.Mock).mockReturnValue(sdk);
+
+    const origAppend = document.body.append;
+    const spyAppend = jest.spyOn(document.body, 'append');
+    spyAppend.mockImplementation((ele: any) => {
+      setTimeout(() => {
+        ele.onload?.();
+      });
+      origAppend.bind(document.body)(ele);
+    });
```
orius123, line 155: you can extract this block to a helper function

orius123 (commented):
Very nice!
Small comments

.github/workflows/ci.yml
nirgur: There is no need to run the CI on every push
Currently it's running twice for every PR

packages/web-component/jest.config.js
nirgur: The scope of this task is to fix the existing tests,
Let's do it on another PR

packages/web-component/test/helpers/templates.test.ts
nirgur: It's testing the same thing so I removed it
Basically the code does not refer to specific components so it should be fine

packages/web-component/test/descope-wc.test.ts
nirgur: Done

.github/workflows/ci.yml
nirgur: Reverted

orius123 (approved):

PR #260, opened 2023-08-30
feat: Support sso-apps and saml-idp

guyp-descope:
Add support for sso appllications and saml idp flags, also handle saml idp response


guyp-descope:
@orius123 @nirgur please review

packages/web-component/test/descope-wc.test.ts
```diff
@@ -131,7 +134,8 @@ describe('web-component', () => {
   });
 
   afterEach(() => {
-    document.getElementsByTagName('html')[0].innerHTML = '';
+    document.getElementsByTagName('head')[0].innerHTML = '';
```
tomerlichtash, line 137: @nirgur We made this change as it seems that our test was contaminated from previous tests. Since all tests populate `body` and `head` it seems right to reset it this way. WDYT?

packages/web-js-sdk/src/sdk/flow.ts
```diff
@@ -4,7 +4,13 @@ import { isSupported } from './webauthn';
 type CoreSdkFlowStartArgs = Parameters<CoreSdk['flow']['start']>;
 type Options = Pick<
   CoreSdkFlowStartArgs[1],
-  'tenant' | 'redirectUrl' | 'redirectAuth' | 'oidcIdpStateId'
+  | 'tenant'
+  | 'redirectUrl'
+  | 'redirectAuth'
+  | 'oidcIdpStateId'
+  | 'samlIdpStateId'
+  | 'samlIdpUsername'
```
orius123, line 12: I don't see this prop in the server PR https://github.com/descope/orchestrationservice/pull/1416/files#diff-ca68e226af5f6291b8d818dca0420419518228c1f47965633c0ad81ce3d1ca4dR172

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -236,6 +251,39 @@ class DescopeWc extends BaseDescopeWc {
...
+        const script = document.createElement('script');
+        if (originalScript.src) {
+          script.src = originalScript.src;
+        } else {
+          script.textContent = originalScript.textContent;
+        }
+        document.body.appendChild(script);
+      });
+      return;
+    }
```
orius123, line 285: Please extract all of this into a utility function

orius123: @nirgur please review this part thoroughly, need to make sure it will work in v2 as well

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -330,10 +375,21 @@ export const getChromiumVersion = (
...
+// - If there is a samlIdpUsername - we can't skip this call because the sdk may
+// - If there is a ssoAppId - we can't skip this call because the sdk may
 export const showFirstScreenOnExecutionInit = (
   startScreenId: string,
-  oidcIdpStateId: string
-): boolean => startScreenId && !oidcIdpStateId;
+  oidcIdpStateId: string,
+  samlIdpStateId: string,
+  samlIdpUsername: string,
+  ssoAppId: string
```
orius123, line 386: Do we want want to make the server call if we have one of these (`samlIdpUsername` || `ssoAppId`) and no `samlIdpStateId`

packages/web-js-sdk/src/sdk/flow.ts
guyp-descope: its getting to the flow by redirecting from onetime service.. thats why..

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: its theoretic question/scenario as we will never have samlIdpUsername without have samlIdpStateId (or ssoAppId)..
Lets keep it simple and perform this call to the server for any change by any of those params (as we dont really have an extra calls here..)

packages/web-component/src/lib/descope-wc/DescopeWc.ts
guyp-descope: Extracted to function `injectSamlIdpForm`

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -365,20 +395,32 @@ class DescopeWc extends BaseDescopeWc {
...
         filenameWithLocale && getContentUrl(projectId, filenameWithLocale),
+      samlIdpUsername,
     };
 
     const lastAuth = getLastAuth(loginId);
 
     // If there is a start screen id, next action should start the flow
-    // But if oidcIdpStateId is not empty, this optimization doesn't happen
-    if (showFirstScreenOnExecutionInit(startScreenId, oidcIdpStateId)) {
+    // But if oidcIdpStateId, samlIdpStateId, ssoAppId is not empty, this optimization doesn't happen
```
nirgur, line 404: Can we add the reason for skipping this optimization so it will be clear for the next developer that's working on this area?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -365,20 +395,32 @@ class DescopeWc extends BaseDescopeWc {
...
+    if (
+      showFirstScreenOnExecutionInit(
```
nirgur, line 406: I see that we are calculating this condition twice, 
can we store it in a variable instead of calling it twice?

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -242,6 +269,21 @@ export const handleUrlParams = () => {
...
+  const samlIdpUsername = getSAMLIDPUsernameParamFromUrl();
+  if (samlIdpStateId) {
+    clearSAMLIDPUsernameParamFromUrl();
+  }
+
+  const ssoAppId = getSSOAppIdParamFromUrl();
+  if (ssoAppId) {
+    clearSSOAppIdParamFromUrl();
+  }
+
```
nirgur, line 286: These vars will be set when the module loads
I'm not sure that this is what we want here,
Is it intentional?

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -330,10 +375,21 @@ export const getChromiumVersion = (
 // As an optimization - We can show first screen if we have startScreenId and we don't have oidcIdpStateId
 // - If there startScreenId it means that the sdk can show the first screen and we don't need to wait for the sdk to return the first screen
 // - If there is a oidcIdpStateId - we can't skip this call because the sdk may
+// - If there is a samlIdpStateId - we can't skip this call because the sdk may
+// - If there is a samlIdpUsername - we can't skip this call because the sdk may
+// - If there is a ssoAppId - we can't skip this call because the sdk may
```
nirgur, line 380: It's good that you added these comments, 
let's just make sure to complete the sentence :)

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -330,10 +375,21 @@ export const getChromiumVersion = (
...
   startScreenId: string,
-  oidcIdpStateId: string
-): boolean => startScreenId && !oidcIdpStateId;
+  oidcIdpStateId: string,
+  samlIdpStateId: string,
+  samlIdpUsername: string,
+  ssoAppId: string
+): boolean =>
+  startScreenId &&
+  !oidcIdpStateId &&
```
nirgur, line 389: Is there a scenario where only some of these params has a value?
Do we want to skip the optimization if only some has a value? or we should have it all in order for this feature to work?

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -360,3 +416,25 @@ export const getInputValueByType = (input: HTMLInputElement): Promise<any> =>
...
+  // create script tags from response scripts
+  const htmlScripts = document.createElement('div');
+  htmlScripts.innerHTML = samlIdpFormResponse;
+  const scripts = Array.from(htmlScripts.getElementsByTagName('script'));
+  scripts.forEach((originalScript) => {
+    const script = document.createElement('script');
+    if (originalScript.src) {
+      script.src = originalScript.src;
+    } else {
+      script.textContent = originalScript.textContent;
```
nirgur, line 436: We might have issues with the [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) here 
@omercnet WDYT?

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -360,3 +416,25 @@ export const getInputValueByType = (input: HTMLInputElement): Promise<any> =>
       }
     }
   });
+
+export const injectSamlIdpForm = (samlIdpFormResponse: string) => {
+  // create form tag from response string
+  const htmlFormTemp = document.createElement('div');
```
nirgur, line 422: I think that using `template` is a better choice in this case

packages/web-component/test/descope-wc.test.ts
nirgur: I think it's ok
we can use `document.body` & `document.head` instead

nirgur (commented):
Looks good,
Added a few comments

packages/web-component/src/lib/helpers/helpers.ts
omercnet: what's going on here?

nirgur: We are generating a script tag and setting its content to a string we got from the server

omercnet: During runtime in the client?! Very sus. 
Can't the script itself conditionally load the right code ?

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: I cant say its intentional per se as Im just follow others code that doing the exact same thing :) 
I did it the same as getRedirectAuthFromUrl, getOIDCIDPParamFromUrl (that we pushed together in the oidc PR),
I prefer to keep it the same (for maintain the same logic) unless its problematic somehow ofc..

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: yea, for example when oidcIdpStateId has a value samlIdpStateId will not have a value (for sure as its another scenarios)

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: Adjusted and Fixed :)

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: Thanks to @tomerlichtash we make the logic here more clear as well.

packages/web-component/src/lib/helpers/helpers.ts
guyp-descope: finally we are generating the form inside the Webcomponent (and getting the url, samlResponse, relayState from orchestration service).

packages/web-component/src/lib/descope-wc/DescopeWc.ts
guyp-descope: Added. in general it because there are scenarios that we may decide not to show the first screen (in cases like a user is already logged in) mostly in SSO scenarios.

packages/web-component/src/lib/descope-wc/DescopeWc.ts
guyp-descope: I think this double calculation is required as by doing a quick test the optimization you suggest causes tests to fail, so I dont wanna go into debug all this in this pr.

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -241,6 +260,29 @@ class DescopeWc extends BaseDescopeWc {
...
     }
 
+    const samlProps = [
+      'samlIdpResponseUrl',
+      'samlIdpResponseSamlResponse',
+      'samlIdpResponseRelayState',
+    ];
+    if (
+      action === RESPONSE_ACTIONS.loadForm &&
+      samlProps.map((samlProp) => isChanged(samlProp)).includes(true)
```
nirgur, line 270:
```suggestion
      samlProps.some(samlProp => isChanged(samlProp))
```

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -595,6 +654,17 @@ class DescopeWc extends BaseDescopeWc {
...
+    if (
+      currentState.samlIdpUsername &&
+      !screenState.form?.loginId &&
+      !screenState.form?.email
+    ) {
+      screenState.form = {
+        loginId: currentState.samlIdpUsername,
+        email: currentState.samlIdpUsername,
+      };
+    }
```
nirgur, line 666: Can we do this instead?

```
screenState.form.loginId ??= currentState.samlIdpUsername;
screenState.form.email ??= currentState.samlIdpUsername;

```

guyp-descope: @orius123 ??
(this is the code that set the username we get from Azure)

orius123: This is for the form we send immediately, if so or doesn't matter, if it is for actual screens we need to append like Nor said

guyp-descope: ok cool so Ill push Nir suggestion..

guyp-descope: @orius123 but are we ok with replacing all this code block with Nirs two lines? as I think we still want to keep the "if" inplace.. am I right?

orius123: yes

nirgur (approved):

PR #261, opened 2023-08-31
feat: add readToken to only read tokens but not persist

Bars92:
## Related Issues

related https://github.com/descope/etc/issues/3871

## Related PRs
https://github.com/descope/console-app/pull/1804

## Description
- add readTokens to all only read of auth tokens instead of persist

## Must

- [ ] Tests

### DIFF
--- a/packages/web-js-sdk/README.md
+++ b/packages/web-js-sdk/README.md
@@ -24,6 +24,8 @@ const sdk = descopeSdk({
   sdk.refresh(...), flow.next(...), etc.) in browser storage. In addition, this will
   make `sdk.getSessionToken()` available, see usage bellow bellow */
   persistTokens: true,
+  /* Read Tokens that exist after a succesful authentication, this does not persist incoming tokens and refreshes in browser storage. */
+  readTokens: true,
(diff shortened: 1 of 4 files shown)


packages/web-js-sdk/README.md
```diff
@@ -24,6 +24,8 @@ const sdk = descopeSdk({
   sdk.refresh(...), flow.next(...), etc.) in browser storage. In addition, this will
   make `sdk.getSessionToken()` available, see usage bellow bellow */
   persistTokens: true,
+  /* Read Tokens that exist after a succesful authentication, this does not persist incoming tokens and refreshes in browser storage. */
+  readTokens: true,
```
asafshen, line 28: `persistTokens` and `readTokens` capabilities overlap, so this api will be tough to explain and reason about

if we are going with this solution, I think that this should be replaced with "enum" string (we can keep `persistTokens` for BC ofc)

asafshen (commented):
cool,
I still didn't understand how this is going to be used TBH, 


maybe worth to have an internal discussion so we can conclude the DoD of this

PR #265, opened 2023-08-31
feat: add preview and storage prefix

Bars92:
## Related Issues

related https://github.com/descope/etc/issues/3871

## Related PRs
https://github.com/descope/descope-js/pull/261

## Description
add preview flag and start option = `preview` is a web-component parameter that is sent as a startOption.
add storage prefix optional flag = `storage-prefix` is a web-component parameter that defines local storage prefix.
fix tests

## Must

- [ ] Tests


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -180,6 +182,14 @@ class BaseDescopeWc extends HTMLElement {
     return res === 'true';
   }
 
+  get storagePrefix() {
+    return this.getAttribute('storage-prefix') || '';
+  }
+
+  get preview() {
+    return !!this.getAttribute('preview');
+  }
```
asafshen, line 191: I'm not sure that those attributes should be exposed as props / inputs of the web-sdk.

maybe we should consider taking them from local storage, so they will not be as a part of the sdk/web-component inputs?

(I'm not sure they both will be needed by customers, and if they do - we can easily expose

@Bars92 @nirgur lmk what you think

packages/web-js-sdk/src/enhancers/withPersistTokens/index.ts
```diff
@@ -40,31 +41,45 @@ export const withPersistTokens =
...
 
-const wrapper: SdkFnWrapper<{}> =
+const wrapper =
+  (prefix?: string): SdkFnWrapper<{}> =>
   (fn) =>
   async (...args) => {
     const resp = await fn(...args);
 
-    clearTokens();
+    clearTokens(prefix);
```
asafshen, line 82: not critical, but why we need to create another scope here for the prefix and pass it here?
image

we can just use `storagePrefix` here

packages/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -34,40 +34,46 @@ function setJwtTokenCookie(
...
-      : setLocalStorage(SESSION_TOKEN_KEY, sessionJwt);
+      : setLocalStorage(`${storagePrefix}${SESSION_TOKEN_KEY}`, sessionJwt);
   }
 };
 
 /** Return the refresh token from the localStorage. Not for production usage because refresh token will not be saved in localStorage. */
-export function getRefreshToken() {
-  return getLocalStorage(REFRESH_TOKEN_KEY) || '';
+export function getRefreshToken(prefix: string = '') {
+  return getLocalStorage(`${prefix}${REFRESH_TOKEN_KEY}`) || '';
```
asafshen, line 54: if we decide not to expose prefix as an argument -  we can simply use the prefix value local storage here, rather then pass it down the chain

packages/web-component/README.md
```diff
@@ -71,14 +71,16 @@ NOTE: This package is a part of a monorepo. so if you make changes in a dependen
...
+| base-url       | Custom Descope base URL                                                                                                                                                                                                                   | **""**        |
+| theme          | **"light"** - Light theme</br>**"dark"** - Dark theme</br>**"os"** - Auto select a theme based on the OS theme settings                                                                                                                   | **"light"**   |
+| debug          | **"true"** - Enable debugger</br>**"false"** - Disable debugger                                                                                                                                                                           | **"false"**   |
+| preview        | **"true"** - Ignore incoming tokens</br>**"false"** - Do not ignore tokens                                                                                                                                                                | **"false"**   |
```
asafshen, line 79: if we decide to expose this, I think that this description is a bit inaccurate, I would say something vague like `Run flow in preview mode`, the flow engine will decide the impact of this

packages/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -34,40 +34,46 @@ function setJwtTokenCookie(
...
-  refreshJwt && setLocalStorage(REFRESH_TOKEN_KEY, refreshJwt);
+  refreshJwt &&
+    setLocalStorage(`${storagePrefix}${REFRESH_TOKEN_KEY}`, refreshJwt);
 
   // persist session token
   if (sessionJwt) {
     sessionTokenViaCookie
       ? setJwtTokenCookie(SESSION_TOKEN_KEY, sessionJwt, cookieParams)
-      : setLocalStorage(SESSION_TOKEN_KEY, sessionJwt);
+      : setLocalStorage(`${storagePrefix}${SESSION_TOKEN_KEY}`, sessionJwt);
```
asafshen, line 48: we should also handle the cookie value here in the same manner, this solution should not be narrowed only for local storage

in general I think that introducing such different behavior for cookie and local storage is something we should avoid

asafshen (commented):
looks great, minor comments, especially about rather those props should be applied as an API

worth getting @nirgur thoughts on this manner

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: i prefer keep it straight forward and not invent any more mechanisms to confuse users. Those props can come in handy, and a local storage is far too dangerous to apply and maintain regardless..

packages/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
Bars92: if we change the cookies nothing will work. mind this is a storage prefix, while cookies should be managed somewhere else including cookies per tenant etc, we cant change the cookie name

packages/web-js-sdk/src/enhancers/withPersistTokens/index.ts
Bars92: how? the `wrapper` is an external function. i can declare it inside the component and then i dont need the extra scope

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
asafshen: I don't think this is for users at the moment, this is currently for internal usage (similar to `content .base.url` local storage)

and making the API external will confuse the users more because it will increase the API surface of an already long API options

I don't think local storage is dangerous by any mean

though it will require more work on the caller, but at the moment - this tradeoff is much better - to simplify the SDK that is used by users/customers, and to internalize the complexity

packages/web-js-sdk/src/enhancers/withPersistTokens/index.ts
asafshen: disregard, I thought I deleted this comment before review 😅

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: The prefix is a good client side solution for multiple flows, if it will be saved in local storage it has to be prefixed with the project and flow, and cleaned whenever used to avoid effecting other flows.. this is way more hard to use and explain, also this will require us to break changes to convert to a prop because the prop is strict by nature. 

Btw the base content local storage was a bad idea that was hard to maintain, this is why it’s mostly automated now

Bars92: we really need a resolution here. @nirgur your opinion is highly desired

packages/web-component/README.md
Bars92: good idea, probably best to keep it vague for now

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
orius123: Maybe preview is for internal usage but prefix (by project ID) is something we already bumped into in the console-app ourself, I think this is something we should expose or have project ID as the default prefix

PR #270, opened 2023-09-03
Prevent button submission on `Enter` when loading/disabled

tomerlichtash:
Fixes https://github.com/descope/etc/issues/3992

### DIFF
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -438,6 +438,12 @@ class BaseDescopeWc extends HTMLElement {
 
       // in case there is a single button on the page, click on it
       if (buttons.length === 1) {
+        if (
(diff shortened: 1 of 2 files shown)


dorsha (approved):

nirgur:
I think it will be better to handle it in the `descope-button` component 
and not as part of the `web-component` lib

WDYT?

tomerlichtash:
> I think it will be better to handle it in the `descope-button` component
> and not as part of the `web-component` lib

Agreed. Do you want to leave the implementation on the `web-component` package as is, though?

tomerlichtash:
@nirgur Added clickable mixin to handle this via the button component: https://github.com/descope/web-components-ui/pull/156

nirgur:
@tomerlichtash is this PR still relevant?

tomerlichtash:
> @tomerlichtash is this PR still relevant?

No, fixed with clickableMixin. Closing.

PR #287, opened 2023-10-02
fix: large session expiration consideration 

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/4292

## Description
when session expiration is above 24.8 days - setTimeout will fire immediately - resulting a loop of refresh where eventually it will result 429

see more details in https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value

the solution here is to limit upper session with that upper bound 

here is where session expiration is 30 days
Descope WC demo app 2023-10-02 19-10-43


shilgapira:
👍👍

PR #297, opened 2023-11-01
feat: AB testing client condition calculation

aviadl:
Also now we know how to handle predicates in conditions
+ tests
fixes https://github.com/descope/etc/issues/2333


orius123 (approved):
Neat!

PR #301, opened 2023-11-07
feat: WC now initializes fingerprint according to config

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/4307

### web-component
**breaking change**: web-component now initializes fingerprint according to the flow config as opposed to the now removed ~`telemetryKey`~.
 - When enabled, the flow config will contain the required public key to enable calculating device fingerprints.
 - When disabled - any fingerprint information will be removed

### web-js
To facilitate this change `web-js` was updated to expose the needed functions to dynamically enable fingerprint through the flow-config. The `web-js` behavior is not changed.
- When using `withFingerprint` it will set the `beforeRequest` even if the FP key is unavailable as it can now dynamically be set by the web component.

## Must

- [X] Tests


meirwah:
@nirgur @tomerlichtash can you review ?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -175,6 +179,15 @@ class DescopeWc extends BaseDescopeWc {
 
     // if there is no execution id we should start a new flow
     if (!executionId) {
+      if (
+        flowConfig.fingerprintEnabled &&
+        flowConfig.fingerprintKey?.length > 0
```
tomerlichtash, line 184: `flowConfig.fingerprintKey`

packages/web-js-sdk/src/enhancers/withFingerprint/index.ts
```diff
@@ -20,20 +21,24 @@ const beforeRequest: BeforeRequestHook = (config) => {
...
-      });
-    }
     if (!IS_BROWSER) {
       // eslint-disable-next-line no-console
       console.warn(
         'Fingerprint is a client side only capability and will not work when running in the server'
       );
-    } else if (fpLoad) {
-      ensureFingerprintIds(fpKey).catch(() => null);
+      return createSdk({
```
tomerlichtash, line 29: return createSdk(config);

tomerlichtash (changes requested):
Add few comments

tomerlichtash (approved):

PR #317, opened 2023-11-15
feat: Select tenant

aviadl:
+tests
related to https://github.com/descope/etc/issues/3078 Added an API call for selecting a tenant and placing it on the jwt as a custom claim


asafshen (approved):
not related to this PR directly - but lets just make sure that this will work from react-sdk/web-js-sdk when the user has a token persisted (the web-sdk should handle this, need to make sure the service impl also does)

image

PR #327, opened 2023-12-05
fix: issue 4877 RELEASE

nirgur:
## Related Issues

Related https://github.com/descope/etc/issues/4877

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -17,7 +17,8 @@ import {
   handleAutoFocus,
   injectSamlIdpForm,
   isConditionalLoginSupported,
-  replaceWithScreenState,
+  updateScreenFromScreenState,
+  updateTemplateFromScreenState,
   setTOTPVariable,
   showFirstScreenOnExecutionInit,
   State,
@@ -113,7 +114,7 @@ class DescopeWc extends BaseDescopeWc {
       } catch (ex) {
         this.loggerWrapper.error(
           `Failed to fetch flow page from ${htmlLocaleUrl}. Fallback to url ${htmlUrl}`,
-          ex
+          ex,
         );
       }
     }
@@ -130,7 +131,7 @@ class DescopeWc extends BaseDescopeWc {
   async onFlowChange(
     currentState: FlowState,
     prevState: FlowState,
-    isChanged: IsChanged<FlowState>
+    isChanged: IsChanged<FlowState>,
   ) {
     const {
       projectId,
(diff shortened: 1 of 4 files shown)


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -739,11 +740,16 @@ class DescopeWc extends BaseDescopeWc {
...
         this.shadowRoot.querySelector('div'),
-        screenState?.totp?.image
+        screenState?.totp?.image,
       );
 
       this.rootElement.replaceChildren(clone);
 
+      // we need to wait for all components to render before we can set its value
+      setTimeout(() =>
+        updateScreenFromScreenState(this.rootElement, screenState),
```
orius123, line 750: So after applying this fix, we will see the former OTP code autofilled?

nirgur: Yep

orius123 (approved):

packages/web-component/src/lib/helpers/templates.ts
```diff
@@ -89,16 +90,15 @@ const replaceProvisionURL = (
...
 
 /**
- * Perform action in base element based on screen state
+ * Update a screen template based on the screen state
  *  - Show/hide error messages
- *  - Replace values of element inputs with screen state's inputs
  *  - Replace element templates ({{...}} syntax) with screen state object
  */
-export const replaceWithScreenState = (
+export const updateTemplateFromScreenState = (
```
asafshen, line 97: what is the case of calling `updateTemplateFromScreenState` without `updateScreenFromScreenState`?

also noting that names are a bit confusing to me, the `template` vs `screen` terminology is not that obvious for me, maybe we can call this one `updateScreenFromScreenState` and the bellow `updateInputsFromScreenState`

nirgur: > what is the case of calling `updateTemplateFromScreenState` without `updateScreenFromScreenState`?

Both fn should be called, just in a different timing

> also noting that names are a bit confusing to me, the `template` vs `screen` terminology is not that obvious for me, maybe we can call this one `updateScreenFromScreenState` and the bellow `updateInputsFromScreenState`

IMO `updateInputsFromScreenState` is a bit misleading, because we are not updating only inputs
One updates the template of the screen, before it's mounted, the second, update the screen, after the template is already mounted

asafshen: > Both fn should be called, just in a different timing


gotcha, maybe one should be called from the other and not to expose both to the caller, not critical

PR #329, opened 2023-12-07
feat: add form and client custom flow inputs override

Bars92:
## Related Issues
related https://github.com/descope/etc/issues/2954

## Related PRs
https://github.com/descope/orchestrationservice/pull/1788
react sdk:

## Description
add tests
add custom form and client context override params

## Must

- [x] Tests


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -597,6 +620,15 @@ class BaseDescopeWc extends HTMLElement {
         tenant: this.tenant,
         redirectUrl: this.redirectUrl,
         locale: this.locale,
+        form: Object.entries(this.form || {}).reduce(
```
nirgur, line 623: Why do we need to generate the `form.` entries?

In addition, why do we need to add it to the flowState?
We can just merge it with the screen state when relevant, WDYT?

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
```
nirgur, line 5: Do we really need it?

nirgur (commented):
Looks good
Added a few comments
We can go over it together if you like

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: 1. due to a small gap where custom inputs are named `form.custom` we attempt to apply to custom inputs on any form key. so if you do `form={{ custom: "test" }}` it will always map to `form.custom` within the form.

2. good point, i can always use the `this.form` in init and `onFlowChange`

pnpm-lock.yaml
Bars92: no idea how it was added 😅 let me check

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -54,6 +54,8 @@ class BaseDescopeWc extends HTMLElement {
       'preview',
       'redirect-url',
       'auto-focus',
+      'form',
+      'client',
```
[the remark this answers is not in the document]
nirgur: I think that because we have a few async actions before the flow starts running, using props should be ok
But this is a good point, we can check it pretty easily if you like

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur: Regarding 1, 
We do not get the custom inputs on the flows api response right now, not sure if this is important or not

When looking on the flow response, I see that all flow inputs are directly under form, for example:
image

So after this change, we will have 
```
    "form": {
        "email": "nir@descope.com",
        "form.email": "nir@descope.com"
    }
```

Is this what we want to achieve? 


Regarding 2, 
I think it's better, it will simplify the code a bit

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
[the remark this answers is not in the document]
Bars92: it is hard to deterministically say that it would be ok. it does depend on the hosting application render cycles and the network speed. leaving it to js is bound for disappointment

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: 2 done 👍 
regarding 1: yes, this is what we want to achieve because the custom inputs are already prefixed with form to avoid overriding the original properties such as email. i didn't want to modify the custom inputs so we will autofill the form for now.
image

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur, line 58: We should not observe

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -146,6 +148,30 @@ class BaseDescopeWc extends HTMLElement {
...
+  get form() {
+    try {
+      const form = JSON.parse(this.getAttribute('form')) as Record<string, any>;
+      return Object.entries(form || {}).reduce(
+        (prev, [key, value]) => ({
+          ...prev,
+          [key]: value,
+          [`form.${key}`]: value,
+        }),
+        {},
```
nirgur, line 160: you can start with `form`

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -146,6 +148,30 @@ class BaseDescopeWc extends HTMLElement {
...
+      );
+    } catch (e) {
+      return undefined;
```
nirgur, line 163: maybe it's better to return an empty object

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -233,6 +233,8 @@ class DescopeWc extends BaseDescopeWc {
             samlIdpStateId,
             samlIdpUsername,
             ssoAppId,
+            client: this.client,
+            form: this.form,
```
nirgur, line 237: merge with input

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -448,6 +454,8 @@ class DescopeWc extends BaseDescopeWc {
             lastAuth,
             preview: this.preview,
             abTestingKey,
+            client: this.client,
+            form: this.form,
```
nirgur, line 458: no need

Bars92:
tested and works good 👍 i can safely remove form from the start options

nirgur (approved):

PR #333, opened 2023-12-10
feat: support dynamic attribute values RELEASE

nirgur:
related: https://github.com/descope/etc/issues/3405


asafshen (approved):
Very nice

asafshen:
Noting that we have some minor code duplication here with other functions (which just assigns content instead value, and with other selector)

I think it is totally ok (and actually better) here because making the code generic will also make it more complex and hard to read and debug, but worth checking  this assumption if we add similar functionality later on

PR #334, opened 2023-12-11
fix: issue 4907 RELEASE

nirgur:
fixes https://github.com/descope/etc/issues/4907

### DIFF
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -3,6 +3,7 @@ import {
   CONFIG_FILENAME,
   PREV_VER_ASSETS_FOLDER,
   THEME_FILENAME,
+  UI_COMPONENTS_FALLBACK_URL,
   UI_COMPONENTS_URL,
   UI_COMPONENTS_URL_VERSION_PLACEHOLDER,
 } from '../constants';
@@ -370,7 +371,7 @@ class BaseDescopeWc extends HTMLElement {
       styleEle.innerText =
         (theme?.light?.globals || '') + (theme?.dark?.globals || '');
 
-      const descopeUi = await this.descopeUI;
+      const descopeUi = await BaseDescopeWc.descopeUI;
       if (descopeUi?.componentsThemeManager) {
         descopeUi.componentsThemeManager.themes = {
           light: theme?.light?.components,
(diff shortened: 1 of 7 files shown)


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -488,66 +489,96 @@ class BaseDescopeWc extends HTMLElement {
...
-    }
+        return scriptEle;
+      };
 
-    this.descopeUI = new Promise((res) => {
-      const onError = () => {
-        this.loggerWrapper.error(
-          'Cannot load DescopeUI',
-          `Make sure this URL is valid and return the correct script: "${scriptEle.src}"`,
+      const generateScriptUrl = (
```
asafshen, line 526: great abstraction and separation of concerns to functions! not critical but maybe we should put them in a util file cuz they don't use `this` or anything from the current scope

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -488,66 +489,96 @@ class BaseDescopeWc extends HTMLElement {
...
+
+        scriptEle.addEventListener('error', onErrorInternal);
       };
 
-      scriptEle.addEventListener('load', () => {
-        if (!globalThis.DescopeUI) onError();
-        res(globalThis.DescopeUI);
-      });
+      (async () => {
+        const componentsVersion = await this.#getComponentsVersion();
```
asafshen, line 557: question about the `this` here, this line is defined as a static function, so it gets the value from the caller, is that right?

maybe worth passing it as an argument instead the implicit implicatio of the caller having this function defined 

not critical, lmk what you think

asafshen (approved):
very cool! and neat abstraction

left 2 small questions/comments, lmk what you think, feel free to merge and we can debate later on the impl

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur: Not sure I understand what do you mean by `it gets the value from the caller`

Basically we are initializing a static member (so we can share it across different instances of the web-component)

The init process is using non static methods (because there is no need to share it with other instances)

Do you suggest to change all the helper methods to static as well?

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur: Thanks!

I thought about it as well, but it's very specific to this implementation and I don't think we will use it elsewhere

That's why I decided to leave it here

If you think we should move it into a separate file, 
I can do it in 2 min

LMK what you think

orius123 (approved):
Nice

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur: But there is no static function here,
It's a static attribute

I suggest that we will have a short zoom call to close this

WDYT?

PR #336, opened 2023-12-19
feat: allow caller to omit cookiePolicy

### LINKED ISSUE #338: Bug on Cloudflare Pages/Workers
When deployed to Cloudflare edge runtime, which has its own native version of `fetch`, there is an issue with one of the parameters passed down to it.

Cloudflare doesn't support the `credentials` parameter in fetch. They documented the reason [here](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/http.h#L591).

In the [createHttpClient](https://github.com/descope/descope-js/blob/main/packages/core-js-sdk/src/httpClient/index.ts#L72) function, there is a parameter `cookiePolicy` that is translated to the `credentials` parameter in `fetch`. This is always passed, and defaults to `include` if not supplied by the caller. This breaks in the Cloudflare version of `fetch`.

mattolson:
## Description

* On Cloudflare, the `fetch` implementation does not support `credentials`. In this library, we create an http client using a `cookiePolicy` option that is then passed to `fetch`.
* This change allows the caller to pass `cookiePolicy: null` to `createHttpClient`, which will omit the `credentials` option when calling `fetch`. If null is not passed, the previous behavior is preserved.
* See https://github.com/cloudflare/workerd/blob/main/src/workerd/api/http.h#L591 for details on why Cloudflare will not implement `credentials` in their version of `fetch`.

Fixes #338 

## Must

- [x] Tests

### DIFF
--- a/packages/core-js-sdk/src/createSdk.ts
+++ b/packages/core-js-sdk/src/createSdk.ts
@@ -11,7 +11,7 @@ type SdkConfig = {
   logger?: Logger;
   baseUrl?: string;
   hooks?: Hooks;
-  cookiePolicy?: RequestCredentials;
+  cookiePolicy?: RequestCredentials | null;
   baseHeaders?: HeadersInit;
   fetch?: Fetch;
 };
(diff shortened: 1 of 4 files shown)


mattolson:
@asafshen

asafshen:
@mattolson thanks for sharing this as well!

mattolson:
I updated the PR title and removed a mistakenly included second commit, so this should be ready to go.

dorsha (approved):
Very nice!

PR #337, opened 2023-12-19
feat: support password managers RELEASE

nirgur:
fixes https://github.com/descope/etc/issues/4430

1password:

https://github.com/descope/descope-js/assets/11805663/fe5ea2c2-99cf-47ce-9f8a-56f643c2cb1f

Safari:

https://github.com/descope/descope-js/assets/11805663/a1a0eb47-7d5e-4b13-98b6-1adaaca35b27

Chrome:

https://github.com/descope/descope-js/assets/11805663/1bef9a0c-9899-41e9-96a7-837e2c311252

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -24,6 +24,7 @@ import {
   State,
   submitForm,
   withMemCache,
+  getFirstNonEmptyValue,
 } from '../helpers';
 import { calculateConditions, calculateCondition } from '../helpers/conditions';
 import { getLastAuth, setLastAuth } from '../helpers/lastAuth';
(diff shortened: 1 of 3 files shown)


dorsha (approved):
Very nice!!!
@meirwah see videos.

meirwah:
Well done !!!

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -831,6 +832,25 @@ class DescopeWc extends BaseDescopeWc {
...
 
+  // handle storing passwords in password managers
+  #handleStoreCredentials(formData = {}) {
+    const idFields = ['externalId', 'email', 'phone'];
+    const passwordFields = ['newPassword', 'password'];
+
+    const id = getFirstNonEmptyValue(formData, idFields);
+    const password = getFirstNonEmptyValue(formData, passwordFields);
+
+    if (id && password) {
```
meirwah, line 843: @nirgur , performance wise, isn't it more affective to first check if :
```
const password = getFirstNonEmptyValue(formData, passwordFields);
if (password) {
```

and only if true, do the next check:
```
const id = getFirstNonEmptyValue(formData, idFields);
``` 

??
since this is called on every componant , it could have a big impact no?

nirgur: In general, yes, but It's completely negligible
This code is running only when clicking on next in a flow screen
we are running on 2 keys, don't know if this is something we can even measure
I think we can ignore it

PR #346, opened 2023-12-31
feat: add auto-submit optional behavior to code input OTP

nmacianx:
## Related Issues

Related to https://github.com/descope/etc/issues/2076

## Description

The passcode component emits a custom 'valid' event whenever its value is valid, so the descope-wc now adds a listener whenever the passcode component has been configured for autosubmit.
This feature can't be tested in an isolated way since we need the actual `descope-passcode` component for it, but an integration test should be added.

## Must

- [x] Tests


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -853,39 +853,58 @@ class DescopeWc extends BaseDescopeWc {
...
+      projectConfig.componentsVersion,
+      actionArgs,
+    );
+
+    this.#handleSdkResponse(sdkResp);
+
+    this.#handleStoreCredentials(formData);
+  }
+
   async #handleSubmit(submitter: HTMLButtonElement, next: NextFn) {
```
nirgur, line 887: I think we can keep the original function and use it also for auto submit

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -902,6 +921,21 @@ class DescopeWc extends BaseDescopeWc {
           this.#handleSubmit(button, next);
         };
       });
+
+    this.rootElement
```
nirgur, line 925: I think it will be better if we can extract it into a function

nirgur (approved):

PR #347, opened 2024-01-04
feat: add backup callback to redirect auth RELEASE

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/5062

## Description

Add new backup callback redirect auth parameter

## Must

- [X] Tests


packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -179,6 +180,7 @@ class DescopeWc extends BaseDescopeWc {
         ? {
             callbackUrl: redirectAuthCallbackUrl,
             codeChallenge: redirectAuthCodeChallenge,
+            backupCallbackUri: redirectAuthBackupCallbackUri,
```
asafshen, line 183: is there is a chance that `redirectAuthBackupCallbackUri` exists and `redirectAuthCallbackUrl` does not exists?

image

packages/web-component/src/lib/helpers/helpers.ts
```diff
@@ -267,6 +275,7 @@ export const handleUrlParams = () => {
     exchangeError,
     redirectAuthCodeChallenge,
     redirectAuthCallbackUrl,
+    redirectAuthBackupCallbackUri,
```
asafshen, line 278: not critical, naming wise, I understand that technically this is a URI, but having the `redirectAuthCallbackUrl` (which is also URI_ named with `Url`, I would rename the backup callback also "url" and not "uri" for consistency 

wdyt?

asafshen (commented):
looks good, left 2 questions

packages/web-component/src/lib/descope-wc/DescopeWc.ts
itaihanski: Good question - It's not allowed

packages/web-component/src/lib/helpers/helpers.ts
itaihanski: This is intentional - the URL is actually an http/https based url (app link / universal link). The backup URI is however a custom scheme URI, used only on Android + Opera at the moment. This is to distinguish between them

asafshen: cool 👍 
as far as I know the fact that there is a different scheme (aka myapp:// instead https://) doesn't affect this terminology between url to uri but if it make sense to you its enough

asafshen (approved):
great 👍

PR #352, opened 2024-01-17
fix: Change persist-tokens hook to always add token (if exists) t http requests

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/5319

## Description
change web-js-sdk to always send the token if it exists

like mentioned in the issue, someone may rely on the fact that the token is also in the local storage\
so we should document it clearly in RN that this may break existing code that relies on the previous buggy behavior (I don't want to bump major for this bug though)


dorsha:
Looks good to me, @nirgur please review as well.

asafshen:
> I'm not sure I fully understand this change
> 
> Can we talk about it?

Would love to :)

asafshen:
closing in favor to another solution

PR #353, opened 2024-01-18
fix: DS cookie domain RELEASE

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/5332

## Description

Store the cookie without a domain 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies
https://github.com/js-cookie/js-cookie?tab=readme-ov-file#domain

image

in IE it will be also available to subdomains


packages/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -18,13 +18,13 @@ import {
...
   value: string,
-  { cookiePath, cookieDomain, cookieExpiration }: Partial<JWTResponse>
+  { cookiePath, cookieExpiration }: Partial<JWTResponse>,
 ) {
   if (value) {
     const expires = new Date(cookieExpiration * 1000); // we are getting response from the server in seconds instead of ms
+    // Since its a JS cookie, we don't set the domain because we want the cookie to be on the same domain as the application
     Cookies.set(name, value, {
       path: cookiePath,
-      domain: cookieDomain,
```
asafshen, line 27: need to check if cookie domain is equal or "parent" domain of application domain
e.g. cookie domain - b.com, app domain a.b.com - should work

asafshen: done, @nirgur lmk what you think

nirgur (approved):

PR #356, opened 2024-01-24
chore: change hooks to be taken by ref

asafshen:
## Related Issues
https://github.com/descope/etc/issues/5319

## Description
change hooks to be calculated on function runtime instead on config time

will later enables taking them in a dynamic manner


packages/core-js-sdk/src/createSdk.ts
```diff
@@ -30,21 +30,26 @@ const withMultipleHooks =
         beforeRequest?: BeforeRequest | BeforeRequest[];
         afterRequest?: AfterRequest | AfterRequest[];
       };
-    }
+    },
   ) => {
-    const beforeRequestHooks = [].concat(config.hooks?.beforeRequest || []);
-    const afterRequestHooks = [].concat(config.hooks?.afterRequest || []);
+    const beforeRequest: BeforeRequest = (c) => {
```
nirgur, line 35: Can we change `c` to `conf`?

packages/core-js-sdk/test/createSdk.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const mockFetch = jest.fn();
+globalThis.fetch = mockFetch;
+
+describe('createSdk', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should call "beforeRequest" and "afterRequest"', async () => {
```
nirgur, line 11: I would split it into 2 different tests and test that when changing the hook it's actually working
I think that we already have tests for calling hooks

nirgur (commented):
Cool
Did you check it's actually working?

asafshen:
> Cool
Did you check it's actually working?

Of Course

packages/core-js-sdk/test/createSdk.test.ts
asafshen: unfortunately we don't have tests for this code
(only with mocks of the inner object)

packages/core-js-sdk/src/createSdk.ts
asafshen: done

packages/core-js-sdk/test/createSdk.test.ts
asafshen: good idea for splitting, done 👍

nirgur: Cool, thanks!
Don't you want to add a test where the hooks are updated after initialization?
just to make sure someone will not change it without noticing

asafshen: good idea 👍  added such test

nirgur (approved):

PR #371, opened 2024-01-30
fix: ErrorHandling RELEASE

nirgur:


packages/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -38,7 +38,7 @@ export const initNotificationsMixin = createSingletonMixin(
       #createNotification(type: Notification['type']) {
         return this.createNotification({
           mode: type,
-          duration: 3000,
+          duration: type === 'error' ? 0 : 3000,
```
tomerlichtash, line 41: Move to duration to const, and the condition to an expression like `isStaticNotification`

packages/user-management-widget/src/lib/widget/state/asyncActions/createUser.ts
```diff
@@ -24,7 +24,14 @@ const reducer = buildAsyncReducer(action)(
   withRequestStatus((state: State) => state.createUser),
   withNotifications({
     getSuccessMsg: () => 'User created successfully',
-    getErrorMsg: () => 'Failed to create user',
+    getErrorMsg: (action) => {
+      const errorMsg = action.error?.message;
+      return `
+      <div>
```
tomerlichtash, line 30: Why not use template?

tomerlichtash (approved):

PR #373, opened 2024-02-03
fix: sso button click

dorsha:
## Related Issues

Fixes https://github.com/descope/etc/issues/5524

## Description

Fix sso button click when there are no other regular/sso buttons

## Must

- [x] Tests

### DIFF
--- a/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -485,6 +485,15 @@ class BaseDescopeWc extends HTMLElement {
       // in case there is a single "generic" button on the page, click on it
       if (genericButtons.length === 1) {
         genericButtons[0].click();
+      } else if (genericButtons.length === 0) {
+        const ssoButtons = Array.from(buttons).filter(
+          (button) => button.getAttribute('data-type') === 'sso',
+        );
+
+        // in case there is a single "sso" button on the page, click on it
+        if (ssoButtons.length === 1) {
+          ssoButtons[0].click();
+        }
       }
     };
   }
--- a/packages/web-component/test/descope-wc.test.ts
+++ b/packages/web-component/test/descope-wc.test.ts
@@ -702,6 +702,50 @@ describe('web-component', () => {
     await waitFor(() => expect(nextMock).not.toHaveBeenCalled());
   });
 
+  it('When there are multiple "sso" buttons and pressing on enter, it does not click any button', async () => {
+    startMock.mockReturnValueOnce(generateSdkResponse());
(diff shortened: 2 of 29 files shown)


packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -485,6 +485,15 @@ class BaseDescopeWc extends HTMLElement {
       // in case there is a single "generic" button on the page, click on it
       if (genericButtons.length === 1) {
         genericButtons[0].click();
+      } else if (genericButtons.length === 0) {
+        const ssoButtons = Array.from(buttons).filter(
```
tomerlichtash, line 489: We can use `find` instead of filter and get the button or undefined. This will make the `length === 1` unnecessary, and will always return the first button, in case there are multiple SSO buttons (if that case is possible).

packages/web-component/test/descope-wc.test.ts
```diff
@@ -618,6 +618,37 @@ describe('web-component', () => {
     await waitFor(() => expect(nextMock).toHaveBeenCalled());
   });
 
+  it('When there is a single "sso" button and pressing on enter, it clicks the button', async () => {
+    startMock.mockReturnValueOnce(generateSdkResponse());
+    nextMock.mockReturnValueOnce(generateSdkResponse({ screenId: '1' }));
+
+    pageContent =
```
tomerlichtash, line 625: I would add a couple of SSO buttons to see the first one is picked, and validate that if there are multiple buttons (some might be not connected) the the first is always picked.

tomerlichtash (changes requested):

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
dorsha: See few lines above, it does the exact same so I used the same paradigm, if we change we should change other places as well, and not sure I want to do that now, since it might be risky change.

packages/web-component/test/descope-wc.test.ts
dorsha: If there are many non should be picked, will add a test for that.

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
dorsha: Just checked and `find` is not good here, since we want to not click anything if there are more than one button (and not click the first one).

packages/web-component/test/descope-wc.test.ts
dorsha: added

dorsha:
Closing, will be treated as feature request.

PR #376, opened 2024-02-05
feat: OneTap support

orius123:
## Related Issues

Related https://github.com/descope/etc/issues/774

### DIFF
--- a/packages/web-js-sdk/src/sdk/fedcm.ts
+++ b/packages/web-js-sdk/src/sdk/fedcm.ts
@@ -1,5 +1,4 @@
 import { JWTResponse, SdkResponse, LoginOptions } from '@descope/core-js-sdk';
-import { IS_BROWSER } from '../constants';
 import { CoreSdk } from '../types';
 
 /**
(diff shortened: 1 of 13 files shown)


packages/web-js-sdk/src/sdk/fedcm.ts
```diff
@@ -0,0 +1,152 @@
...
+};
+
+/**
+ * Constructs a higher level FedCM API that wraps the functions from code-js-sdk.
+ * @param sdk The CoreSdk instance.
+ * @returns The FedCM API.
+ */
+const createFedCM = (sdk: CoreSdk) => ({
+  async oneTap(
+    provider: string,
```
nirgur, line 76: Is this the API we want?
passing the provider is a bit useless in this case no?
I guess we will have a dedicated provider for it

WDYT?

nirgur (commented):
Well done!
Only one question regarding the API

packages/web-js-sdk/src/sdk/fedcm.ts
orius123: Made it optional with default to google

nirgur (approved):

PR #390, opened 2024-02-25
fix: get async ele

tomerlichtash:
## Related Issues

Fixes <link_to_github_issue>

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| service a PR | Link to PR |
| service b PR | Link to PR |

## Description

A few sentences describing the overall goals of the pull request's commits.

## Must

- [ ] Tests

### DIFF
--- a/packages/user-management-widget/src/lib/mixins/formMixin.ts
+++ b/packages/user-management-widget/src/lib/mixins/formMixin.ts
@@ -46,7 +46,6 @@ export const formMixin = createSingletonMixin(
         this.getFormInputs(rootEle).forEach((input) => {
           // eslint-disable-next-line no-param-reassign
           input.value = '';
-          input.removeAttribute('invalid');
         });
       }
     },
--- a/packages/user-management-widget/src/lib/mixins/formMixin.ts
+++ b/packages/user-management-widget/src/lib/mixins/formMixin.ts
@@ -46,6 +46,7 @@ export const formMixin = createSingletonMixin(
         this.getFormInputs(rootEle).forEach((input) => {
           // eslint-disable-next-line no-param-reassign
           input.value = '';
+          input.removeAttribute('invalid');
         });
       }
     },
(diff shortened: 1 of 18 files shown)


packages/user-management-widget/src/lib/mixins/formMixin.ts
```diff
@@ -46,6 +46,7 @@ export const formMixin = createSingletonMixin(
         this.getFormInputs(rootEle).forEach((input) => {
           // eslint-disable-next-line no-param-reassign
           input.value = '';
+          input.removeAttribute('invalid');
```
nirgur, line 49: I don't think we should change the components invalid attribute from the driver
Why do we need it?

packages/user-management-widget/src/lib/widget/drivers/BaseDriver.ts
```diff
@@ -3,6 +3,21 @@ import { Logger } from '../../mixins/loggerMixin/types';
 type Empty = null | undefined;
 type RefOrRefFn = Element | (() => HTMLElement | Empty) | Empty;
 
+const waitForElement = async (ele: RefOrRefFn, timeout: number) =>
```
nirgur, line 6: Helper file?

nirgur (commented):
👍 
Added 2 comments

packages/user-management-widget/src/lib/mixins/formMixin.ts
tomerlichtash: We reset the form, but the validations states remains dirty. This is intended to clear the validation state as well.

packages/user-management-widget/src/lib/widget/drivers/BaseDriver.ts
tomerlichtash: Done. Also moved types to a separate file to avoid circular dependency.

nirgur (approved):

PR #411, opened 2024-03-11
chore: fingerprinting with custom domain

talaharoni:
## Related Issues

Related to https://github.com/descope/etc/issues/5367

## Description
Support fingerprinting with custom domain

## Must
- [x] Tests

### DIFF
--- a/packages/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -185,7 +185,7 @@ class DescopeWc extends BaseDescopeWc {
     // if there is no execution id we should start a new flow
     if (!executionId) {
       if (flowConfig.fingerprintEnabled && flowConfig.fingerprintKey) {
-        ensureFingerprintIds(flowConfig.fingerprintKey);
+        await ensureFingerprintIds(flowConfig.fingerprintKey, this.baseUrl);
       } else {
         clearFingerprintData();
       }
--- a/packages/web-js-sdk/src/enhancers/withFingerprint/constants.ts
+++ b/packages/web-js-sdk/src/enhancers/withFingerprint/constants.ts
@@ -3,10 +3,12 @@ import { IS_BROWSER } from '../../constants';
 const FINGERPRINT_PUBLIC_KEY = 'fingerprint.public.key';
 const FINGERPRINT_ENDPOINT_URL = 'fingerprint.endpoint.url';
 
-/** Fingerprint.js custom API endpoint */
+/** Fingerprint.js cloudflare integration */
 export const FP_EP_URL =
   (IS_BROWSER && localStorage?.getItem(FINGERPRINT_ENDPOINT_URL)) ||
-  'https://fp.descope.com';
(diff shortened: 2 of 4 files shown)


itaihanski (approved):
Excellent!

packages/web-js-sdk/src/enhancers/withFingerprint/constants.ts
```diff
@@ -3,10 +3,12 @@ import { IS_BROWSER } from '../../constants';
 const FINGERPRINT_PUBLIC_KEY = 'fingerprint.public.key';
 const FINGERPRINT_ENDPOINT_URL = 'fingerprint.endpoint.url';
 
-/** Fingerprint.js custom API endpoint */
+/** Fingerprint.js cloudflare integration */
 export const FP_EP_URL =
   (IS_BROWSER && localStorage?.getItem(FINGERPRINT_ENDPOINT_URL)) ||
```
asafshen, line 8: just noting that this is somewhat duplication from core-js const (api.descope.com) 

I think its ok, just worth tracking in case change 👍

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -185,7 +185,7 @@ class DescopeWc extends BaseDescopeWc {
     // if there is no execution id we should start a new flow
     if (!executionId) {
       if (flowConfig.fingerprintEnabled && flowConfig.fingerprintKey) {
-        ensureFingerprintIds(flowConfig.fingerprintKey);
+        await ensureFingerprintIds(flowConfig.fingerprintKey, this.baseUrl);
```
asafshen, line 188: I understand why we need to await, but this will slow the flow

do you think we should maybe communicate this in the "Enable FP" toggle button in Flow Editor?

packages/web-js-sdk/src/enhancers/withFingerprint/helpers.ts
```diff
@@ -63,25 +69,47 @@ const getFPFromStorage = (returnExpired = false): FingerprintObject => {
...
-    const agentP = load({ apiKey: fpKey, endpoint: FP_EP_URL });
+
+    const endpointUrl = new URL(baseUrl);
+    endpointUrl.pathname = FP_CF_ENDPOINT_PATH;
+
+    const patterUrl = new URL(baseUrl);
+    patterUrl.pathname = FP_CF_SCRIPT_PATH;
+    const scriptUrlPattern =
+      patterUrl.toString() +
+      '?apiKey=<apiKey>&version=<version>&loaderVersion=<loaderVersion>';
```
asafshen, line 91: I'm not too familiar with this pattern, what are those query params? CF thing?

asafshen (approved):
looks great, left some comments

packages/web-js-sdk/src/enhancers/withFingerprint/helpers.ts
talaharoni: Fingerprint library replace them with the correct values when performing the call.
The CF thing is only the `path` part

PR #498, opened 2024-04-09
fix(core-js-sdk): allow path in baseurl

omercnet:
this will allow using a base url with a path in it for use cases where descopers want to use a rewrite url to proxy to descope instead of a custom domain


asafshen:
code looks good
any idea what this is? perhapse Nir will know
image

omercnet:
@asafshen @nirgur ready

asafshen (approved):

PR #501, opened 2024-04-10
Add session and refresh expiration utility functions

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/6151

## Description

Add session and refresh expiration utility functions

## Must

- [X] Tests


packages/web-js-sdk/src/enhancers/withPersistTokens/index.ts
```diff
@@ -66,10 +69,14 @@ export const withPersistTokens =
 
     const refreshToken = () => getRefreshToken(storagePrefix);
     const sessionToken = () => getSessionToken(storagePrefix);
+    const isRefreshTokenExpired = () => isTokenExpired(refreshToken());
+    const isSessionTokenExpired = () => isTokenExpired(sessionToken());
```
asafshen, line 73: We also need to give ways to give session, similarly to how `getJwtPermissions` works

for example - there is a scenario where the user doesn't want to persist token, need to consider that case as well

itaihanski: @asafshen I found this in core SDK helpers, isn't it exactly that?
image

asafshen: cool, didn't know this code exists

yeah, so we can expose it in react/vue/etc sdks like this, wdyt?
image

itaihanski: Looks good to me. I would add an identical function for refresh as well. That snippet is not from this repo right?

asafshen: yeah I agree about the refresh 👍 

yes, its from react-sdk repo

PR #520, opened 2024-04-15
feat: Disable email input component using flow input

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/6164

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| service a PR | Link to PR |
| service b PR | Link to PR |

## Description

An enhancement for using  Input Flows API to enable/disable inputs.

API:

```
form='{"kuku": "123", "email":"my@email.com", "kuku.disabled":"false", "email.disabled": "true"}'
```
Results in:
image

## Must

- [x] Tests


packages/web-component/test/descope-wc.test.ts
```diff
@@ -3831,4 +3831,22 @@ describe('web-component', () => {
       });
     });
   });
+
+  describe('Input Flows', () => {
```
tomerlichtash, line 3835: WIP

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -774,10 +774,22 @@ class DescopeWc extends BaseDescopeWc {
       screenState.form.email = currentState.samlIdpUsername;
     }
 
+    const formAttr = this.getAttribute('form') || '';
```
nirgur, line 777: We already have `this.form` that returns the parsed form

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -155,18 +155,52 @@ class BaseDescopeWc extends HTMLElement {
...
+
+      return Object.entries(formData).reduce(
+        (prev, [name, value]) => ({
           ...prev,
-          [`form.${key}`]: value,
+          [`form.${name}`]: value,
         }),
-        form,
+        formData,
       );
```
Bars92, line 188: i dont think we need to force single type configurations. You can probably add validation making sure its a string or an object with `value` before, but this should always be good:

```
return Object.entries(form).reduce(
        (prev, [name, value]) => ({
          ...prev,
          [`form.${name}`]: typeof value === 'string' ? { value } : {...value},
        }),
        formData,
      );
```

also mind that `typeof j === 'object'` is true for functions and null objects, might cause issues.

packages/web-component/src/lib/helpers/templates.ts
```diff
@@ -104,6 +104,20 @@ const replaceHrefByDataType = (
...
+  formData: Record<string, string>,
+) => {
+  Object.keys(formData).forEach((name) => {
+    const eles = baseEle.querySelectorAll(`[name="${name}"]`);
+    eles.forEach((ele) => {
+      Object.keys(formData[name]).forEach((attr) => {
+        ele.setAttribute(attr, formData[name][attr]);
+      });
+    });
+  });
```
Bars92, line 118: this could be a bit risky. setting any attribute given by the user inputs can cause a security issue here. i think we need to force only what we support (disabled)

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
tomerlichtash: Not sure I follow - the configuration is backwards compatible, but we transform it here for consistency in all other operations.
the user can provide the existing configuration, or an object structure.

I'll take care of the `typeof` issue, thanks.

tomerlichtash: Also, [`typeof () => {}` returns `Function`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description), no?

packages/web-component/src/lib/helpers/templates.ts
tomerlichtash: @Bars92 Added allow list.

Bars92: @tomerlichtash where?

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: @tomerlichtash we can easily allow `{ email: "test@descope.com", email2: { value: "test@descope.com", disabled: true }}` - this is why i suggested just
```
return Object.entries(form).reduce(
        (prev, [name, value]) => ({
          ...prev,
          [`form.${name}`]: typeof value === 'string' ? { value } : {...value},
        }),
        form,
      );
```

tomerlichtash: IMHO seems inconsistent... what are the benefits of allowing this? Why not keep a single structure?

packages/web-component/src/lib/helpers/templates.ts
tomerlichtash: Weird github preview. Here: https://github.com/descope/descope-js/pull/520/commits/d1423d664618017a50ed384fa8462d27f1784a78#diff-14e4f9a9bb67ac1b8961131ee345a4b2eeec8792f6e6a081d0abb1c32d7dba3bR8

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
Bars92: there is no real reason to prevent this, even for testing purposes. forcing to change the object for consistency is futile. mind that also if value is empty maybe you need to force it to empty string to avoid undefined values

nirgur (commented):
Nice
Added a few comments

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
tomerlichtash: Changed to support both formats, added test as well.

packages/web-component/src/lib/helpers/templates.ts
```diff
@@ -104,6 +106,22 @@ const replaceHrefByDataType = (
...
+  Object.keys(formData).forEach((name) => {
+    const eles = baseEle.querySelectorAll(`[name="${name}"]`);
+    eles.forEach((ele) => {
+      Object.keys(formData[name]).forEach((attr) => {
+        if (ALLOWED_INPUT_CONFIG_ATTRS.includes(attr)) {
+          ele.setAttribute(attr, formData[name][attr]);
+        }
+      });
+    });
+  });
```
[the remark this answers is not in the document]
tomerlichtash: Changed.

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -155,18 +155,52 @@ class BaseDescopeWc extends HTMLElement {
...
     } catch (e) {
       return {};
     }
   }
 
+  getFormAttr(attr: string) {
```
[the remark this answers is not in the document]
tomerlichtash: Changed

packages/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -155,18 +155,52 @@ class BaseDescopeWc extends HTMLElement {
...
+    return Object.fromEntries(
+      Object.entries(this.form).map(([name, values]) => [name, values[attr]]),
+    );
+  }
+
+  get formValues() {
```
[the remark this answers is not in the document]
tomerlichtash: Changed. Will look into suggested optimization.

packages/web-component/src/lib/helpers/flowInputs.ts
```diff
@@ -0,0 +1,37 @@
+export const transformFlowInputFormData = (data: string) => {
+  try {
+    const form = JSON.parse(data);
+
+    let formData = form;
+    const vals = Object.values(form);
+
+    // transform values to object structure if needed
+    if (vals.some((s) => typeof s === 'string')) {
```
nirgur, line 9: Why do we need this block?
Can't we just do it in the reduce below?

tomerlichtash: I believe we want to duplicate the field's names (e.g. have `email` and `form.email` as well, @Bars92 - fact checking this with @Bars92), so we normalize the structure, otherwise - the `prev` accumulator merges with the original values that don't match the nested structure.

Instead, we can return this :

```
Object.fromEntries(
      Object.keys(form)
        .map((key) =>
          typeof form[key] !== 'string'
            ? [key, form[key]]
            : [key, { value: form[key] }]
        ).flatMap(([name, value]) => ([
          [name, value],
          [`form.${name}`, value]
        ])))
```

WDYT @nirgur?

tomerlichtash: @nirgur Ok, validated with @Bars92 - we do need the duplicated field names. Pushed the above fix.

nirgur (dismissed):

packages/web-component/src/lib/helpers/templates.ts
```diff
@@ -104,6 +106,23 @@ const replaceHrefByDataType = (
   });
 };
 
+const enableDisableInputs = (
```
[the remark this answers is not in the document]
tomerlichtash: Renamed. Please re-approve.

Bars92 (approved):

PR #561, opened 2024-05-03
fix: wrap visibility change in condition

asafshen:
## Related Issues
https://github.com/descope/react-sdk/issues/505

## Description
protect the document access with an `IS_BROWSER` condition


dorsha (approved):

dorsha:
Looks like tests fail because of the change.

PR #581, opened 2024-05-16
Transform resp hook

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/6428
https://github.com/descope/etc/issues/6304

## Description

- Remove `URL` completely as it adds unwanted behavior on react native and gives nothing in return
- Add new hook to allow for global response manipulation. This new hook will be used by the react native SDK

## Must

- [x] Tests
- [x] Documentation (if applicable)

### DIFF
--- a/packages/core-js-sdk/src/httpClient/helpers/index.ts
+++ b/packages/core-js-sdk/src/httpClient/helpers/index.ts
@@ -1,1 +1,14 @@
 export { default as createFetchLogger } from './createFetchLogger';
+
+export function transformSetCookie(setCookieHeader: string) {
+  // Split the header by semicolons to separate different attributes
+  var cookiesString = setCookieHeader.split(';');
+
+  return cookiesString.reduce((acc, cookie) => {
+    const [key, value] = cookie.split('=');
+    return {
+      ...acc,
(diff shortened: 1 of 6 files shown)


nirgur (approved):

packages/core-js-sdk/src/httpClient/helpers/index.ts
```diff
@@ -1 +1,14 @@
 export { default as createFetchLogger } from './createFetchLogger';
+
+export function transformSetCookie(setCookieHeader: string) {
+  // Split the header by semicolons to separate different attributes
+  var cookiesString = setCookieHeader.split(';');
+
+  return cookiesString.reduce((acc, cookie) => {
+    const [key, value] = cookie.split('=');
```
asafshen, line 8: @nirgur FYI I just had an interesting issue (with query params) that their value contains `=`

this means, that if the value of the cookie will contain `=`, the cookies won't get it
there, I did something quite naive to catch it
```
       const indexOfEquals = part.indexOf('=');
        const key = part.slice(0, indexOfEquals);
        const value = part.slice(indexOfEquals + 1);
        ```
        
        consider adding something similar here

PR #582, opened 2024-05-16
feat: Handle Password External Inputs RELEASE

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/6314


dorsha (dismissed):

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -658,6 +658,31 @@ class DescopeWc extends BaseDescopeWc {
...
+    }
+
+    const isExternalInput = ele.getAttribute('external-input') === 'true';
+
+    if (!isExternalInput) {
+      return;
+    }
+
+    const origInput = ele.shadowRoot.querySelector('input');
+    const id = `input-${origInput.getAttribute('id')}`;
```
nirgur, line 673: I think we can get the id directly from the element

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -799,9 +824,16 @@ class DescopeWc extends BaseDescopeWc {
...
-      setTimeout(() =>
-        updateScreenFromScreenState(this.rootElement, screenState),
-      );
+      setTimeout(() => {
+        updateScreenFromScreenState(this.rootElement, screenState);
+
+        const eles = this.rootElement.querySelectorAll('descope-password');
+
+        // remove existing external inputs
+        document.querySelectorAll('[data-hidden-input="true"]').forEach(ele => ele.remove());
```
nirgur, line 833: Why do we need to do it?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -896,7 +931,8 @@ class DescopeWc extends BaseDescopeWc {
     const id = getFirstNonEmptyValue(formData, idFields);
     const password = getFirstNonEmptyValue(formData, passwordFields);
 
-    if (id && password) {
+    // PasswordCredential not supported in Firefox
+    if (id && password && globalThis.PasswordCredential) {
```
nirgur, line 935: We have the same check 2 lines below, can we remove one?

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -658,6 +658,31 @@ class DescopeWc extends BaseDescopeWc {
...
+  #handleDescopePassword(ele: Element) {
+    if (!ele) {
+      return;
+    }
+
+    const isExternalInput = ele.getAttribute('external-input') === 'true';
+
+    if (!isExternalInput) {
+      return;
+    }
```
nirgur, line 670:
```suggestion
    if(ele?.getAttribute('external-input') !== 'true') {
       return;
    }
```

packages/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -658,6 +658,31 @@ class DescopeWc extends BaseDescopeWc {
     }
   }
 
+  #handleDescopePassword(ele: Element) {
```
nirgur, line 661: I think it worth a comment about the reason we are doing this

packages/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Removed

packages/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: To clean up hidden elements after each screen

nirgur (approved):

PR #586, opened 2024-05-19
fix: issue 6494

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/6494


dorsha:
@nirgur do we know when the regression started?

nirgur:
Based on the current code I think this is something that happened long ago

I'm not even sure if it's a regression or not

asafshen (approved):

dorsha:
@nirgur build fails

asafshen (approved):

PR #593, opened 2024-05-22
feat: use base url for static content

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/6022


orius123:
what about packages/libs/sdk-mixins/src/mixins/staticResourcesMixin/staticResourcesMixin.ts ?

dorsha (approved):

dorsha:
@nirgur need to release all sdks.
Also check @orius123 comment.

nirgur:
> what about packages/libs/sdk-mixins/src/mixins/staticResourcesMixin/staticResourcesMixin.ts ?

This is being used in the widgets
Do we want it also there?

orius123:
I think we do

dorsha:
Yes, should be also for widgets.

PR #659, opened 2024-06-30
fix: add json content to flow api

Bars92:
## Related Issues

related https://github.com/descope/etc/issues/6880

## Related PRs
https://github.com/descope/consoleservice/pull/2026

## Description
- fix content type for flow start and next routes

## Must

- [ ] Tests


asafshen:
I think it will be better to handle this in the receiving side if its possible,

if not, please write why, and make sure that this is communicated throughout the code comments, as future reader will not know why those APIs are the only one that needs this

orius123 (dismissed):
@nirgur @dorsha do we want to it for all APIs?

Bars92:
@asafshen its not possible. this is the content type of the request itself. sadly we haven't changed it and default is plain text which proto ignores. some apis may need this as well

asafshen:
> @asafshen its not possible. this is the content type of the request itself. sadly we haven't changed it and default is plain text which proto ignores. some POST apis may need this as well - they are all addressing to us so it probably doesn't really matter

noting that the fact that we can avoid this header so far improved simplicity (and even a bit of performance)

I'm not sure if its possible, but I would do the extra mile and attempty handle thiw in the internal service (even as plain text), rather then introducing this complexity here, but this current solution is also ok I guess, worth to add explanation

Bars92:
@asafshen ill add explanation but i rather do this which will fix the incorrect content type we are trying to use and avoid ignoring content type in console service to avoid security implications

asafshen:
cool, 

plus one for changing this in all places then 👍

Bars92:
@asafshen done

packages/core-js-sdk/src/httpClient/index.ts
```diff
@@ -36,6 +40,16 @@ const createDescopeHeaders = () => {
   };
 };
 
+const isJson = (value?: string) => {
+  try {
+    value = JSON.parse(value);
```
asafshen, line 45: What's the harm of adding it always?

asafshen (dismissed):

packages/core-js-sdk/src/httpClient/index.ts
Bars92: its usually bad practice on non POST/PUT and sometimes they may not send json body. i wouldn't want to change anything

asafshen (approved):
👍

PR #668, opened 2024-07-02
feat: init sdk scripts support

asafshen:
## Related Issues

https://github.com/descope/etc/issues/6582

## Description
mechanism for sdk scripts


nirgur:
We moved these files,
now it's under packages/sdks/web-component/...

Sorry for the inconvenience

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -212,6 +240,7 @@ class DescopeWc extends BaseDescopeWc {
 
     // if there is no execution id we should start a new flow
     if (!executionId) {
+      this.loadSdkScripts();
```
nirgur, line 243: Don't you think we should do it once the component is mounted?

packages/sdks/web-component/src/lib/descope-wc/sdkScripts/forter.ts
```diff
@@ -0,0 +1,29 @@
+const forterScript = `
```
nirgur, line 1: maybe add a comment to where this script is taken from

nirgur (commented):
Good job
added 2 small comments

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: I don't think so, it has a different lifecycle than the component, it may collect things that is not related to the component, and may actually benefit from loading as soon as possible

packages/sdks/web-component/src/lib/descope-wc/sdkScripts/forter.ts
asafshen: this code was provided to us by Forter team, I added a comment noting that 👍

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: I think it can be problematic because we might get here more than once 
but we can change it later in case we have issues

nirgur (dismissed):

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: I think you might be right, but I think that loading it also after mount may create other challenges 

we'll investigate moving further 👍

nirgur (approved):

PR #671, opened 2024-07-02
feat: Allow to keep last auth login ID after logout RELEASE

orius123:
## Related Issues

Related https://github.com/descope/etc/issues/6773

## Description 

Add new property to SDK `keepLastAuthenticatedUserAfterLogout` and to wc `keep-last-authenticated-user-after-logout` that when set will make the SDK not to clear the last auth login Id when the user logs out


packages/web-component/README.md
```diff
@@ -71,16 +71,17 @@ NOTE: This package is a part of a monorepo. so if you make changes in a dependen
...
+| auto-focus                                | **"true"** - Automatically focus on the first input of each screen</br>**"false"** - Do not automatically focus on screen's inputs</br>**"skipFirstScreen"** - Automatically focus on the first input of each screen, except first screen  | **"true"**    |
+| storage-prefix                            | **String** - A prefix to add to the key of the local storage when persisting tokens                                                                                                                                                        | **""**        |
+| store-last-authenticated-user             | **"true"** - Stores last-authenticated user details in local storage when flow is completed</br>**"false"** - Do not store last-auth user details. Disabling this flag may cause last-authenticated user features to not function properly | **"true"**    |
+| keep-last-authenticated-user-after-logout | **"true"** - Do not clear the last authenticated user details from the browser storage after logout</br>**"false"** - Clear the last authenticated user details from the browser storage after logout                                      | **"false"**   |
```
asafshen, line 83: the default value is true? ( I think it should be)   but I didn't see in the code

also the comment now is sort of double negative (do not clear) and its a bit confusing imo, maybe worth to have a positive sentence , "keep the last authenticated ...." 
wdyt?

packages/web-js-sdk/README.md
```diff
@@ -43,6 +43,10 @@ In addition, some browsers (e.g. Safari) may not store `Secure` cookie if the ho
   Note: Disabling this flag may cause features related to the last-authenticated user to not function properly.
   */
   storeLastAuthenticatedUser: false, // default is true
+  /* Do not clear the last authenticated user details from the browser storage after logout (default is false).
```
asafshen, line 46: reminder about the comment

packages/web-js-sdk/src/enhancers/withLastLoggedInUser/types.ts
```diff
@@ -0,0 +1,11 @@
+export type LastLoggedInUserOptions = {
+  // If true, the last authenticated user will be stored in local storage
+  // and will be used in the next flow start request
+  // Default is true
+  storeLastAuthenticatedUser?: boolean;
+  // If true, the last authenticated user will be kept in local storage even after logout
+  // and will be used in the next flow start request
+  // This option is relevant only if storeLastAuthenticatedUser is true
+  // Default is false
+  keepLastAuthenticatedUserAfterLogout?: boolean;
```
asafshen, line 10: is this option make sense even if `storeLastAuthenticatedUser` is not in used?
image

if it doens't make snese to use it without `storeLastAuthenticatedUser` maybe have it as conditional to `storeLastAuthenticatedUser: true`

not a must ofc

asafshen (commented):
looks great

just didn't understand the default value true (which I think should be)

packages/web-js-sdk/src/enhancers/withLastLoggedInUser/types.ts
orius123: it will just not delete stuff that are not there, so should be good

asafshen (approved):

PR #736, opened 2024-07-15
fix: Move deferredRedirect initialization to WC from base

itaihanski:
## Related Issues

Fixes https://github.com/descope/etc/issues/7075

## Description

Fix state initialization to avoid the base state overriding the WC state

## Must

- [x] Tests


nirgur (dismissed):

meirwah:
@nirgur @itaihanski CI fails :
image

nirgur (approved):

meirwah:
@nirgur @itaihanski any risk for regression with this change ?
something worth double checking/testing ?

itaihanski:
@meirwah Shouldn't be, the only platform that is affected is Android, and it fixes it

PR #743, opened 2024-07-18
fix: polling timeout when click is response t triggers another polling

asafshen:
## Description
Fix a case when click is response t triggers another polling

tested ML and EL with short and long polling


nirgur (approved):

dorsha:
@asafshen no need to release?

asafshen:
@dorsha it was released FYI

PR #744, opened 2024-07-18
feat: support multi style and style prop

Bars92:
related https://github.com/descope/etc/issues/4875

- add `styleName` prop
- allow load style based on style prop or default theme


packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -46,12 +46,16 @@ export const themeMixin = createSingletonMixin(
         return theme || 'light';
       }
 
+      get styleName(): string {
+        return this.getAttribute('style-name') || THEME_DEFAULT_NAME;
```
nirgur, line 50: We need to observe this attribute and call the relevant fns on change

packages/sdks/web-component/src/lib/constants/content.ts
```diff
@@ -11,5 +11,6 @@ export const ASSETS_FOLDER = 'v2-beta';
 export const PREV_VER_ASSETS_FOLDER = 'v2-alpha';
 
 // Those files are saved on a new folder to prevent breaking changes
-export const THEME_FILENAME = 'theme.json';
+export const THEME_DEFAULT_NAME = 'theme';
+export const THEME_DEFAULT_FILENAME = `${THEME_DEFAULT_NAME}.json`;
```
nirgur, line 15: Are we using it somewhere besides tests?

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -265,6 +270,7 @@ class BaseDescopeWc extends HTMLElement {
       'form',
       'client',
       'validate-on-blur',
+      'style-name',
```
nirgur, line 273: I think we need to support switching it in runtime
Currently this value can be set only on initial loading

nirgur (commented):
Looks good
added a few comments

In addition, I know it's not part of the task, but I really think we should start using the mixins also in the web component and this is a good opportunity to do it
It should simplify the code dramatically and we will not have to maintain this logic in 2 different places

WDYT? we can do it together if you like

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -170,6 +174,11 @@ export const themeMixin = createSingletonMixin(
           this.#loadFonts();
           this.#applyTheme();
         });
+
+        this.observeAttributes(['style-id'], () => {
+          this.#_themeResource = null;
+          this.#loadComponentsStyle();
```
nirgur, line 180: We also need to load the global style in this case

packages/sdks/web-component/src/lib/constants/content.ts
```diff
@@ -11,5 +11,6 @@ export const ASSETS_FOLDER = 'v2-beta';
 export const PREV_VER_ASSETS_FOLDER = 'v2-alpha';
 
 // Those files are saved on a new folder to prevent breaking changes
-export const THEME_FILENAME = 'theme.json';
+export const THEME_DEFAULT_NAME = 'theme';
```
nirgur, line 14:
```suggestion
export const THEME_DEFAULT_ID = 'theme';
```

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -127,34 +125,13 @@ class BaseDescopeWc extends HTMLElement {
   }
 
   #initShadowDom() {
-    this.attachShadow({ mode: 'open' });
+    //this.attachShadow({ mode: 'open' });
```
nirgur, line 128: We can remove it

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -782,47 +782,6 @@ class DescopeWc extends BaseDescopeWc {
     }
   }
 
-  async loadDescopeUiComponents(clone: DocumentFragment) {
```
nirgur, line 785: I think we should clean up the imports

packages/web-component/package-lock.json
nirgur, line 1: This file should be removed

nirgur (commented):
Looks good!
Added some comments

nirgur (dismissed):

nirgur (approved):

PR #745, opened 2024-07-19
fix: Revert "fix: polling timeout when click is response t triggers another polling"

asafshen:
Reverts descope/descope-js#743


dorsha (approved):

dorsha:
Make sure to release

PR #748, opened 2024-07-20
feat: OIDC error redirect URI

orius123:
Read oidc error redirect uri from the URL and pass it through to the flow


packages/sdks/web-component/src/lib/types.ts
```diff
@@ -34,6 +34,32 @@ export interface ScreenState {
...
+  ssoAppId?: string;
+} & OIDCOptions;
+
+export type OIDCOptions = {
+  oidcLoginHint?: string;
+  oidcPrompt?: string;
+  oidcErrorRedirectUri?: string;
+};
+
+export const extractOIDCOptions = (params: SSOQueryParams): OIDCOptions => {
```
nirgur, line 51: I don't think we should put fns in this file
What do we need it for?

nirgur (commented):
Looks good
Added a few comments

nirgur (approved):

PR #757, opened 2024-07-25
fix: issue 7166 RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/7166

Verified the scenario in the issue, 
In this case there is no response because there is no network so no error event will be dispatched,
Also added a default error object in case we did not get one from the server with the default values

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -311,7 +311,7 @@ class BaseDescopeWc extends HTMLElement {
           // return a generic error object in case of an error
           return {
             error: {
-              errorCode: 'E151000',
+              errorCode: 'J151000',
               errorDescription: e.toString(),
             },
           };
--- a/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -307,6 +307,14 @@ class BaseDescopeWc extends HTMLElement {
         try {
           const resp = await origFn(...args);
           return resp;
+        } catch (e) {
+          // return a generic error object in case of an error
+          return {
+            error: {
+              errorCode: 'E151000',
+              errorDescription: e.toString(),
+            },
+          };
         } finally {
           this.nextRequestStatus.update({ isLoading: false });
         }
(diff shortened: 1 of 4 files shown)


dorsha (dismissed):
Cool
1. Build fails
2. What about increasing the timeout and adding retries? Can be done in another pr

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -307,6 +307,14 @@ class BaseDescopeWc extends HTMLElement {
         try {
           const resp = await origFn(...args);
           return resp;
+        } catch (e) {
+          // return a generic error object in case of an error
+          return {
+            error: {
+              errorCode: 'E151000',
```
orius123, line 314: Use a different error code, so we can tell it is coming from here

nirgur: Which one?

asafshen: good point Eliya, `E` is used by Descope service,

in go-sdk Descope uses `G...` for generated errors (not sure g stands for generated or go btw)

image


maybe `G` or dedicated prefix for this (`J` for JS?)

nirgur: This is what @dorsha suggested in the issue
https://github.com/descope/etc/issues/7166#issuecomment-2249727456

dorsha: We can use J

nirgur: So `J151000` in both places?

orius123: `J151000` here and `J151001` in `packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts` if we don't have any error at all

nirgur: Done

orius123 (approved):

dorsha:
@nirgur build fails

orius123:
Don't we wamt to also release it?

nirgur:
> Cool
> 
> 1. Build fails
> 2. What about increasing the timeout and adding retries? Can be done in another pr

Let's talk about it first and decide how to handle

nirgur:
> Don't we wamt to also release it?

Released

PR #758, opened 2024-07-25
fix: nested params form not working in screen 

Bars92:
## Related Issues
fixes https://github.com/descope/etc/issues/6958

## Description
- fix screen dynamic key does not read nested values properly 

## Must

- [ ] Tests - missing

### DIFF
--- a/packages/sdks/web-component/src/lib/helpers/templates.ts
+++ b/packages/sdks/web-component/src/lib/helpers/templates.ts
@@ -127,9 +127,9 @@ const replaceHrefByDataType = (
 
 const setFormConfigValues = (
   baseEle: DocumentFragment,
-  formData: Record<string, string>,
+  formData?: Record<string, string>,
 ) => {
-  Object.entries(formData).forEach(([name, config]) => {
+  Object.entries(formData || {}).forEach(([name, config]) => {
     const eles = baseEle.querySelectorAll(`[name="${name}"]`);
 
     eles.forEach((ele) => {
--- a/packages/sdks/web-component/src/lib/helpers/templates.ts
+++ b/packages/sdks/web-component/src/lib/helpers/templates.ts
@@ -47,11 +47,24 @@ const replaceElementInputs = (
 /**
  * Get object nested path.
  * Examples:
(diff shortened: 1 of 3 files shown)


dorsha:
@Bars92 build fails

packages/sdks/web-component/src/lib/helpers/templates.ts
```diff
@@ -47,11 +47,24 @@ const replaceElementInputs = (
...
  * Get object nested path.
  * Examples:
- *  - getByPath({ { a { b: 'rob' } }, 'a.b') => 'hey rob'
+ *  - getByPath({ { a { b: 'rob' } }, 'a.b') => 'rob'
+ *  - getByPath({ { a { 'b.c': 'rob' } }, 'a.b.c') => 'rob'
  *  - getByPath({}, 'a.b') => ''
  */
 const getByPath = (obj: Record<string, any>, path: string) =>
-  path.split('.').reduce((prev, next) => prev?.[next] || '', obj);
+  {
```
nirgur, line 55: Why do we had to change it?
For which cases it wouldn't work?

Bars92: cases where the key may contain `.` or the screen state is flatten for example `{ "form": { "test.id": "value" } }`. seems like the screen values support that so i preferred to change the `getByPath` to support that as well

nirgur: Thanks for the clarification

I think it won't work for some objects,
e.g.
```
{a: {'b.c': {d: 1}}}

{'a.b': {'c.d': 1}}
```
Do we want to support it?

In general, I think there is a problem here, 
if the keys might contains dots, there are many possibilities for each received path, and I don't think we are covering them all

WDYT?

Bars92: you are right, although this is not possible because the screen state must contain the key first (such as form / client etc) but this handle internal screen state that is generate by the sdk so this internal logic can come from multiple places

nirgur: Not sure I understand
Do you think it's problematic?

PR #765, opened 2024-07-31
fix: issue 7219 RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/7219


dorsha:
@nirgur let's make sure to add test to catch this in the future

dorsha (approved):

PR #766, opened 2024-07-31
chore: Migrate angular RELEASE

nirgur:


asafshen (commented):
awesome work!

fyi noticed that the angular-sdk app triggers a 400/401 requests after login:
image

but this reproduces also in the previous repo, so I'll open separate issue

lets add `environment.development.ts` to gitignore and merge 👍

nirgur:
> awesome work!

Thanks!

> fyi noticed that the angular-sdk app triggers a 400/401 requests after login
> but this reproduces also in the previous repo, so I'll open separate issue

👍 

> lets add `environment.development.ts` to gitignore and merge 👍

Done

asafshen (approved):
great job!

PR #772, opened 2024-08-02
fix: fix multiple fetch user

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/7260

## Description
fix multiple fetch user when `useUser` happen multiple times
1. add ref on auth provider
2. consider that ref in hook

(this is similar to to approach that we have in `useSession`)


asafshen:
@nirgur, let's sit together on this tomorrow/Sunday due to the sensitivity of this code

nirgur (approved):

PR #779, opened 2024-08-15
fix: issue 7402 RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/7402


Bars92 (approved):

meirwah:
@nirgur @Bars92 where is the test here ?

nirgur:
There is no easy way to test it
Need to think what can be done in order to prevent such issues in the future

PR #783, opened 2024-08-21
feat: Allow dialect localization

orius123:
## Related Issues

Fixes https://github.com/descope/etc/issues/7388


packages/sdks/web-component/test/descope-wc.test.ts
```diff
@@ -71,7 +71,7 @@ const defaultOptionsValues = {
   client: {},
   redirectAuth: undefined,
   tenant: undefined,
-  locale: 'en',
```
orius123, line 74: Note that this changes what is being sent to the server from the browser navigate to also include the dialect

Bars92 (approved):

PR #785, opened 2024-08-22
fix: use prefix for last auth

Bars92:
## Related Issues

fixes https://github.com/descope/etc/issues/7423

## Related PRs
https://github.com/descope/console-app/pull/3365

## Description
- add storage prefix to last auth

## Must

- [x] Tests

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -227,10 +227,10 @@ class DescopeWc extends BaseDescopeWc {
     const redirectAuth =
       redirectAuthCallbackUrl && redirectAuthCodeChallenge
         ? {
-          callbackUrl: redirectAuthCallbackUrl,
-          codeChallenge: redirectAuthCodeChallenge,
-          backupCallbackUri: redirectAuthBackupCallbackUri,
-        }
+            callbackUrl: redirectAuthCallbackUrl,
+            codeChallenge: redirectAuthCodeChallenge,
+            backupCallbackUri: redirectAuthBackupCallbackUri,
+          }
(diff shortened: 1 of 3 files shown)


nirgur:
Maybe we should prefix all local storage keys with the project id?
In this way we will avoid such issues in the future, we won't need to add additional attributes to the web-component, and we will be more consistent

WDYT?

Bars92:
@nirgur we don't add any new attributes, we use the existing local storage prefix attribute which already prefix the session and refresh token so it should be more consistent that way. prefixing based on the project id is a great way but to make it consistent we need a broader change and some break on compatibility for all localstorage keys to make it ready. currently local storage prefix is not really used anywhere

nirgur:
> @nirgur we don't add any new attributes, we use the existing local storage prefix attribute which already prefix the session and refresh token so it should be more consistent that way. prefixing based on the project id is a great way but to make it consistent we need a broader change and some break on compatibility for all localstorage keys to make it ready. currently local storage prefix is not really used anywhere

We added this attribute to the web-component and the web-js-sdk 
Should this attribute be exposed to descopers? Do we expect them to set its value, or is it only for internal use?

If we agree that a better solution is to use the project-id, let’s consider how we can go with this approach and avoid increasing technical debt by using this attribute in more places

Bars92:
@nirgur this suppose to be internal property. Keep in mind that using project-id as a prefix will break backward compatibility and thus makes this solution much harder.. It's much easier to replace it all with project id and break compatibility all together.

nirgur:
> @nirgur this suppose to be internal property. Keep in mind that using project-id as a prefix will break backward compatibility and thus makes this solution much harder.. It's much easier to replace it all with project id and break compatibility all together.

Let's talk about it when you have time

PR #786, opened 2024-08-22
feat: add react sdk next release

Bars92:
## Related Issues
related https://github.com/descope/etc/issues/4454

## Related PRs
tbd

## Description
- add next release to react sdk to be consumed by the console app

### DIFF
--- a/.github/workflows/release-next.yml
+++ b/.github/workflows/release-next.yml
@@ -49,7 +49,7 @@ jobs:
           GITHUB_TOKEN: ${{ steps.get_token.outputs.token }}
       - name: Run pnpm publish
         run: |
-          pnpm --filter "@descope/react-sdk" npm version "0.0.0-next-$(git rev-parse HEAD)" --git-tag-version=false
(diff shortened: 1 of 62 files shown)


.github/workflows/release-next.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: pnpm install --frozen-lockfile --ignore-scripts
+        env:
+          CI: true
+      - name: Build
+        run: npm run version:ci
+        env:
+          CI: true
+          GITHUB_TOKEN: ${{ steps.get_token.outputs.token }}
+      - name: Run pnpm publish
+        run: pnpm -r publish --access=public --no-git-checks --tag next
```
nirgur, line 51: it will publish all packages, not just react
In addition, I think it's not running on affected packages only, but on all packages
Another thing is the version numbers, I'm not sure if it will generate the same version pattern we have for our next versions

Did you give it a try?

nirgur (approved):

PR #800, opened 2024-09-14
fix: Encode query params 

aviadl:
fixes https://github.com/descope/etc/issues/7714


asafshen (approved):

packages/sdks/core-js-sdk/src/httpClient/urlBuilder.ts
```diff
@@ -31,7 +31,7 @@ export const urlBuilder = ({
     url = `${url}?`;
     const keys = Object.keys(queryParams);
     keys.forEach((key: string, index: number) => {
-      url = `${url}${key}=${queryParams[key]}${
+      url = `${url}${key}=${encodeURI(queryParams[key])}${
```
nirgur, line 34: Do we decode it in the backend?

How can we tell if it's encoded or not?
Old sdks will not encode it

PR #804, opened 2024-09-23
feat(descope-ui-mixin): use dscpcdn.net

omercnet:
https://github.com/descope/etc/issues/7818

### DIFF
--- a/packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
+++ b/packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
@@ -7,10 +7,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
 
 export const UI_COMPONENTS_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://dscpcdn.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
 
 export const UI_COMPONENTS_FALLBACK_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://cdn.jsdelivr.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
 
 export const UI_COMPONENTS_URL_VERSION_PLACEHOLDER = '<version>';
--- a/packages/sdks/web-component/src/lib/constants/uiComponents.ts
+++ b/packages/sdks/web-component/src/lib/constants/uiComponents.ts
@@ -4,10 +4,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
 
 export const UI_COMPONENTS_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://descpcdn.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
 
 export const UI_COMPONENTS_FALLBACK_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://cdn.jsdelivr.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
 
 export const UI_COMPONENTS_URL_VERSION_PLACEHOLDER = '<version>';
--- a/packages/widgets/access-key-management-widget/README.md
+++ b/packages/widgets/access-key-management-widget/README.md
@@ -19,7 +19,7 @@ DESCOPE_WIDGET_ID=  # default: "access-key-management-widget"
 
 ```
 // replace x.x.x with the latest release of the widget: https://www.npmjs.com/package/@descope/access-key-management-widget
-<script src="https://static.descope.com/npm/@descope/access-key-management-widget@x.x.x/dist/index.js"></script>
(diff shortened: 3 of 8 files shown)


packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
```diff
@@ -7,10 +7,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
 
 export const UI_COMPONENTS_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://dscpcdn.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
```
asafshen, line 10: @omercnet why this is the main url? I thought that this should be the fallback (it definitely seems, to me at least, like a fallback / .net, the domain "looks" less reliable than descope

But maybe I am missing something 

Also, if customer now had csp that allows descope.com, won't it break something? Or at least they will have an error in the console log?

asafshen (commented):
Cool 😎 

Lmk what you think 


Also because this is a major change , if possible to make sure it doesn't break anything?

packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
omercnet: There's a fallback to old primary url so if for some reason the new fails it still uses the previous one in a non-breaking manner

Re testing, am I wrong to assume to the e2e cover this?

meirwah (changes requested):
instead of doing all at once

let's split it, start with applying CDN change to only 1 thing ! like widgets  (less production critical)
monitor for a while , only than apply to other (more critical componants)

also, aren't you removing a fallback here ? we said to keep all existing fallbacks, and add another fallback for new domain 

and we need to change domain IMO

packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
```diff
@@ -7,10 +7,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
 
 export const UI_COMPONENTS_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
+  'https://descopecdn.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
 
 export const UI_COMPONENTS_FALLBACK_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://cdn.jsdelivr.net/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
```
meirwah, line 14: we still want cdn.jsdelivr.net as fallback

so need to add UI_COMPONENTS_FALLBACK_URL2 for that

packages/sdks/web-component/src/lib/constants/uiComponents.ts
```diff
@@ -4,10 +4,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
 
 export const UI_COMPONENTS_URL =
   (IS_LOCAL_STORAGE && localStorage.getItem(UI_COMPONENTS_URL_KEY)) ||
-  'https://static.descope.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
```
meirwah, line 7: can we start this change ti impact on widget ONLY ? without flows ?

meirwah (changes requested):

meirwah:
@omercnet lint fails here

packages/libs/sdk-mixins/src/mixins/descopeUiMixin/constants.ts
```diff
@@ -2,14 +2,20 @@ import { IS_LOCAL_STORAGE } from '../../constants';
 
 export const DESCOPE_UI_SCRIPT_ID = 'load-descope-ui';
 export const DESCOPE_UI_FALLBACK_SCRIPT_ID = 'load-descope-ui-fallback-script';
+export const DESCOPE_UI_FALLBACK_2_SCRIPT_ID =
+  'load-descope-ui-fallback-script';
```
nirgur, line 6: We cannot have the same ID twice
```suggestion
  'load-descope-ui-fallback-script-2';
```

packages/sdks/web-component/src/lib/constants/uiComponents.ts
```diff
@@ -4,10 +4,10 @@ export const UI_COMPONENTS_URL_KEY = 'base.ui.components.url';
...
+  'https://descopecdn.com/npm/@descope/web-components-ui@<version>/dist/umd/index.js';
```
nirgur, line 7: I think this file is no longer relevant

packages/libs/sdk-mixins/src/mixins/descopeUiMixin/descopeUiMixin.ts
```diff
@@ -75,6 +77,12 @@ export const descopeUiMixin = createSingletonMixin(
               `Cannot load DescopeUI from fallback URL, Make sure this URL is valid and return the correct script: "${fallbackScriptEle.src}"`,
             ),
           );
+
+          // in case we could not load DescopeUI from the main URL, we are trying to load it from a fallback URL
+          this.#handleFallback2Script(
+            fallbackScriptEle[this.#errorCbsSym],
+            fallbackScriptEle[this.#loadCbsSym],
+          );
         });
```
nirgur, line 86: Instead of duplicating this code, we can reuse the `#handleFallbackScript` function 
We can add more parameters to it if needed

Let me know if you’d like to work on it together

nirgur (commented):
👍 
Added a few comments

meirwah:
@omercnet can you follow up on review here ?

meirwah:
@omercnet reminder to fix lint and comments here

nirgur (approved):

omercnet:
@meirwah  need your review too

meirwah:
> @meirwah need your review too

in our last call, we talked about also testing manually the fallbacks
using some VPN/proxy to block and simulate

@omercnet did you do that ?  after all the changes ?

meirwah (approved):

PR #815, opened 2024-09-30
feat: Add a new native action and state

itaihanski:
## Related Issues

https://github.com/descope/etc/issues/5992

## Related PRs
https://github.com/descope/orchestrationservice/pull/2787

## Description

The idea here is to add **native options** that will be provided via the mobile SDKs and piped to the BE via the `startOptions`. On the other side - a new `native` action is added and transfers responsibility of handling it to the mobile SDKs.
The design aims to make any future updates to take place in the BE, and in the mobile SDKs only, and leaving the `web-component` SDK untouched as much as possible. Hence, whenever a unique response is required by the mobile SDK, a dynamic payload is sent with a corresponding `type` field to allow the mobile SDKs to correctly parse and process it.

## Must

- [x] Tests


packages/sdks/core-js-sdk/src/sdk/types.ts
```diff
@@ -276,6 +281,8 @@ export type FlowResponse = {
     options: string;
     create: boolean;
   };
+  // native data - if the action is 'native'
+  native?: string;
```
asafshen, line 285: I don't quite relates to the term `native` 

Can you convince me why you chose this name as the action name ?

Also, what this string holds? A JSON string? If so? Shouldn't we get it as a such?

packages/sdks/core-js-sdk/src/sdk/types.ts
```diff
@@ -14,6 +14,11 @@ type RedirectAuth = {
   codeChallenge: string;
 };
 
+type NativeOptions = {
+  platform: string;
+  nativeOAuthProvider?: string;
```
asafshen, line 19: Maybe just OAuth provider? It's inside native context already

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -234,6 +248,9 @@ class DescopeWc extends BaseDescopeWc {
             backupCallbackUri: redirectAuthBackupCallbackUri,
           }
         : undefined;
+    const nativeOptions = native
+      ? { platform: native, nativeOAuthProvider }
```
asafshen, line 252: Correct me if I'm wrong- we can have emitter native oauth provider here, no?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -997,7 +1036,8 @@ class DescopeWc extends BaseDescopeWc {
           ...eleDescopeAttrs,
           ...formData,
           // 'origin' is required to start webauthn. For now we'll add it to every request
-          origin: window.location.origin,
+          origin:
+            this.flowState.current.originOverride || window.location.origin,
```
asafshen, line 1040: Where do origin override comes from ?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -420,6 +438,24 @@ class DescopeWc extends BaseDescopeWc {
       this.#handleSdkResponse(sdkResp);
     }
 
+    if (action === RESPONSE_ACTIONS.native) {
+      // prepare a callback to receive the response and failure from the Android layer
+      this.nativeComplete = async (input: any) => {
+        const sdkResp = await this.sdk.flow.next(
```
asafshen, line 444: Where/when do we call this fn?

asafshen (commented):
Great work 

Left some comments

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
shilgapira: Overridden by the Android SDK in case the native code does the WebAuthn instead of the webpage, in which case the origin value isn't the URL, but rather a signature of the running app

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
itaihanski: Not sure what you mean here, let's talk IRL so I can understand

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
itaihanski: The native layer will call this function when it is finished processing the payload received from the previous response

asafshen (approved):

PR #818, opened 2024-10-07
fix: handle mismatch flow version

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/7797


talaharoni:
@nirgur  I'm thinking about the OIDC scenario. the flow is starting with some query params. will they be used also where restarting the flow?

meirwah:
is this config turned on or off by default ?
i suggested to start off, suggest a specific customer to turn it on, see in logs if gets less version mismatch

nirgur:
> @nirgur I'm thinking about the OIDC scenario. the flow is starting with some query params. will they be used also where restarting the flow?

We are reseting only the 

> @nirgur I'm thinking about the OIDC scenario. the flow is starting with some query params. will they be used also where restarting the flow?

We are resetting only the `stepId` & `executionId` so we should still have it in the context,
But it worth checking this scenario to make sure it's working, the question is how we can do it?
It's not easy to reproduce this error

nirgur:
> is this config turned on or off by default ? i suggested to start off, suggest a specific customer to turn it on, see in logs if gets less version mismatch

Yes, it's on by default, do you think we should add another flag for it?
Just to be clear, the descoper will still get the error, the change is that now, in some cases, the flow will restart

meirwah:
these delicate client side logic tends to have some edge cases

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -366,14 +366,22 @@ const compareArrays = (array1: any[], array2: any[]) =>
 export const withMemCache = <I extends any[], O>(fn: (...args: I) => O) => {
```
orius123, line 366: Can you add a test to reset as well?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -610,6 +631,15 @@ class DescopeWc extends BaseDescopeWc {
         sdkResp?.error?.errorDescription || defaultMessage,
         sdkResp?.error?.errorMessage || defaultDescription,
       );
+
+      // E102004 = Flow requested is in old version
+      // E103205 = Flow timed out
+      if (
```
orius123, line 637: I think we should still log, but maybe just with warning if we are able to recover?

orius123: Also, do we want to have some way to disable this functionality? if the descoper want to handle it by themselves?

nirgur: We are dispatching an error in this case, this was not changed 
The only change is reloading the flow in some scenarios

Regarding the option to disable it, 
I think we have too many flags already, but we can add another one
Let's decide how we should call it and what should be the default value and i'll add it

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: done

nirgur:
We will add a flag for restarting the flow that will be disabled by default
I thought about calling it `restart-on-error`
Does anyone have a better name?

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -331,8 +331,12 @@ class BaseDescopeWc extends BaseClass {
     this.#componentsContext = { ...this.#componentsContext, ...e.detail };
   }
 
+  get isRestartOnError() {
+    return this.getAttribute('restart-on-error') === 'true';
```
orius123, line 335: We said default should be false, no?
@dorsha @talaharoni

dorsha: Yes, and we need to have good documentation what to do and how to handles these cases.

nirgur: The default value is indeed false
This is not the default value

orius123 (approved):

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -187,6 +187,27 @@ class DescopeWc extends BaseDescopeWc {
...
+      this.loggerWrapper.debug(
+        'Components version was not changed, restarting flow',
+      );
+      this.flowState.update({
+        stepId: null,
+        executionId: null,
+      });
+    } else {
+      this.loggerWrapper.error(
+        'Components version mismatch, please reload the page',
```
dorsha, line 206: @nirgur we said we will have callback in that case, no?

nirgur: I remember we talked about adding a flag to reload the flow
Which callback we should have in this case?

dorsha: What happens if component version  has changed as well? How can the descoper react to that?
Right now we only console log it, but they should have a way to react so they can refresh the browser and do whatever they want in that case.

dorsha: If they have a way to handle that error, let's document it.

nirgur: We are also dispatching an error 
If you believe the SDK needs an additional change to support this case, let’s define the expected behavior

PR #819, opened 2024-10-10
feat: send all flow versions during start

talaharoni:
## Related Issues

Related to https://github.com/descope/etc/issues/329

## Description
On flow start send versions of all flows to be stored on the context.

## Must

- [x] Tests


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -226,6 +226,15 @@ class DescopeWc extends BaseDescopeWc {
     const loginId = this.sdk.getLastUserLoginId();
     const flowConfig = await this.getFlowConfig();
     const projectConfig = await this.getProjectConfig();
+    const flowVersions = {
+      version: flowConfig.version,
+      componentsVersion: projectConfig.componentsVersion,
+      versions: Object.keys(projectConfig.flows || []).reduce((acc, key) => {
```
asafshen, line 232: why do we need other flows version here?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -226,6 +226,15 @@ class DescopeWc extends BaseDescopeWc {
...
+        const flowConf = projectConfig.flows[key];
```
asafshen, line 233: not critical at all, but I lov to use `ObjectEntries([key, value] => {}` that already get the matching value 
versions: Object.values(projectConfig.flows || []).reduce((acc, [key, flowConf) => {
...}, {})

packages/sdks/core-js-sdk/src/sdk/flow/types.ts
```diff
@@ -6,3 +6,9 @@ type JSONSerializable =
   | Array<JSONSerializable>;
 
 export type FlowInput = Record<string, JSONSerializable>;
+
+export type FlowVersions = {
+  version: number;
+  componentsVersion: string;
+  versions: Record<string, number>;
```
asafshen, line 13: maybe worth to name the attributes here `currentVersion` and `otherVersions`? your choice

asafshen (dismissed):
looks good! minor comments / question

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
talaharoni: When starting flow with subflows. we need to know the version for all participate flows.

asafshen: cool, maybe we should add something like `// pass other flow versions, that may be used as a part of the current flow`

asafshen (dismissed):

packages/sdks/core-js-sdk/src/sdk/flow/index.ts
```diff
@@ -31,6 +32,7 @@ const withFlow = (httpClient: HttpClient) => ({
           interactionId,
           version,
           componentsVersion,
```
orius123, line 34: Why do we send it twice?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -508,6 +520,7 @@ class DescopeWc extends BaseDescopeWc {
           interactionId,
           version,
```
nirgur, line 521: Do we need version as well?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -282,6 +293,7 @@ class DescopeWc extends BaseDescopeWc {
           '',
           flowConfig.version,
```
nirgur, line 294: Do we need to send it?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -226,6 +226,17 @@ class DescopeWc extends BaseDescopeWc {
     const loginId = this.sdk.getLastUserLoginId();
     const flowConfig = await this.getFlowConfig();
     const projectConfig = await this.getProjectConfig();
+    const flowVersions = {
```
nirgur, line 229: I would expect this to be in the BaseDescopeWc class

In addition, why do we need here also?
```
version: flowConfig.version,
componentsVersion: projectConfig.componentsVersion,
```

`componentsVersion` is sent as a separate param, 
and flow version should be already included in `allVersions`

nirgur (commented):
👍 
Added a few comments

packages/sdks/core-js-sdk/src/sdk/flow/index.ts
talaharoni: I wanted to keep it for backward compatibility so i can remove it in phases. 
But perhaps better to deliver the backend first and only after it's in production to deliver the sdk

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
talaharoni: i wanted in next phase to remove `version` and `componentsVersion` so they will be only in this object. 
But decided to remove now. 

I don't know in all places the current flowID so prefer to have the option to know it without that

packages/sdks/core-js-sdk/src/sdk/flow/index.ts
orius123: You forgot to remove the flow version `version?: number,`

talaharoni: removed.

orius123:
@nirgur why did the build fail?

orius123 (approved):

nirgur:
> @nirgur why did the build fail?

e2e tests are not stable

talaharoni:
> > @nirgur why did the build fail?
> 
> e2e tests are not stable

Any way to resolve it? actually it's pretty stable. fails every time :/

PR #828, opened 2024-10-26
fix: Add longish timeout for polling calls and a shorter one when detecting iOS throttling RELEASE

shilgapira:
## Related Issues
Fixes https://github.com/descope/etc/issues/8031

## Description
- This PR adds a hard timeout of `6 seconds` on the calls to `/next` in the enchanted link polling loop.
- This ensures that the polling always recovers after a reasonable time if network requests hang for whatever reason – bad routing, dropped packets, throttling (see below).
- It also adds a mechanism for detecting network request throttling in mobile Safari on iOS devices, possibly only on iOS 18 devices, to fail the poll early and retry as soon as possible.

## Throttling
When the flow is running in mobile Safari and the app is in the background, the `this.#pollingTimeout = setTimeout(...)` callback is not executed until the app returns to the foreground, but the network request that's executed straight after hangs for anywhere between 20-50 seconds.

If we do nothing then the request will hang for `6 seconds` until the timeout fires, and then the request will retry after a further `2 seconds`, so while this will technically "fix" the bug, the UX will be far from ideal as the user will be waiting for too long.

The throttle code tries to detect whether "too much" time has passed since the timer was scheduled, and if the webpage appears to be hidden, in which case it'll use a much lower timeout of `1 second` for the `/next` API call to fail early. If the request times out as expected, then another polling call is rescheduled to take place `500ms` later, instead of `2 seconds` as usual.

By the way, Apple has always had [problems in their HTTP layer](https://developer.apple.com/forums/thread/106838) with throttling network requests in the background.

## Alternatives
1. We could try to detect the throttling and delay the `/next` call rather than let it timeout, but this makes for a more brittle code as whatever delay we use might not be enough (I implemented this and around `100ms` was in fact not enough), and we'll just end up in the same place.

2. We could also add a `navigator.userAgent` check to try to detect that we're running on mobile Safari, but I wasn't 100% sure how to make it precise enough to not have false negatives.

3. We could also just always use a shorter timeout if `document.hidden` is `true`. This would save a lot of this complexity but might cause legitimate calls to fail if the network conditions are not ideal.

## Must
- [X] Tests

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -31,6 +31,7 @@ import {
   handleReportValidityOnBlur,
   getUserLocale,
   clearPreviousExternalInputs,
+  timeoutPromise,
 } from '../helpers';
 import { calculateConditions, calculateCondition } from '../helpers/conditions';
 import { getLastAuth, setLastAuth } from '../helpers/lastAuth';
@@ -615,19 +616,27 @@ class DescopeWc extends BaseDescopeWc {
   }
 
   #handlePollingResponse = (
-    executionId,
-    stepId,
-    action,
-    flowVersion,
-    componentsVersion,
+    executionId: string,
+    stepId: string,
+    action: string,
+    flowVersion?: number,
+    componentsVersion?: string,
+    rescheduled: boolean = false,
   ) => {
+    const pollingDefaultDelay = 2000;
+    const pollingDefaultTimeout = 6000;
+    const pollingThrottleDelay = 500;
+    const pollingThrottleThreshold = 500;
+    const pollingThrottleTimeout = 1000;
     if (action === RESPONSE_ACTIONS.poll) {
       // schedule next polling request for 2 seconds from now
       this.logger.debug('polling - Scheduling polling request');
+      const scheduledAt = Date.now();
+      const delay = rescheduled ? pollingThrottleDelay : pollingDefaultDelay;
       this.#pollingTimeout = setTimeout(async () => {
         this.logger.debug('polling - Calling next');
 
-        const sdkResp = await this.sdk.flow.next(
+        const nextCall = this.sdk.flow.next(
           executionId,
           stepId,
           CUSTOM_INTERACTIONS.polling,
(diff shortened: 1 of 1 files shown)


dorsha (approved):
Looks good to me.
Maybe need also @nirgur @asafshen and @itaihanski to review, since they touched this area several times.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -636,6 +645,41 @@ class DescopeWc extends BaseDescopeWc {
...
+        }
+
+        let sdkResp: Awaited<typeof nextCall>;
+        try {
+          const timeout = throttled
+            ? pollingThrottleTimeout
+            : pollingDefaultTimeout;
+          sdkResp = await timeoutPromise(timeout, nextCall);
+        } catch (err) {
+          this.logger.warn(
```
asafshen, line 667: can the promise may reject regardless the timeout, no? if server will return non 200 response then we may arrive here

I'm not recall if sever returns an  error throws or just return `resp.ok == false`

(if the former, then I think we can get infinite loop here)

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -636,6 +645,41 @@ class DescopeWc extends BaseDescopeWc {
           {},
         );
 
+        // Try to detect whether the tab is being throttled when running in a mobile browser, specifically on iOS.
+        // We check whether the tab seems to hidden and the polling callback was called much later than expected,
+        // in which case we allow a much shorter timeout for the polling request. The reschedule check ensures
+        // this cannot happen twice consecutively.
+        const throttled =
+          document.hidden &&
```
asafshen, line 653: if I understand this code properly, every request, in a hidden document, that will take more than 500ms will be considered as throttled, is that right?

asafshen (commented):
looks good, maybe we should have a short chat on this one as well

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
shilgapira: Not the request itself, but rather then scheduled callback being called > 500ms after it should have been called. So we schedule the `/next` call to happen 2 seconds from now, and it takes more than 2.5 seconds, then the condition will be true.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
shilgapira: I believe @nirgur changed the networking layer so that it always catches thrown errors and converts them to `resp.error`, etc.

shilgapira:
@asafshen We might want to consider whether the 4th alternative I listed might be good enough

PR #833, opened 2024-10-28
fix: support next build for app router without transpile   

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/7785

## Description
add build aliases that convert 

tested on next13/14 both app router and pages router


nirgur (approved):
Nice 💪

PR #838, opened 2024-11-03
fix: expose restartOnError on all sdks RELEASE

nirgur:
related: https://github.com/descope/etc/issues/7797


orius123 (commented):
Can you please also add to the readmes?

dorsha (approved):

PR #843, opened 2024-11-10
Support custom fields in user-profile-widget

Nitzperetz:
## Related Issues

Part of https://github.com/descope/etc/issues/6501

## Description

Widget Profile widget to support custom fields presentation - created a new mixin to support custom fields userAttribute component.

image


packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        Array.from(allCustomAttributesComponents).forEach((nodeEle) => {
+          const attrName = nodeEle.getAttribute('data-id');
+          const customAttrName = attrName.replace('customAttributes.', '');
+
+          const compInstance = new UserAttributeDriver(nodeEle, {
+            logger: this.logger,
+          });
+
+          const userCustomAttributesData = getUserCustomAttrs(this.state);
```
nirgur, line 30: There is no need to run the selector for every component, we should run it once

In addition, we should register to state changes, so whenever we will get a new state, in case the custom attributes are changed, the components will update

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,41 @@
...
+          compInstance.value = userCustomAttributesData[customAttrName];
```
nirgur, line 31: What if we do not have this attribute set for the user?
We will show a component with `undefined` value
I think in this case we should hide it

packages/sdks/react-sdk/tsconfig.json
```diff
@@ -17,5 +17,5 @@
     "noErrorTruncation": true,
   },
   "include": ["**/*.ts", "**/*.tsx"],
-  "exclude": ["node_modules", "build", "dist", "test", "examples"]
+  "exclude": ["node_modules", "build", "dist", "test"]
```
nirgur, line 20: This should be reverted

nirgur (commented):
Added some comments
If something is not clear we can go over it together

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
Nitzperetz: Right - in that case I would show an empty line (as the editor did add the field). and potentially the user could edit it.
I would change to empty string in that case.

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
Nitzperetz: good call about this line, changed it's scope.

Since custom attributes are in the stateManagementMixin I believe it is being observed. 
Correct me if I am wrong.

Nitzperetz:
@nirgur added subscribe to `initUserCustomAttributesMixin`

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,46 @@
...
+export const initUserCustomAttributesMixin = createSingletonMixin(
+  <T extends CustomElementConstructor>(superclass: T) =>
+    class UserCustomAttributesMixinClass extends compose(
+      stateManagementMixin,
+      loggerMixin,
+      initWidgetRootMixin,
+    )(superclass) {
+      customValueUserAttr: UserAttributeDriver;
+
+      #initCustomValueUserAttrs() {
```
nirgur, line 17: Now there is no need to call the selector inside the fn
you get the data in the fn params
In addition, we should wrap it with `withMemCache` 
See example [here](https://github.com/descope/descope-js/blob/f159ae8ebd6ac37c9e707252f5bf16e73b144171/packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initAvatarMixin.ts#L74)

If something is not clear we can do it together

Nitzperetz: @nirgur , tested it just now and seems like I must have the first call (otherwise the values are not being presented).

Added the caching, please check.

Nitzperetz:
@nirgur , ready

nirgur (approved):

Nitzperetz:
I will be adding flows support in this PR. let's keep it open for now and not merge.

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,144 @@
...
+            baseStaticUrl: this.baseStaticUrl,
+          }),
+        );
+        this.#deleteFlows[flowId]?.onSuccess(() => {
+          this.#deleteModals[flowId]?.close();
+          this.actions.getMe();
+        });
+      }
+
+      #updateCustomValueUserAttrs = withMemCache(
```
nirgur, line 67: This function is a bit big and contains some repetitive code
Do you think we could split it into smaller functions?

nirgur (commented):
Looks good 👍 
Added one comment

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
Nitzperetz: It was really about keeping the same standards as the other mixin - to begin with a preferred extracting code to functions and reuse.
Let me see what I can do here

Nitzperetz:
@nirgur ready with new API call to custom fields.
image

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export const initUserCustomAttributesMixin = createSingletonMixin(
+  <T extends CustomElementConstructor>(superclass: T) =>
+    class UserCustomAttributesMixinClass extends compose(
+      stateManagementMixin,
+      loggerMixin,
+      initWidgetRootMixin,
+      modalMixin,
+    )(superclass) {
+      customValueUserAttr: UserAttributeDriver;
```
nirgur, line 25: Are we using it somewhere?

packages/widgets/user-profile-widget/src/lib/widget/state/selectors.ts
```diff
@@ -17,3 +18,33 @@ export const getIsPhoneVerified = createSelector(
...
 export const getHasPasskey = createSelector(getMe, (me) => me.webauthn);
 export const getHasPassword = createSelector(getMe, (me) => me.password);
+
+export const getCustomAttributes = (state: State) =>
+  state.customAttributes.data;
+
+export const getUserCustomAttrs = createSelector(
+  getMe,
+  getCustomAttributes,
+  (userData, allCustomAttrs = []) => {
```
nirgur, line 28: Missing type

packages/widgets/user-profile-widget/src/lib/widget/state/selectors.ts
```diff
@@ -17,3 +18,33 @@ export const getIsPhoneVerified = createSelector(
...
+    const res: Record<string, string> = {};
+    const userCustomAttributes = userData['customAttributes'] || {};
+
+    Object.keys(userCustomAttributes).forEach((key: string) => {
```
nirgur, line 32: I think that in this case, it will be better to use reduce on Object.entries

```
Object.entries(userCustomAttributes).reduce((acc, [key, val]) => { ...
```

packages/widgets/user-profile-widget/jest.config.js
```diff
@@ -11,8 +11,8 @@ module.exports = {
     global: {
       branches: 0,
       functions: 12,
-      lines: 38,
-      statements: 37,
+      lines: 36,
+      statements: 36,
```
nirgur, line 15: Can we add a test for the API call?

packages/widgets/user-profile-widget/src/lib/widget/api/types.ts
```diff
@@ -4,7 +4,7 @@ export type Sdk = ReturnType<typeof createWebSdk>;
 
 type CustomAttributeType = string | boolean | number;
 
-type CustomAttributes = Record<string, CustomAttributeType>;
+export type CustomAttributes = Record<string, CustomAttributeType>;
```
nirgur, line 7: Are we using it outside this file?

packages/widgets/user-profile-widget/src/lib/widget/api/sdk/createUserSdk.ts
```diff
@@ -21,7 +21,21 @@ export const createUserSdk = ({
     return res.json();
   };
 
+  const getCustomAttributes = async (): Promise<CustomAttr[]> => {
+    if (mock) {
+      return [];
```
nirgur, line 26: Don't we want to add mock for it?

nirgur (commented):
Looks good
Added a few comments

packages/widgets/user-profile-widget/src/lib/widget/api/sdk/createUserSdk.ts
Nitzperetz: We can't, as we don't know which would be the user's custom fields.

Nitzperetz:
Updated PR, addressed all comments.
you can re-review.
Please don't merge, as this PR is pending managementservice changes.

Nitzperetz:
@nirgur , updated without the API

Nitzperetz:
@nirgur I get Application-widget-portal test issue, which must be unrelated to this PR..
https://cloud.nx.app/runs/mYDeHCyByI

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initUserCustomAttributesMixin.ts
```diff
@@ -0,0 +1,169 @@
...
+              // to full date time
+              compInstance.value = new Date(val).toLocaleString();
+            } else if (
+              type === AttributeTypeName.BOOLEAN &&
+              val !== undefined
+            ) {
+              compInstance.value = !val ? 'False' : 'True';
+            } else {
+              compInstance.value = (val || '').toString();
+            }
```
nirgur, line 94: Consider extrcting this logic into a separate fn
`getFormattedValue(val, type) => string`

packages/widgets/user-profile-widget/jest.config.js
nirgur: You marked it as resolved but it seems there are no new tests

Nitzperetz:
@nirgur updated

packages/widgets/user-profile-widget/jest.config.js
Nitzperetz: as there is not API :)

nirgur (dismissed):

Bars92 (approved):

PR #856, opened 2024-12-04
feat: Add thirdPartyAppStateId and scopes parameters for third party application

guyp-descope:
Related to https://github.com/descope/etc/issues/7890


packages/sdks/core-js-sdk/src/sdk/types.ts
```diff
@@ -335,6 +335,8 @@ export type Options = {
   oidcPrompt?: string;
   oidcErrorRedirectUri?: string;
   nativeOptions?: NativeOptions;
+  thirdPartyAppStateId?: string;
+  scopes?: string;
```
nirgur, line 339: Shouldn’t we give it a more meaningful name? 
It might be difficult to understand the purpose of this parameter later on

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -504,6 +534,7 @@ export const showFirstScreenOnExecutionInit = (
     oidcLoginHint,
     oidcPrompt,
     oidcErrorRedirectUri,
+    scopes,
```
nirgur, line 537: Don't we need to send the state as well?

packages/sdks/core-js-sdk/src/sdk/types.ts
guyp-descope: Yea, I missed that.. adding now.

packages/sdks/web-component/src/lib/helpers/helpers.ts
guyp-descope: fixed

guyp-descope:
@nirgur please review again, I fixed your comment + move the thirdPartyAppId to be inside the SSOQueryParams object.

packages/sdks/core-js-sdk/src/sdk/types.ts
nirgur: My concern is that in a few months from now, when we will want to refactor the web-component, it will be hard to tell which params are related to which feature

nirgur: At least let's add a comment that this param is related to the third party apps

guyp-descope: Renamed and added comment as per your request (I understood its important for you so..:) ) 
please review.

nirgur (approved):

PR #857, opened 2024-12-07
fix: add setters for wc to work well with react 19

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/8299
## Description

The error occurs in React 19 due to changes in how React handles Custom Elements and their properties.

before (18) React was more lenient when setting properties on Custom Elements. If a property had only a getter and no setter, React would often silently fail without throwing an error.
In React 19 - react has improved its support for Custom Elements and aligns with the HTML specification. When React attempts to set a property on a Custom Element, and the property has only a getter (no setter), it throws a TypeError since it violates JavaScript’s property descriptor rules

some sources:
https://blog.logrocket.com/working-custom-elements-react
https://github.com/facebook/react/issues/29037
https://codingmall.com/knowledge-base/25-global/19398-how-does-react-19-support-custom-elements-differently-from-previous-versions

another approach will be also to change getters to internal (`_getTheme`) and avoid setters, might work as well


asafshen:
closingin favor of https://github.com/descope/descope-js/pull/860

PR #860, opened 2024-12-14
fix: support react-19

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/8299
## Description

The error occurs in React 19 due to changes in how React handles Custom Elements and their properties.

before (18) React was more lenient when setting properties on Custom Elements. If a property had only a getter and no setter, React would often silently fail without throwing an error.
In React 19 - react has improved its support for Custom Elements and aligns with the HTML specification. When React attempts to set a property on a Custom Element, and the property has only a getter (no setter), it throws a TypeError since it violates JavaScript’s property descriptor rules

some sources:
https://blog.logrocket.com/working-custom-elements-react
https://github.com/facebook/react/issues/29037
https://codingmall.com/knowledge-base/25-global/19398-how-does-react-19-support-custom-elements-differently-from-previous-versions

another approach will be also to change getters to internal (`_getTheme`) and avoid setters, might work as well


packages/sdks/react-sdk/src/components/WebComponentBridge.tsx
```diff
@@ -0,0 +1,78 @@
...
+          if (
+            key.endsWith(propSuffixes.attr) ||
+            key.endsWith(propSuffixes.prop)
+          ) {
+            return acc;
+          }
+          return { ...acc, [kebabCase(key)]: value };
+        }, {}),
+      [props],
+    );
```
nirgur, line 43: What about transforming the props before, so the CB fn will be much simple? 
```suggestion
  const compProps = useMemo(() => {
  // can be extracted into helper file
    const transformKey = (key) => {
      const [, trimmedKey, category] = /(.*)\.(prop|attr)$/.exec(key) || [
        ,
        key,
        'rest',
      ];

      return { trimmedKey, category };
    };
    
  // can be extracted into helper file
    const transformAttrValue = (value) =>
      typeof value === 'string' || value == null
        ? value
        : JSON.stringify(value);

    return Object.entries(props).reduce(
      (acc, [key, value]) => {
        const { trimmedKey, category } = transformKey(key);
        if (category === 'prop') acc.prop.push([trimmedKey, value]);
        else if (category === 'attr')
          acc.attr.push([kebabCase(trimmedKey), transformAttrValue(value)]);
        else Object.assign(acc.rest, { [trimmedKey]: value })

        return acc;
      },
      { attr: [], prop: [], rest: {} },
    );
  }, [props]);

```

packages/sdks/react-sdk/src/components/WebComponentBridge.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    const currRef = useRef<HTMLElement | null>(null);
+
+    const setInnerRef = useCallback(
+      (innerRef) => {
+        currRef.current = innerRef;
+        if (innerRef) {
+          Object.entries(props).forEach(([key, value]) => {
+            if (key.endsWith(propSuffixes.prop)) {
+              if (value !== undefined && value !== null) {
```
nirgur, line 53: Why we cannot reset the prop?

packages/sdks/react-sdk/src/components/UserProfile.tsx
```diff
@@ -66,10 +45,16 @@ const UserProfile = React.forwardRef<HTMLElement, UserProfileProps>(
...
-          innerRef={setInnerRef}
-          theme={theme}
           styleId={styleId}
-          debug={debug}
+          ref={setInnerRef}
+          {...{
+            // attributes
+            'theme.attr': theme,
+            'debug.attr': debug,
+            'style-id.attr': styleId,
```
nirgur, line 54: You can use camelCase and it's also duplicate

packages/sdks/react-sdk/src/components/UserProfile.tsx
```diff
@@ -66,10 +45,16 @@ const UserProfile = React.forwardRef<HTMLElement, UserProfileProps>(
           widgetId={widgetId}
           baseUrl={baseUrl}
           baseStaticUrl={baseStaticUrl}
```
nirgur, line 47: What about these props? don't we want to use the `.attr`?

nirgur (commented):
Looks good
added some comments

packages/libs/sdk-mixins/src/mixins/createValidateAttributesMixin/createValidateAttributesMixin.ts
```diff
@@ -33,12 +33,11 @@ const createValidateAttributesMixin =
       }
 
       async init() {
-        await super.init?.();
-
         // check attributes initial values
         mappingsNames.forEach((attr) =>
           this.#handleError(attr, this.getAttribute(attr)),
         );
+        await super.init?.();
```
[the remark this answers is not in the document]
asafshen: we concluded together that its is better to have it here, but we can do this in a different PR

packages/sdks/react-sdk/src/components/AccessKeyManagement.tsx
```diff
@@ -1,60 +1,31 @@
...
-  lazy,
-  Suspense,
-  useEffect,
-  useImperativeHandle,
-  useState,
-} from 'react';
+import React, { lazy, Suspense, useImperativeHandle, useState } from 'react';
 import Context from '../hooks/Context';
 import { AccessKeyManagementProps } from '../types';
+import WebComponentBridge from './WebComponentBridge';
```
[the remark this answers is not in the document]
asafshen: good idea, changed 👍

packages/sdks/react-sdk/src/types.ts
```diff
@@ -14,7 +14,7 @@ import UserProfileWidget from '@descope/user-profile-widget';
 import ApplicationsPortalWidget from '@descope/applications-portal-widget';
 import createSdk from './sdk';
 
-declare global {
+declare module 'react' {
```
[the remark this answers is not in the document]
asafshen: yes, tested on react 18 (and next 13/14 FYI)

packages/sdks/react-sdk/src/components/WebComponentBridge.tsx
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * - Props ending in `.prop` are set as properties on the web component.
+ * - Props ending in `.attr` are set as attributes on the web component (kebab-case).
+ * - All other props are set as props
+ *
+ * This resolves attribute/property behavior differences in React 19.
+ *
+ * @see https://github.com/facebook/react/issues/29037
+ */
+const WebComponentBridge = <P extends Record<string, any>>(
```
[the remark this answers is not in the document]
asafshen: image

i think its needed fro this, it helps TS to infer type of props

packages/sdks/react-sdk/src/components/UserProfile.tsx
asafshen: we can, but it is not needed, they are not referred as such (because the name is different because the camelCase vs kebab-case
image

packages/sdks/react-sdk/src/components/WebComponentBridge.tsx
asafshen: thanks!

asafshen: great suggestions, thanks for that!

packages/sdks/react-sdk/src/components/WebComponentBridge.tsx
asafshen: fixed

packages/sdks/react-sdk/src/components/UserProfile.tsx
nirgur: Might be confusing, your call

nirgur (approved):

nirgur (approved):

PR #861, opened 2024-12-16
chore: Bundle size

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/6291

reduced React SDK bundle size from `455KB` to `115KB` (`112KB` to `36KB` gzipped)

A new entry in React SDK was added in order to expose only flow related stuff: `@descope/react-sdk/flows`
e.g.: `import { AuthProvider, Descope } from '@descope/react-sdk/flows';`


packages/sdks/react-sdk/tsconfig.json
```diff
@@ -11,11 +11,10 @@
...
     "isolatedModules": true,
-    "declaration": true,
-    "declarationDir": "dts",
+    "declaration": false,
     "jsx": "react",
-    "noErrorTruncation": true,
+    "noErrorTruncation": true
   },
   "include": ["**/*.ts", "**/*.tsx"],
-  "exclude": ["node_modules", "build", "dist", "test", "examples"]
```
orius123, line 20: 👏

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -350,7 +350,7 @@ class BaseDescopeWc extends BaseClass {
   }
 
   async #handleDebugMode({ isDebug }) {
-    if (isDebug) {
+    if (isDebug && process.env.NODE_ENV === 'development') {
```
orius123, line 353: Need to doc somehow
And log a warning if debug is true but env is not set to development
```suggestion
    if (isDebug && process.env.NODE_ENV !== 'production') {
```

packages/sdks/react-sdk/src/flows.ts
```diff
@@ -0,0 +1,25 @@
...
+  getRefreshToken,
+  getSessionToken,
+  isSessionTokenExpired,
+  isRefreshTokenExpired,
+} from './sdk';
+
+// Export ref to baseHeaders so it can be overridden
+export { baseHeaders } from './constants';
+
+export type { ILogger } from './types';
```
asafshen, line 25: imho - having all non-flows here (such as hooks, utils, types etc) is a bit problematic in the long run 

if we would want to add more "chunks" in the future 

few use cases that I can think of - 
- lets say someone want only widgets - they will need also ILogger, and maybe hooks
- maybe someone will want only hooks, and not to import flows


I don't think we should go and do this now ofc, but I would start with a dedicate file `common`/`shared` that all the non flow things will go there if that is possible

asafshen (commented):
very cool

@nirgur I would also add a line in react-sdk readme about this topic, perhaps something like 

```
## Performance / Bundle Size

To improve modularity and reduce bundle size, all flow-related utilities are avaialble in `@descope/react-sdk/flows` subpath.

```
### Migration Guide
If you were previously importing flow utilities like this:

```js
import { Descope, useSession, ... } from '@descope/react-sdk/flows';
```
```

File: packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts [hunk 4]
```diff
@@ -67,6 +67,10 @@ class DescopeWc extends BaseDescopeWc {
     BaseDescopeWc.sdkConfigOverrides = config;
   }
 
+  static get sdkConfigOverrides() {
```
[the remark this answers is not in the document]
nirgur: There is a test in one of the frameworks that verifies the sdkConfigOverrides is updated 
However, since we do not have a getter, the test cannot retrieve the updated value 
I’m not sure how it was working in the first place

packages/sdks/react-sdk/tsconfig.json
```diff
@@ -11,11 +11,10 @@
...
+  "exclude": ["node_modules", "build", "dist", "test"]
```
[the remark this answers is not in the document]
nirgur: No, we do not need to exclude it anymore
I fixed it as part of this PR

orius123 (approved):

packages/sdks/react-sdk/src/flows.ts
nirgur: The internal structure is our call, but what’s important are the entry points we expose

Right now, we have the default entry and a `/flows` entry 
Are you thinking we need more entry points at this stage?

IMO, we should balance between adding optimizations and not overcomplicating things with too many entry points Plus, all widgets need to authenticate first, and hooks don’t really make sense without authentication

Let's have a quick sync on it if needed

asafshen: I just noting that from a file structure perspective perspective, I think its better to put all those things in separate file, thats all

asafshen (approved):

PR #862, opened 2024-12-16
feat: Support for isDismissed from FedCM

gaokevin1:
## Related Issues

Fixes https://github.com/descope/etc/issues/8619

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| service a PR | Link to PR |
| service b PR | Link to PR |

## Description

Add the ability to extract [Dismissed Event and Reason](https://developers.google.com/identity/gsi/web/reference/js-reference#PromptMomentNotification) from the FedCM component. 

This will allow you to be able to see the reason behind why a OneTap modal was skipped, whether it was induced by the user or whether the lack of a session triggered it to not show up.

This is important for analytics and tracking whether or not the OneTap component showed up because the user wasn't signed in with Google, or because the user **actually** closed the modal but a valid credential was found.

## Must

- [ ] Tests


packages/sdks/web-js-sdk/src/sdk/fedcm.ts
```diff
@@ -100,6 +100,8 @@ type OneTapInitialize = ({
 
 type PromptNotification = {
   isSkippedMoment: () => boolean;
+  isDismissedMoment: () => boolean;
+  getMomentType: () => string;
```
gaokevin1, line 104: Do you think we should use this instead of just including `isDismissedMoment`? @orius123 

Maybe we can do a switch based on the value from `getMomentType` instead like in my first commit?

image

gaokevin1:
This is still useful I think, but for GoodRx we found another workaround so I will close this PR for now.

orius123 (dismissed):

packages/sdks/web-js-sdk/src/sdk/fedcm.ts
```diff
@@ -100,6 +100,9 @@ type OneTapInitialize = ({
 
 type PromptNotification = {
   isSkippedMoment: () => boolean;
+  isDismissedMoment: () => boolean;
```
asafshen, line 103: should we document this?

meirwah:
@gaokevin1 lint fails

packages/sdks/web-js-sdk/src/sdk/fedcm.ts
gaokevin1: @asafshen we should document it, but I don't think that documentation resides in the SDK right? 

This PR will open a doc issue when merged which I will take care of.

asafshen: I'm not sure what is the case for fedCM, but usually we document things in the SDK as well
as a developer I prefer github readme docs and not product website docs, but I see that here we don't have it anyway so probably a different concern 👍

gaokevin1:
@orius123 @asafshen can you please review again?

orius123 (approved):

PR #866, opened 2024-12-18
feat(web-component): prefer static from api base url

omercnet:
## Related Issues

Fixes https://github.com/descope/etc/issues/8253


omercnet:
done in https://github.com/descope/descope-js/pull/871

PR #868, opened 2024-12-19
fix multiple flows on the same page

nirgur:


packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -109,10 +114,17 @@ export function getAnimationDirection(
...
 }
 
-export const getRunIdsFromUrl = () => {
-  const [executionId = '', stepId = ''] = (getFlowUrlParam() || '').split('_');
+export const getRunIdsFromUrl = (flowId: string) => {
+  let [executionId = '', stepId = ''] = (getFlowUrlParam() || '').split('_');
+  const executionFlowId = getFlowIdFromExecId(executionId);
+
+  // if the flow id does not match, this execution id is not for this flow
+  if (executionFlowId && executionFlowId !== flowId) {
```
dorsha, line 122: Let's also add if `flowId` exists

packages/sdks/web-component/test/helpers/index.test.ts
```diff
@@ -59,8 +59,12 @@ describe('helpers', () => {
       writable: true,
       value: new URL('http://localhost'),
     });
-    window.location.search = `?${URL_RUN_IDS_PARAM_NAME}=8_9`;
-    expect(getRunIdsFromUrl()).toEqual({ executionId: '8', stepId: '9' });
+    window.location.search = `?${URL_RUN_IDS_PARAM_NAME}=8|#|a_9`;
+    expect(getRunIdsFromUrl('8')).toEqual({
```
dorsha, line 63: let's also add test in case of empty flow id, and cases with wrong location.search query param

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -57,6 +57,11 @@ function resetUrlParam(paramName: string) {
   }
 }
 
+const getFlowIdFromExecId = (executionId: string) => {
+  const regex = /(.*)\|#\|.*/;
+  return regex.exec(executionId)?.[1];
```
dorsha, line 62: can it fail if length is 0 or 1?

packages/sdks/web-component/test/helpers/index.test.ts
nirgur: done

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: AFAIR if there is no flow id there is an error and the SDK is not running
Added it anyway

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: No, the regex.exec return an array or null
Changed to return an empty string in this case

dorsha (approved):

PR #876, opened 2024-12-29
feat: Country Codes package

tomerlichtash:
## Related Issues

Related https://github.com/descope/etc/issues/8022

## Description

Currently, Phone Country Codes are consumed by the ConsoleApp from the Phone Component.

As part of the Components Versioning effort, we want to separate the content consumption from the component, as well as create a single source of truth of our country code list.

## Must

- [ ] Tests

### DIFF
--- a/packages/libs/country-codes/.eslintrc.json
+++ b/packages/libs/country-codes/.eslintrc.json
@@ -1,0 +1,90 @@
+{
+  "root": true,
+  "env": {
+    "browser": true,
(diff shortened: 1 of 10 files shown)


tomerlichtash:
@nirgur Since the console-app consumes this directly from WCUI, I think it'll be better to separate it instead of chaining it down through flow-components and the dynamic module, so we can consume only types from the dynamic module, and not content.

WDYT?

PR #878, opened 2024-12-30
fix: return sdk script results after conditional UI

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/8612

## Description
return sdk script results after conditional UI (passkeys)

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -517,6 +517,7 @@ class DescopeWc extends BaseDescopeWc {
         flowConfig.version,
         projectConfig.componentsVersion,
         {
+          ...this.getComponentsContext(),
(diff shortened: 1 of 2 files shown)


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur, line 1: I’m not sure I understand the problem and how this change solves it
Can you please elaborate?

asafshen: the problem is that after biometrics auto-fill (which trigger this `next(...)` call) - sdk scripts (such as forter/fingerprint) don't work

the reason that they don't work is because we don't send their state like on other "regular" next calls
 
@nirgur if you prefer we can talk

shilgapira:
@asafshen @nirgur Did we forget this PR or is it not needed anymore?

Issue #884, opened 2025-01-08
Session not working in nextjs 15


rutmus:
I've been trying to use the session to access the jwt props but it seems to not work well due to incompatibilities with next 15 

<img width="1028" alt="image" src="https://github.com/user-attachments/assets/8f36c864-c725-4336-ab98-73eec821136f" />

I saw you've already referenced this but I think that you will need to change the session to async or export sessionPromise in addition to make next 15 happy

https://github.com/descope/descope-js/blob/656acd8dc3618a3887f448917f88c9b48671db6a/packages/sdks/nextjs-sdk/src/server/session.ts#L28

allenzhou101:
Taking a look!

allenzhou101:
Reproduced and we're working on a fix.

asafshen:
hey @rutmus 

as a workaround to this issue - you can use `getSession` to extract the descope session (I created `tempSession` to demonstrate that )

```
import { getSession } from "@descope/nextjs-sdk/server";
import { headers } from "next/headers";

// tempo
async function tempSession() {
  const currHeaders = await headers();
  const req = {
    headers: {
      'x-descope-session': currHeaders.get('x-descope-session'),
    }
  }
  return getSession(req as any);
}

export default async function Page() {
  // get current request
  const currSession = await tempSession();
  ```

asafshen:
@afogel thanks for the patience, we are working on a fix

I'll keep you posted here of course!

afogel:
any update on this @asafshen , this is particularly noisy in our logs and we'd strongly prefer not to have to use a hack in order to reduce noise

asafshen:
@afogel I completely understand , We'll prioritise this issue, hopefully in the next day or two

PR #885, opened 2025-01-08
fix(nextjs-sdk): add support for NextJS 15

allenzhou101:
## Related Issues

Fixes https://github.com/descope/descope-js/issues/884

## Description

Migrates the SDK to NextJS 15

## Must

- [ ] Tests


packages/sdks/nextjs-sdk/src/server/session.ts
```diff
@@ -24,10 +20,8 @@ const extractSession = (
 };
 // returns the session token if it exists in the headers
 // This function require middleware
-export const session = (): AuthenticationInfo | undefined => {
-	// from Next.js 15, headers() returns a Promise
-	// It can still be used synchronously to facilitate migration
-	const reqHeaders = headers() as never as HeaderTypes;
+export const session = async (): Promise<AuthenticationInfo | undefined> => {
```
asafshen, line 23: Think break BC 
Also for someone that is not using next 15

Let's think about how to achieve that breaking

asafshen (commented):
Cool 

Let's see how to solve this issue without breaking BC

packages/sdks/nextjs-sdk/src/server/session.ts
allenzhou101: Seems to maintain backwards compatibility and access the session synchronously server-side, we would need to switch off of headers, cookies, or any other API from NextJS that is now asynchronous. So doing the same thing where we set the session into whatever storage mechanism (like how it's headers right now) in middleware and accessing via session function.

Options are:
- File-Based Storage (https://www.npmjs.com/package/@types/better-sqlite3)
- Different cookie library that's synchronous (https://www.npmjs.com/package/cookie)

Thoughts on this? At some point we'll probably have to bite the bullet and make the session call async, but maybe once NextJS fully moves off?

PR #887, opened 2025-01-13
fix: handle close timeouts

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/8912
Fixes https://github.com/descope/etc/issues/7424

## Description

handle too close timeouts in the following ways
 - after auth request - if calculated timeout is less than 20 seconds - do not set refresh timeout
 - when window becomes visible - do not refresh if refresh failed


asafshen:
closed in favor of https://github.com/descope/descope-js/pull/1048

PR #888, opened 2025-01-19
fix: trying to fix polling error

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/8986

@shilgapira @itaihanski 
I'm not sure if this change breaks something that related to the way mobile devices are handling the polling requests

Can you please review?
(if this fix is ok, I will fix the failed test)

Thanks

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -724,15 +724,19 @@ class DescopeWc extends BaseDescopeWc {
           return;
         }
 
+        this.#handleSdkResponse(sdkResp);
+
         this.logger.debug('polling - Got a response');
+
         if (sdkResp?.error) {
           this.logger.debug(
             'polling - Response has an error',
             JSON.stringify(sdkResp.error, null, 4),
           );
+
+          return;
         }
 
-        this.#handleSdkResponse(sdkResp);
         // will poll again if needed
         this.#handlePollingResponse(
           executionId,
--- a/packages/sdks/web-component/test/descope-wc.test.ts
+++ b/packages/sdks/web-component/test/descope-wc.test.ts
@@ -2075,12 +2075,12 @@ describe('web-component', () => {
             expect.any(Object),
           ),
         {
-          timeout: 8000,
+          timeout: 20000,
         },
       );
 
       await waitFor(() => expect(clearTimeout).toHaveBeenCalled(), {
-        timeout: 8000,
+        timeout: 20000,
       });
     });
 


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -724,15 +724,19 @@ class DescopeWc extends BaseDescopeWc {
           return;
         }
 
+        this.#handleSdkResponse(sdkResp);
```
itaihanski, line 727: It's better to leave the handle response after the error check. The SDK relies on the "onError" callback for fatal errors

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -724,15 +724,19 @@ class DescopeWc extends BaseDescopeWc {
...
+
         this.logger.debug('polling - Got a response');
+
         if (sdkResp?.error) {
           this.logger.debug(
             'polling - Response has an error',
             JSON.stringify(sdkResp.error, null, 4),
           );
+
+          return;
```
itaihanski, line 737: I think we should use the onError callback at this point instead of simply stopping the polling without a way to continue.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Why does it matter?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: But we need to stop the polling on error because otherwise we have an endless error loop (see the issue related to this PR)
Maybe in case of a specific error (mobile BG related error) we can continue polling

WDYT?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
itaihanski: There's no use sending an error polling response to the native SDK, it serves no purpose.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
itaihanski: I think this should be done regardless of native. If we stop polling - it should be considered a fatal error, because there's no way to recover at this point.

nirgur:
@itaihanski Do we want to merge this change or we are waiting for the server side changes?

itaihanski:
I think we should solve the real issue

PR #1011, opened 2025-02-09
fix: add cjs build fir nextjs sdk

asafshen:
## Related Issues

fixes https://github.com/descope/etc/issues/9233

## Description

add cjs build for next-sdk


packages/sdks/nextjs-sdk/rollup.config.mjs
```diff
@@ -65,45 +65,77 @@ const configurations = ['server', 'client', ''].map((entry) => {
...
-		output: [{ file: outputFile, format: 'esm' }],
-		plugins: [
-			dts(),
-			del({
-				hook: 'buildEnd',
-				targets: srcDir
-			})
-		]
-	};
+const endConfigurations = ['server', 'client', ''].flatMap((entry) => {
```
nirgur, line 96: Do you think we can simplify the config and include it somehow in the config above?
Why do we need to split the configs and merge it later?

nirgur (commented):
Looks good
added a few comments

nirgur (approved):

PR #1012, opened 2025-02-09
feat: Custom screens support RELEASE

nirgur:
This PR brings custom screens support to the flows SDKs
It introduce a new function `onScreenUpdate` (more details in the readme)

It's dependent on [this](https://github.com/descope/orchestrationservice/pull/3145/) PR to be deployed so we will get also the first screen name

## Related Issues

Related https://github.com/descope/etc/issues/570


packages/sdks/web-component/src/lib/constants/customScreens.ts
```diff
@@ -0,0 +1,8 @@
+export const EXCLUDED_STATE_KEYS = [
```
nirgur, line 1: We need to decide if we want to add/remove keys from this list

shilgapira:
I'll review later today 👍👍

packages/sdks/react-sdk/src/components/Descope.tsx
```diff
@@ -42,7 +42,7 @@ const DescopeWC = lazy(async () => {
   return {
     default: withPropsMapping(
       React.forwardRef<HTMLElement>((props, ref) => (
-        <descope-wc ref={ref} {...props} />
+	<descope-wc ref={ref} {...props} />
```
itaihanski, line 45: This indentation seems weird

itaihanski (dismissed):
LGTM, perhaps @shilgapira should give his approval as well

packages/libs/sdk-helpers/src/state.ts
```diff
@@ -73,7 +73,7 @@ export class State<T extends StateObject> {
       typeof newState === 'function' ? newState(this.#state) : newState;
 
     const nextState = { ...this.#state, ...internalNewState };
-    if (!this.#updateOnlyOnChange || !compareObjects(this.#state, nextState)) {
+    if (this.#forceUpdate || !compareObjects(this.#state, nextState)) {
```
asafshen, line 76: can you explain this change? (not the renaming, why the default is true now)

packages/sdks/react-sdk/src/types.ts
```diff
@@ -128,6 +128,16 @@ export type DescopeProps = {
   // use to override client context in flow execution
   client?: Record<string, any>;
   styleId?: string;
+  onScreenUpdate?: (
+    screenName: string,
```
asafshen, line 132: why do we not exposing screen id as well?

packages/sdks/react-sdk/src/components/Descope.tsx
nirgur: It's auto formatting

packages/sdks/react-sdk/src/components/Descope.tsx
```diff
@@ -151,9 +153,9 @@ const Descope = React.forwardRef<HTMLElement, DescopeProps>(
        * it can be removed once this issue will be solved
        * https://bugs.chromium.org/p/chromium/issues/detail?id=1404106#c2
        */
-      <form>
-        <Suspense fallback={null}>
-          <DescopeWC
+	<form>
+		<Suspense fallback={null}>
+			<DescopeWC
```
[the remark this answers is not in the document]
nirgur: It's auto formatting

packages/sdks/web-component/src/lib/types.ts
```diff
@@ -99,11 +125,21 @@ export type StepState = {
   openInNewTabUrl?: string;
 } & OIDCOptions;
 
+export type CustomScreenState = Omit<
+  ScreenState,
+  'cssVars' | 'componentsConfig' | 'inputs'
```
asafshen, line 130: should the omit values match also client scripts, errorText, etc?

image

packages/sdks/web-component/src/lib/types.ts
```diff
@@ -10,10 +10,23 @@ export type SdkFlowNext = Sdk['flow']['next'];
 export type ComponentsConfig = Record<string, any>;
 export type CssVars = Record<string, any>;
 
-type OmitFirstArg<F> = F extends (x: any, ...args: infer P) => infer R
-  ? (...args: P) => R
+type KeepArgsByIndex<F, Indices extends readonly number[]> = F extends (
```
asafshen, line 13: sometimes, Typescript scares me 🇸🇪

asafshen (commented):
looks good

might be out of scope for this, but it would be valueable to have examples for screen states that does
 - passwords (just a regular form)
 - polling (enchanted link)
 - webauthn 
 - totp

packages/libs/sdk-helpers/src/state.ts
nirgur: `updateOnlyOnChange=false` is the same as `forceUpdate=true`

packages/sdks/react-sdk/src/types.ts
nirgur: Because this is something the descoper cannot edit, 
We wanted to avoid a scenario where replacing a screen in a flow breaks the application code (because the id is different)

packages/sdks/web-component/src/lib/types.ts
nirgur: Yes, the errors keys are under an error object
The client scripts are running in the SDK, the descoper does not need to run it as part of his screen
If you think there are more keys that should be added/removed please tell

nirgur:
> looks good
> 
> might be out of scope for this, but it would be valueable to have examples for screen states that does
> 
> * passwords (just a regular form)
> * polling (enchanted link)
> * webauthn
> * totp

What do you mean?
I don't think we are returning the passwords in clear text
regarding the rest, how it can be vulnerable? can you please explain

asafshen:
> > looks good
> > might be out of scope for this, but it would be valueable to have examples for screen states that does
> > 
> > * passwords (just a regular form)
> > * polling (enchanted link)
> > * webauthn
> > * totp
> 
> What do you mean? I don't think we are returning the passwords in clear text regarding the rest, how it can be vulnerable? can you please explain

I don't think anything is vulnerable -  I'm just saying its not trivial to know how you run it, and it will be nice to have examples, especially when screens does "complex" stuff like polling, webauthn, etc

asafshen (approved):

PR #1013, opened 2025-02-10
feat(web-component): add base-cdn-url attribute

omercnet:
## Related Issues

Fixes https://github.com/descope/etc/issues/9009

Adding a way to customize CDN

PR #1014, opened 2025-02-10
fix: add baseCdnUrl attribute in all packages

omercnet:
followup on https://github.com/descope/descope-js/pull/1013

adding baseCdnUrl attribute anywhere descopeUi is used
removing redundant baseStaticUrl mixin as it's already declared in the static resource mixin


packages/sdks/vue-sdk/src/types.ts
nirgur, line 1: I think we should add it to vue the components as well

packages/sdks/react-sdk/src/types.ts
```diff
@@ -97,6 +97,7 @@ export interface IContext {
   baseUrl?: string;
   styleId?: string;
   baseStaticUrl?: string;
+  baseCdnUrl?: string;
```
nirgur, line 100: I think it's missing in the Descope component (which render the descope-wc)

packages/sdks/core-js-sdk/src/createSdk.ts
```diff
@@ -10,6 +10,8 @@ type SdkConfig = {
   projectId: string;
   logger?: Logger;
   baseUrl?: string;
+  baseStaticUrl?: string;
+  baseCdnUrl?: string;
```
nirgur, line 14: I don't think this is needed here

nirgur (commented):
Looks good but I think it's missing in some places

packages/sdks/vue-sdk/src/types.ts
omercnet: done

nirgur (approved):

PR #1015, opened 2025-02-10
feat(web-js-sdk/withPersistTokens): allow customizing SameSite RELEASE

omercnet:
## Related Issues

Fixes https://github.com/descope/etc/issues/9285

## Description

Providing a way to customize the SameSite property by passing an object instead of just `true` to `sessionTokenViaCookie` helper

Updated react/vue/angular/nextjs SDKs to support the new type

## Must

- [x] Tests
- [x] Documentation (if applicable)


packages/sdks/react-sdk/examples/app/index.tsx
```diff
@@ -14,6 +14,7 @@ root.render(
       projectId={process.env.DESCOPE_PROJECT_ID!}
       baseUrl={process.env.DESCOPE_BASE_URL}
       baseStaticUrl={process.env.DESCOPE_BASE_STATIC_URL}
+      sessionTokenViaCookie={{ SameSite: 'Lax' }}
```
asafshen, line 17: I think its a good idea leaving react sdk sample to manage session token on local storage and not on cookie, lets revert this

packages/sdks/vue-sdk/example/main.ts
```diff
@@ -10,6 +10,7 @@ app.use(descope, {
   projectId: process.env.VUE_APP_DESCOPE_PROJECT_ID || '',
   baseUrl: process.env.VUE_APP_DESCOPE_BASE_URL || '',
   baseStaticUrl: process.env.VUE_APP_DESCOPE_BASE_STATIC_URL || '',
+  sessionTokenViaCookie: { SameSite: 'Lax' },
```
asafshen, line 13: see my comment about react-sdk sample app, I would revert

packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/index.ts
```diff
@@ -11,24 +11,24 @@ import {
...
-import { PersistTokensOptions } from './types';
+import { CookieConfig, PersistTokensOptions, SameSite } from './types';
 
 /**
  * Persist authentication tokens in cookie/storage
  */
 export const withPersistTokens =
   <T extends CreateWebSdk>(createSdk: T) =>
-  <A extends boolean>({
+  <A extends boolean | CookieConfig>({
```
asafshen, line 21: not critical, but cookie config is already boolean so you can omit it here

packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -67,7 +73,13 @@ export const persistTokens = (
   // persist session token
   if (sessionJwt) {
     sessionTokenViaCookie
-      ? setJwtTokenCookie(SESSION_TOKEN_KEY, sessionJwt, cookieParams)
+      ? setJwtTokenCookie(SESSION_TOKEN_KEY, sessionJwt, {
+          ...cookieParams,
+          cookieSameSite:
+            sessionTokenViaCookie === true
```
asafshen, line 79: if someone will pass invalid value (e.g. `same-site` instead `SameSite` or just `{}`, they can do it even though TS allow this) - the fallback here would be browser's - which might be lax

there could be a theoretical BC issue, if someone passed before `sessionTokenViaCookie={'true'}` or  `sessionTokenViaCookie={{}}`

I would do something more stable such as `sessionTokenViaCookie?.SameSite || 'Strict'`

packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/types.ts
```diff
@@ -1,4 +1,7 @@
-export type PersistTokensOptions<A extends boolean> = {
+export type SameSite = 'Strict' | 'Lax' | 'None';
+export type CookieConfig = boolean | { SameSite: SameSite };
```
asafshen, line 2: general note about this input, I think its more consistent with JS to call the attribute `sameSite` and not `SameSite` ( I'm refering to the key, not the the type name, which is upper case)

I would change to 
```suggestion
export type CookieConfig = boolean | { sameSite: SameSite };
```

also across all inputs

asafshen (commented):
looks great!
well done on the smooth implementation

asafshen (dismissed):

asafshen (approved):

PR #1025, opened 2025-02-16
feat: support cookie rename

asafshen:
## Related Issues
https://github.com/descope/etc/issues/2225

## Description

support `refreshCookieName`  / `refresh-cookie-name`

require
- implement on angular/vue

### DIFF
--- a/packages/widgets/access-key-management-widget/src/lib/widget/mixins/apiMixin.ts
+++ b/packages/widgets/access-key-management-widget/src/lib/widget/mixins/apiMixin.ts
@@ -30,6 +30,7 @@ export const apiMixin = createSingletonMixin(
           this.tenantId,
           this.mock === 'true',
           this.widgetId,
+          this.getAttribute('refresh-cookie-name'),
         );
       }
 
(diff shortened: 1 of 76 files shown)


asafshen:
> Looks good to me, @nirgur please review as well.
> 
> Also, need to make sure to merge that only after next deployment.

absolutely 👍

meirwah:
about this feature, need to take into consideration
we have in CF a rule, that basically returns automatically 400 error for refresh api that is called without DSR cookie OR Authz header

so if DSR is renamed, we need to adjust this CF rule

packages/widgets/access-key-management-widget/src/lib/widget/mixins/apiMixin.ts
```diff
@@ -26,7 +26,11 @@ export const apiMixin = createSingletonMixin(
       #createSdk() {
         this.logger.debug('creating an sdk instance');
         this.#api = createSdk(
-          { projectId: this.projectId, baseUrl: this.baseUrl },
+          {
+            projectId: this.projectId,
+            baseUrl: this.baseUrl,
+            refreshCookieName: this.getAttribute('refresh-cookie-name'),
```
nirgur, line 32: Can we use `refreshCookieName` from `cookieConfigMixin`?

asafshen: yes 👍

nirgur (approved):

nirgur (approved):

PR #1027, opened 2025-02-19
fix: TS override

asafshen:
## Related Issues

might fix https://github.com/descope/etc/issues/9400

## Description
adjust TS override

### DIFF
--- a/packages/sdks/react-sdk/src/types.ts
+++ b/packages/sdks/react-sdk/src/types.ts
@@ -14,7 +14,7 @@ import UserProfileWidget from '@descope/user-profile-widget';
 import ApplicationsPortalWidget from '@descope/applications-portal-widget';
 import createSdk from './sdk';
 
-declare module 'react' {
+declare global {
   namespace JSX {
     interface IntrinsicElements {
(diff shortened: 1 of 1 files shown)


asafshen:
tested on both react 18 and 19 apps

the relevant change in react 19 https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript

packages/sdks/react-sdk/src/types.ts
```diff
@@ -14,7 +14,7 @@ import UserProfileWidget from '@descope/user-profile-widget';
 import ApplicationsPortalWidget from '@descope/applications-portal-widget';
 import createSdk from './sdk';
 
-declare module 'react' {
+declare global {
```
nirgur, line 17: Did we check it also with React 19?
To make sure there are no type errors

nirgur (approved):

packages/sdks/react-sdk/src/types.ts
asafshen: I have a react ts app that I built with vite and it works well there 👍

PR #1030, opened 2025-02-22
feat: support components attributes

nmacianx:
## Related Issues

Related to https://github.com/descope/etc/issues/8207

## Description

Adds support for components' attributes via SDK

## Must

- [X] Tests

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -721,9 +721,9 @@ class DescopeWc extends BaseDescopeWc {
           name: this.sdk.getLastUserDisplayName() || loginId,
         },
         componentsConfig: {
-          ...screenState?.componentsConfig,
           ...flowConfig.componentsConfig,
           ...conditionComponentsConfig,
+          ...screenState?.componentsConfig,
         },
       },
       htmlFilename: `${readyScreenId}.html`,
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -48,6 +48,7 @@ import {
 } from '../helpers/templates';
 import {
   ClientScript,
+  ComponentsAttrs,
   CustomScreenState,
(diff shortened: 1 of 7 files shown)


nirgur:
It seems similar to componentsConfig
Why do we need both?

nmacianx:
@Bars92 @tomerlichtash @nirgur refactored!

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -717,6 +720,11 @@ class DescopeWc extends BaseDescopeWc {
           loginId,
           name: this.sdk.getLastUserDisplayName() || loginId,
         },
+        componentsConfig: {
+          ...screenState?.componentsConfig,
+          ...flowConfig.componentsConfig,
+          ...conditionComponentsConfig,
```
Bars92, line 726: shouldnt the screen state take priority here?

packages/sdks/web-component/src/lib/helpers/templates.ts
```diff
@@ -201,10 +201,26 @@ const setElementConfig = (
...
+  const { componentsDynamicAttrs, ...rest } = componentsConfig;
+
+  const configMap = Object.keys(rest).reduce((acc, componentName) => {
+    acc[`[name=${componentName}]`] = rest[componentName];
+    return acc;
+  }, {});
+
+  if (componentsDynamicAttrs) {
+    Object.keys(componentsDynamicAttrs).forEach((componentSelector) => {
+      const { attributes } = componentsDynamicAttrs[componentSelector];
```
Bars92, line 213: might want to avoid deconstruct here if componentsDynamicAttrs[componentSelector] is undefined this will fail

Bars92 (commented):
nice work! mind few comments 👍

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nmacianx: i thought about this but wasn't sure since i thought this would happen only as the flow starts, but it makes sense

Bars92 (approved):

PR #1031, opened 2025-02-23
fix BYOS redundant onScreenUpdate calls

nirgur:
**Background:**
Open-Web reported that the `onScreenUpdate` callback is being called more often than expected
It should only be triggered after state changes and `next` calls, but they noticed it’s also firing in other scenarios

**Root Cause:**
After investigating, I found that this happens because we modified the flow state behavior to trigger callbacks on every call to the update function, even when there’s no actual state change
As a result, `onScreenUpdate` is called on any flow state update, even when it’s not needed

**For example:**
We update the state when the application’s visibility changes, and this causes unnecessary `onScreenUpdate` calls

**The Fix:**
I made the following changes to address this:
	•	Reverted the flow state behavior: Callbacks will now only be triggered when there’s an actual state change, not on every update call
• Reverted the step state behavior: Callbacks will now only be triggered when there’s an actual state change, not on every update call
	•	Added a next call timestamp: This acts as a cache buster so that `onScreenUpdate` still gets triggered when next is called, even if the state hasn’t changed
	•	Updated the polling mechanism: The polling mechanism started going out of sync because of the above change, so I adjusted it accordingly

**Verification & Testing:**
I validated these changes with several flows, and everything seems to be running without issues 
However, I’m not familiar with every use case of the SDK, so I’d really appreciate your help here

I also noticed that unit tests alone aren’t enough to catch all potential regressions 
I encountered some issues only when running actual flows, even though all unit tests passed
I’ve added more unit tests to cover these cases, but we still need some manual testing to be sure

**Request:**
Could you please review this change and run some basic tests in the areas you’re familiar with? 
This will help us ensure there are no regressions and that we can release this fix with greater confidence

Thanks in advance for your cooperation!

### DIFF
--- a/packages/sdks/web-component/src/app/index.html
+++ b/packages/sdks/web-component/src/app/index.html
@@ -84,9 +84,11 @@
       descopeEle.onScreenUpdate = async (screenName, context, next, ref) => {
         console.log('onScreenUpdate', screenName, context);
 
+        // if (screenName !== 'Sign In') return false;
+
         // ref.innerHTML = `
         //   <form>
-        //     <input type="text" name="email" placeholder="Email" />
+        //     <input type="text" name="password" placeholder="Password" />
         //     <button type="submit">Submit</button>
         //   </form>
         // `;
@@ -96,10 +98,10 @@
         //   const formData = new FormData(e.target);
         //   const data = Object.fromEntries(formData.entries());
 
-        //   next('Ppb_65tyyn', data);
-        // });
+        //   next('pXVwWREG7M', data);
+        // }, { once: true });
 
-        return false;
+        // return true;
       };
     </script>
   </body>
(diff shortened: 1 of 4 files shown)


packages/sdks/web-component/src/app/index.html
```diff
@@ -96,10 +98,10 @@
         //   const formData = new FormData(e.target);
         //   const data = Object.fromEntries(formData.entries());
 
-        //   next('Ppb_65tyyn', data);
-        // });
+        //   next('pXVwWREG7M', data);
+        // }, { once: true });
```
asafshen, line 102: in this example - is the `once: true` is by purpose?

this means that the developer will have to add it every time

asafshen (approved):
great 👍 

just sharing my thought -  the thing I am worried about the most are regression polling, and specifically in mobile runtimes

packages/sdks/web-component/src/app/index.html
nirgur: Yes, it's on purpose 
Otherwise, every time this fn is called we will add another listener to the submit

itaihanski:
Tested Android polling and everything looks intact 👍

asafshen (approved):

PR #1032, opened 2025-02-25
feat: http session cookie

asafshen:
## Description
support http session cookie. 
 - take the session expiration from JWT response's top level attribute, and not from the session jwt (which might be on cookie)
 - improve high level sdk state management - to not consider session token existence as `is authenticated` state, but rather  manage this separately 
 
 still missing tests and some other considerations  (see comment in vue sdk)

### DIFF
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/access-key-management/access-key-management.component.spec.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/access-key-management/access-key-management.component.spec.ts
@@ -18,8 +18,6 @@ describe('DescopeAccessKeyManagementComponent', () => {
   let component: AccessKeyManagementComponent;
   let fixture: ComponentFixture<AccessKeyManagementComponent>;
   let mockedCreateSdk: jest.Mock;
-  const onSessionTokenChangeSpy = jest.fn();
-  const onIsAuthenticatedChangeSpy = jest.fn();
   const onAccessKeyChangeSpy = jest.fn();
   const afterRequestHooksSpy = jest.fn();
   const mockConfig: DescopeAuthConfig = {
@@ -30,8 +28,6 @@ describe('DescopeAccessKeyManagementComponent', () => {
     mockedCreateSdk = mocked(createSdk);
 
     mockedCreateSdk.mockReturnValue({
-      onSessionTokenChange: onSessionTokenChangeSpy,
-      onIsAuthenticatedChange: onIsAuthenticatedChangeSpy,
       onAccessKeyChange: onAccessKeyChangeSpy,
       httpClient: {
         hooks: {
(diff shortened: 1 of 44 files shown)


packages/sdks/angular-sdk/projects/angular-sdk/src/lib/services/descope-auth.service.ts
```diff
@@ -63,22 +64,6 @@ export class DescopeAuthService {
       isSessionLoading: true
     });
     return this.descopeSdk.refresh().pipe(
-      tap((data) => {
```
asafshen, line 66: this is not required because the `onSessionTokenChange`  handles this already

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/access-key-management/access-key-management.component.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('DescopeAccessKeyManagementComponent', () => {
   let fixture: ComponentFixture<AccessKeyManagementComponent>;
   let mockedCreateSdk: jest.Mock;
   const onSessionTokenChangeSpy = jest.fn();
+  const onIsAuthenticatedChangeSpy = jest.fn();
```
nirgur, line 22: Is it relevant for widgets as well?

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/services/descope-auth.service.ts
```diff
@@ -185,9 +170,16 @@ export class DescopeAuthService {
   private setSession(sessionToken: string | null) {
     const currentSession = this.sessionSubject.value;
     this.sessionSubject.next({
-      sessionToken,
-      isAuthenticated: !!sessionToken,
-      isSessionLoading: currentSession.isSessionLoading
```
nirgur, line 190: Don't we need the `isSessionLoading`?

nirgur (dismissed):

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/services/descope-auth.service.ts
asafshen: we do, its in `...currentSession,`

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/access-key-management/access-key-management.component.spec.ts
asafshen: we don't, also don't need `onSessionTokenChangeSpy` so removing them both 👍

nirgur (dismissed):

nirgur (approved):

PR #1036, opened 2025-03-04
chore: add `dcr` query parameter 

asafshen:
## Description
related to https://github.com/descope/etc/issues/9656
add `dcr` query parameter (either `dcr=t` or `dcr=f` )


packages/sdks/web-js-sdk/test/sdk.test.ts
```diff
@@ -66,8 +66,9 @@ describe('sdk', () => {
     });
   });
 
-  it('should set dsc query param to false on refresh when the session token does not exist', async () => {
+  it('should set dsc query param to false on refresh when the refresh and session token do not exist', async () => {
```
orius123, line 69:
```suggestion
  it('should set dcs and dcr query params to false on refresh when the refresh and session token do not exist', async () => {
```

orius123 (dismissed):

PR #1045, opened 2025-03-09
feat: added option to dismiss screen error on input

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/8643

### DIFF
--- a/packages/sdks/web-component/src/app/index.html
+++ b/packages/sdks/web-component/src/app/index.html
@@ -32,6 +32,7 @@
         validate-on-blur="false"
         keep-last-authenticated-user-after-logout="true"
         nonce="rAnd0m"
+        dismiss-screen-error-on-input="true"
       ></descope-wc>
       <div class="loading">Loading ...</div>
     </div>
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -42,6 +42,7 @@ import { getLastAuth, setLastAuth } from '../helpers/lastAuth';
 import { IsChanged } from '../helpers/state';
 import {
   disableWebauthnButtons,
+  replaceElementMessage,
   setCssVars,
   setNOTPVariable,
   setPhoneAutoDetectDefaultCode,
@@ -249,6 +250,41 @@ class DescopeWc extends BaseDescopeWc {
     return Promise.race([promises, toPromise]);
   }
 
+  get isDismissScreenErrorOnInput() {
+    return this.getAttribute('dismiss-screen-error-on-input') === 'true';
+  }
+
+  #handleGlobalErrors({
+    errorText,
+    errorType,
+  }: {
+    errorText: string;
+    errorType: string;
+  }) {
+    const updateGlobalError = () => {
+      let transformedErrorText = errorText;
+      try {
+        transformedErrorText =
+          this.errorTransformer?.({
+            text: errorText,
+            type: errorType,
+          }) || errorText;
+      } catch (e) {
+        this.loggerWrapper.error('Error transforming error message', e.message);
+      }
+      replaceElementMessage(
+        this.contentRootElement,
+        'error-message',
+        transformedErrorText,
+      );
+    };
+
+    // we do not know if the page is going to be updated or not,
+    // so we are updating the error message component before and after the screen update
+    this.addEventListener('screen-updated', updateGlobalError, { once: true });
+    updateGlobalError();
+  }
+
   init() {
     // when running in a webview (mobile SDK) we want to lazy init the component
     // so the mobile SDK will be able to register all the necessary callbacks
@@ -262,11 +298,25 @@ class DescopeWc extends BaseDescopeWc {
     return undefined;
   }
 
+  #subscribeStepState() {
+    this.stepState?.subscribe(
+      this.onStepChange.bind(this),
+      ({
+        screenState: { errorText, errorType, ...screenState } = {},
+        ...state
+      }) => ({ ...state, screenState }),
+    );
+    this.stepState?.subscribe(this.#handleGlobalErrors.bind(this), (state) => ({
+      errorText: state?.screenState?.errorText,
+      errorType: state?.screenState?.errorType,
+    }));
+  }
+
   // eslint-disable-next-line no-underscore-dangle
   async _init() {
     if (this.shadowRoot.isConnected) {
       this.flowState?.subscribe(this.onFlowChange.bind(this));
-      this.stepState?.subscribe(this.onStepChange.bind(this));
+      this.#subscribeStepState();
 
       window.addEventListener(
         'visibilitychange',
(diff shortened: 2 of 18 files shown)


packages/sdks/web-component/src/app/index.html
```diff
@@ -32,6 +32,7 @@
         validate-on-blur="false"
         keep-last-authenticated-user-after-logout="true"
         nonce="rAnd0m"
+        dismiss-screen-error-on-input="true"
```
asafshen, line 35: do we want it to be the default?

packages/sdks/web-component/src/app/index.html
asafshen, line 1: I think we should also add this in readmes and in higher level sdks (at least react)

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1445,6 +1494,22 @@ class DescopeWc extends BaseDescopeWc {
       });
 
     this.#addPasscodeAutoSubmitListeners(next);
+
+    if (this.isDismissScreenErrorOnInput) {
+      // listen to all input events in order to clear the global error state
+      this.contentRootElement.querySelectorAll('[name]').forEach((ele) => {
```
asafshen, line 1500: image

shouldn't we use this selector?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -380,10 +430,10 @@ class DescopeWc extends BaseDescopeWc {
       const subscribeId = this.stepState.subscribe(() => {
         // after state was updated, we want to re-subscribe the onStepChange listener
         this.stepState.unsubscribe(subscribeId);
-        this.stepState?.subscribe(this.onStepChange.bind(this));
+        this.#subscribeStepState();
       });
     }
-
+    this.stepState.forceUpdate = isCustomScreen;
```
asafshen, line 436: remind me what forceUpdate menas?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -249,6 +250,41 @@ class DescopeWc extends BaseDescopeWc {
...
+      replaceElementMessage(
+        this.contentRootElement,
+        'error-message',
+        transformedErrorText,
+      );
+    };
+
+    // we do not know if the page is going to be updated or not,
+    // so we are updating the error message component before and after the screen update
+    this.addEventListener('screen-updated', updateGlobalError, { once: true });
```
asafshen, line 284: when does screen-update event is fired? in custom screens only?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -262,11 +298,25 @@ class DescopeWc extends BaseDescopeWc {
     return undefined;
   }
 
+  #subscribeStepState() {
+    this.stepState?.subscribe(
```
asafshen, line 302: if I understand correctly, we now subscribe twice - once for error state, and another time for anything else

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -262,11 +298,25 @@ class DescopeWc extends BaseDescopeWc {
...
+      this.onStepChange.bind(this),
+      ({
+        screenState: { errorText, errorType, ...screenState } = {},
+        ...state
+      }) => ({ ...state, screenState }),
+    );
+    this.stepState?.subscribe(this.#handleGlobalErrors.bind(this), (state) => ({
```
asafshen, line 309: if I understand correctly, the first `onStepChange` will remove the errorText,errorType from state, and the second will assign them again, so order matters, here... is that right? or am I missing something?

packages/sdks/web-component/src/lib/helpers/state.ts
```diff
@@ -73,22 +86,37 @@ class State<T extends StateObject> {
...
+          );
+        }
+      });
+    }, 0);
   };
 
-  subscribe(cb: SubscribeCb<T>) {
+  subscribe<R extends any | Partial<T>>(
+    cb: SubscribeCb<R>,
+    selector: (state: T) => R = (state: T) => state as unknown as R,
```
asafshen, line 116: what makes the `selector` input optional?

asafshen (commented):
looks good, few questions

packages/sdks/web-component/src/app/index.html
nirgur: It's not the default
It's just for the dev app

packages/sdks/web-component/src/app/index.html
nirgur: Will do once we will agree on the logic

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: To call the registered callbacks on every call to update (even when there is no state changes)

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: When rendering a flow screen or a custom screen

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Yep

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: No, I added the option to subscribe to a partial state by providing a selector 
In the first subscription, we exclude the errors, and in the second one, we exclude the rest of the state

packages/sdks/web-component/src/app/index.html
asafshen: I meant default in dev-app :)

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
asafshen: I looked and its only the phone inputs which we should consider in this case,

nirgur: It's the phone on `page-editor-components` which is the V1 components

packages/sdks/web-component/src/app/index.html
nirgur: It's not important, we can remove it later

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: I updated the selector to be on the safe side

asafshen (approved):

PR #1048, opened 2025-03-13
fix: short timeout

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/8912
Fixes https://github.com/descope/etc/issues/7424

## Description

handle too close timeouts in the following ways
 - after auth request - if calculated timeout is less than 20 seconds - do not set refresh timeout
 - when window becomes visible - do not refresh if refresh failed

### DIFF
--- a/packages/sdks/web-js-sdk/src/enhancers/withAutoRefresh/index.ts
+++ b/packages/sdks/web-js-sdk/src/enhancers/withAutoRefresh/index.ts
@@ -5,15 +5,12 @@ import { addHooks, getAuthInfoFromResponse } from '../helpers';
 import {
   createTimerFunctions,
   getTokenExpiration,
-  millisecondsUntilDate,
+  getAutoRefreshTimeout,
 } from './helpers';
 import { AutoRefreshOptions } from './types';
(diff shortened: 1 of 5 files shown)


packages/sdks/web-js-sdk/src/enhancers/withAutoRefresh/index.ts
```diff
@@ -66,16 +64,15 @@ export const withAutoRefresh =
           return;
         }
         refreshToken = refreshJwt;
-        let timeout =
-          millisecondsUntilDate(sessionExpirationDate) - REFRESH_THRESHOLD;
+        const timeout = getAutoRefreshTimeout(sessionExpirationDate);
+        clearAllTimers();
 
-        if (timeout > MAX_TIMEOUT) {
+        if (timeout <= REFRESH_THRESHOLD) {
```
nirgur, line 70: What will happen in this case?
will it break the refresh cycle?

asafshen: added a comment

nirgur (approved):

PR #1051, opened 2025-03-17
feat: force on first use

asafshen:
## Related Issues

instead https://github.com/descope/descope-js/pull/877

https://github.com/descope/etc/issues/2862

## Description
support a property for eager refresh (default is true, like today)
if `forceRefreshOnFirstUse` off, sdk only refresh if user was logged in before

few notes
 - keeping the default behavior as `true` even though imho a sane default is `false`, but we don't want to break backwards
 - if we agree on the solution - I'll write tests and implement also in other sdks (vue/angular)

### DIFF
--- a/packages/sdks/react-sdk/examples/app/index.tsx
+++ b/packages/sdks/react-sdk/examples/app/index.tsx
@@ -15,6 +15,11 @@ root.render(
       baseUrl={process.env.DESCOPE_BASE_URL}
       baseStaticUrl={process.env.DESCOPE_BASE_STATIC_URL}
       refreshCookieName={process.env.DESCOPE_REFRESH_COOKIE_NAME}
+      // we want to pass undefined if the value is not set
+      forceRefreshOnFirstUse={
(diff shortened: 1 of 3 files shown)


packages/sdks/react-sdk/examples/app/index.tsx
```diff
@@ -15,6 +15,11 @@ root.render(
       baseUrl={process.env.DESCOPE_BASE_URL}
       baseStaticUrl={process.env.DESCOPE_BASE_STATIC_URL}
       refreshCookieName={process.env.DESCOPE_REFRESH_COOKIE_NAME}
+      // we want to pass undefined if the value is not set
+      forceRefreshOnFirstUse={
+        process.env.DESCOPE_REFRESH_ON_FIRST_USE &&
+        process.env.DESCOPE_REFRESH_ON_FIRST_USE === 'true'
```
nirgur, line 21:
```suggestion
        process.env.DESCOPE_REFRESH_ON_FIRST_USE === 'true'
```

nirgur (approved):
Looks good 👍

PR #1055, opened 2025-03-18
feat: OIDC client

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/7012

Expose oidc client, with hooks considerations


packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
```diff
@@ -81,6 +93,20 @@ const AuthProvider: FC<IAuthProviderProps> = ({
   const isSessionFetched = useRef(false);
   const isUserFetched = useRef(false);
 
+  // if oidc config is enabled, and we have oidc params in the url
+  // we will finish the login (this should run only once)
+  useEffect(() => {
+    if (sdk && oidcConfig && hasOidcParamsInUrl() && !isOidcFinishedLogin.current) {
```
nirgur, line 99: Don't you think this logic should be part of the web-sdk?

packages/sdks/react-sdk/src/hooks/useSession.ts
```diff
@@ -5,19 +5,20 @@ const useSession = () => {
...
+    isOidcFinishLogin,
     fetchSession,
     isSessionFetched,
     isAuthenticated,
   } = useContext();
 
   // when session should be received, we want the return value of "isSessionLoading" to be true starting from the first call
   // (and not only when receiving an update from the context)
-  const isLoading = useRef(isSessionLoading);
+  const isLoading = useRef(isSessionLoading || isOidcFinishLogin);
```
nirgur, line 16: Maybe we should call it `isOidcLoading`?
`finish login` sounds to me like the login is already finished

packages/sdks/web-js-sdk/src/sdk/oidc/index.ts
```diff
@@ -0,0 +1,291 @@
...
+  oidcConfig?: OidcConfigOptions,
+) => {
+  if (!scriptLoadingPromise) {
+    scriptLoadingPromise = loadOIDCModule();
+  }
+  const { OidcClient, WebStorageStateStore } = await scriptLoadingPromise;
+
+  if (!OidcClient) {
+    throw new Error(
+      'oidc-client-ts is not installed. Please install it by running `npm install oidc-client-ts`',
```
nirgur, line 111: Is this error correct?
if it's not installed, we are trying to fetch it from remote URL

nirgur (commented):
Looks good 👍 
Added a few comments

packages/sdks/react-sdk/src/hooks/useSession.ts
asafshen: good point, adjusted 👍

packages/sdks/web-js-sdk/src/sdk/oidc/index.ts
asafshen: this is after the fallback, (to be honest, actually inherited this code from previous PR, but I think its ok)

packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
asafshen: we can, but it implies that when the sdk is init, it will set the `isOidcLoading` state, which we don't want to do if its not required

nirgur: Not sure I understand, let's talk about it when you have 5 min

nirgur (approved):

PR #1063, opened 2025-03-24
feat: add scripts support

Bars92:
## Related Issues

related https://github.com/descope/etc/issues/8914

## Related PRs
https://github.com/descope/content/pull/805

## Description
- add npm scripts mixin loader
- refactor descope ui loader
- add scripts loader to web components based on client scripts

## Must

- [x] Tests


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -229,7 +229,12 @@ class DescopeWc extends BaseDescopeWc {
           moduleRes?.start?.();
           return moduleRes;
         }
-        const module = await loadSdkScript(script.id);
+        await this.injectNpmLib(
+          '@descope/flow-scripts',
+          'latest',
```
nirgur, line 234: I think this is very dangerous
Maybe we should consider a different strategy

Once we will release a new version of our scripts, all customers will get it in production
In case something is broken, their prod env will be down

packages/sdks/web-component/test/descope-wc.test.ts
```diff
@@ -201,7 +209,28 @@ describe('web-component', () => {
     });
     (createSdk as jest.Mock).mockReturnValue(sdk);
 
+    document.body.classList = [];
+    Object.defineProperty(document.body.classList, 'toggle', {
```
nirgur, line 213: Why this is needed?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: I think our approach should be: 
you didn’t change anything - everything remains the same
No unexpected changes

What I can think of right now is setting a fixed version for the scripts and updating it with SDK releases,
just like any other standard dependency

WDYT?

Bars92: that would defeat the purpose of this pr tho, as customers will have to update sdks for new scripts. i think its easer to mitigate the problem in our devops infra rather then forcing customers to use a fixed version

nirgur (approved):

PR #1065, opened 2025-03-25
feat: react sdk next release

Bars92:
## Related Issues

related <> 

## Related PRs
https://github.com/descope/descope-js/actions/runs/14043385381/job/39318563986

## Description
- fix react sdk next release attempt to release widget packages

## Must

- [ ] Tests


.github/workflows/release-next-react-sdk.yml
```diff
@@ -1,9 +1,11 @@
 name: Release - React SDK next
 
 on:
-  push:
-    branches:
-      - main
+  workflow_run:
+    workflows: ['Release']
```
nirgur, line 5: There are cases were `main` changes but there is no release
This workflow should run also in those cases
Will it run with this config?

Bars92: yea, based on the documentation the workflow will be completed even if failed or skipped, which is good to make sure the current existing packages are up to date

nirgur: But it won't run if we are not releasing an official version
And in case we are releasing an official version, there is no need in next version

I'm not sure if my explanation is clear, if not, let's talk for 5 min

PR #1067, opened 2025-03-27
feat: pass external token

asafshen:
## Related Issues
https://github.com/descope/etc/issues/8987

## Description
support passing external token


packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
nirgur, line 1: What about the other SDKs?

packages/sdks/web-js-sdk/src/sdk/index.ts
```diff
@@ -36,10 +41,24 @@ const createSdk = (config: WebSdkConfig) => {
...
       const currentRefreshToken = getRefreshToken();
-      return coreSdk.refresh(token, {
-        dcs: currentSessionToken ? 't' : 'f',
-        dcr: currentRefreshToken ? 't' : 'f',
-      });
+
+      let externalToken = '';
+      if (config.getExternalToken) {
+        try {
+          externalToken = await config.getExternalToken();
```
nirgur, line 48:
```suggestion
          externalToken = await config.getExternalToken?.();
```

packages/sdks/core-js-sdk/src/sdk/index.ts
```diff
@@ -63,10 +63,19 @@ export default (httpClient: HttpClient) => ({
...
    */
   refresh: withOptionalTokenValidations(
-    (token?: string, queryParams?: { [key: string]: string }) =>
-      transformResponse<JWTResponse>(
-        httpClient.post(apiPaths.refresh, {}, { token, queryParams }),
-      ),
+    (
+      token?: string,
+      queryParams?: { [key: string]: string },
+      externalToken?: string,
```
nirgur, line 69: Maybe we should think about this fn signature
Does it make sense to pass token & external token on the same call?

asafshen: I don't want to interleave with the generic token mechanism which has other implications on the entire Descope API

but let me think this through

asafshen: though that though and I think that sending it as the same token may lead to inferior API, the reciver need to get them as two different attributes

asafshen:
@nirgur thanks for the review, addressed / fixed your point

packages/sdks/core-js-sdk/src/sdk/index.ts
nirgur: What about extending the function signature to support another argument type, like an object?
Then we could do something like this
```
{ token?: string, queryParams?: { [key: string]: string } } | { externalToken?: string, queryParams?: { [key: string]: string } }
```

nirgur (approved):

Issue #1070, opened 2025-03-28
Missing dependency: oidc-client-ts


joeeinertson-gravie:
Importing `@descope/web-component:3.40.5` gives a missing dependency error:

```
The required JS dependency "oidc-client-ts" is not available, it was required by "node_modules/@descope/web-js-sdk/dist/cjs/index.cjs.js".

Dependency Trace:
	client/core.cljs
	client/descope.cljs
	shadow.js.shim.module$$descope$web_js_sdk$default.js
	node_modules/@descope/web-js-sdk/dist/cjs/index.cjs.js
```

I confirmed that `oidc-client-ts` is not found in `package-lock.json` after `npm install`. Older versions do not have this issue.

asafshen:
@joeeinertson-gravie are you using the web-component directly? Or via react sdk ?

joeeinertson-gravie:
@asafshen web-component directly

asafshen:
this problem only happens on local env (its a TS typing issue)

we are working on a fix
as a workaround, you can install the oidc-client-ts as a dev-dependency: 
```
npm i --save-dev oidc-client-ts
```

markCapstack:
> [@joeeinertson-gravie](https://github.com/joeeinertson-gravie) are you using the web-component directly? Or via react sdk ?

I'm using via react-sdk and have the same problem

`npm i --save-dev oidc-client-ts` as a temporary fix worked

asafshen:
thanks, I proposed a fix here https://github.com/descope/descope-js/pull/1072

asafshen:
thanks, I proposed a fix here https://github.com/descope/descope-js/pull/1072

asafshen:
FYI, the issue has been resolved in the latest version. Please let me know if any issues arise

PR #1072, opened 2025-03-30
fix: OIDC client ts

asafshen:
## Related Issues

Fixes https://github.com/descope/descope-js/issues/1070

## Description
from experimenting - it seems that require works better in this case, since bundlers don’t try to statically analyze require calls in the same way they do for import statements,


tomerlichtash (approved):
THANK YOU <3

PR #1095, opened 2025-04-23
fix: remove engines declarations

asafshen:
## Related Issues

fixes https://github.com/descope/etc/issues/10088

## Description
remove engines declarations from next/angular sdks

testing
 - [x] next + 18
 - [x] next + 20
 - [x] next + 22
 - [x] next + 23
 - [x] angular + 20
 
image


asafshen:
there will be a separate issue to manage engines to all js packages, but I think that this one is a better sane default for that moment

omercnet (approved):

PR #1096, opened 2025-04-24
fix: created time type

asafshen:
## Description

fix type (mocks etc are aligned)


packages/sdks/core-js-sdk/src/sdk/types.ts
```diff
@@ -64,7 +64,7 @@ export type UserResponse = User & {
   picture?: string;
   roleNames?: string[];
   userTenants?: UserTenant[];
-  createTime: number;
+  createdTime: number;
```
dorsha, line 67: This might break compatibility, so maybe we need to support both?

dorsha: And mark the original one as deprecated

asafshen: we don't support `createTime`, its merely a typo on the original PR

dorsha (approved):

PR #1104, opened 2025-04-30
BugFix in createFetchLogger

Cha195:
## Description

A fix to avoid error while parsing JSON. In some cases `respText` turns out to be HTML. This is being invalid JSON, `resp.json()` thrown an error instead of resolving the promise with error message.

I haven't added any error log in this case. Do mention if you would like to add logs / return an error message

### DIFF
--- a/packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
+++ b/packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
@@ -91,7 +91,13 @@ const fetchWrapper =
     const respText = await resp.text();
 
     resp.text = () => Promise.resolve(respText);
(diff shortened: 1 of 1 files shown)


packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
```diff
@@ -91,7 +91,13 @@ const fetchWrapper =
     const respText = await resp.text();
 
     resp.text = () => Promise.resolve(respText);
-    resp.json = () => Promise.resolve(JSON.parse(respText));
+    resp.json = () => Promise.resolve(() => {
```
nirgur, line 94: @meirwah  Could it be that due to Cloudflare rate limiting, we’re receiving HTML responses from our APIs?

PR #1109, opened 2025-05-06
fix: CSSStyleSheet fallback for older browser RELEASE

nirgur:


asafshen:
looks good, but lets review F2F

asafshen (approved):

PR #1110, opened 2025-05-06
fix: issue 10504 RELEASE

nirgur:
fixes: https://github.com/descope/etc/issues/10504

### DIFF
--- a/packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
+++ b/packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
@@ -137,7 +137,7 @@ export const themeMixin = createSingletonMixin(
       }
 
       async #loadComponentsStyle() {
-        const theme = { ...(await this.#themeResource) } as Record<string, any>;
+        const theme = await this.#themeResource;
         if (!theme) return;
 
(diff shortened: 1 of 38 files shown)


asafshen (approved):

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -206,9 +206,15 @@ export const themeMixin = createSingletonMixin(
...
-        this.#loadGlobalStyle();
-        this.#loadComponentsStyle();
         this.#onThemeChange();
+        try {
+          await Promise.all([
+            this.#loadGlobalStyle(),
+            this.#loadComponentsStyle(),
+          ]);
+        } catch (e) {
+          this.logger.error('Failed to load theme', e);
```
asafshen, line 216: do we want to throw the error as well?

nirgur: No need to catch here
committed by mistake

asafshen (dismissed):

dorsha:
@nirgur CI fails

dorsha (approved):

PR #1112, opened 2025-05-08
feat: implement flow nonce header handling 

orius123:
## Related Issues

Related https://github.com/descope/etc/issues/8633

## Description

Handle flow nonce from header and store it in local-storage with ttl

## Must

- [x] Tests

### DIFF
--- a/packages/sdks/web-js-sdk/src/enhancers/withFlowNonce/index.ts
+++ b/packages/sdks/web-js-sdk/src/enhancers/withFlowNonce/index.ts
@@ -1,13 +1,20 @@
 import { CreateWebSdk } from '../../sdk';
 import { AfterRequestHook, BeforeRequestHook } from '../../types';
 import { addHooks } from '../helpers';
-import { FLOW_NONCE_KEY } from './constants';
 import {
-  extractFlowNonceFromResponse,
+  FLOW_NEXT_PATH,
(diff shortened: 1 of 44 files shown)


packages/sdks/web-js-sdk/src/enhancers/withFlowNonce/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      enableFlowNonce = true,
+      nonceStoragePrefix = FLOW_NONCE_PREFIX,
+      ...sdkConfig
+    } = config;
+
+    if (!enableFlowNonce) {
+      return createSdk(sdkConfig) as ReturnType<T>;
+    }
+
+    cleanupExpiredNonces(nonceStoragePrefix);
```
nirgur, line 35: Why do we want to clean all expired keys on SDK init and not do it lazily?

nirgur (approved):
Looks good
Added a few small comments but nothing is critical

packages/sdks/web-js-sdk/src/enhancers/withFlowNonce/index.ts
orius123: We do it lazy, thought to add here for other execution ids

nirgur (approved):

PR #1116, opened 2025-05-13
fix: issue-9913 RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/9913


packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -716,9 +716,24 @@ export const transformStepStateForCustomScreen = (
...
+  }
+
   return sanitizedState;
 };
 
+export const transformScreenInputs = (inputs: Record<string, any>) => {
+  const res = { ...inputs };
+
+  if (inputs.inboundAppApproveScopes) {
+    res.thirdPartyAppApproveScopes = inputs.inboundAppApproveScopes;
```
asafshen, line 731: this is for BC, no? worth a comment on that

also what if `inboundAppApproveScopes` is empty? can't we have `res.thirdPartyAppApproveScopes = inputs.inboundAppApproveScopes;`?

packages/sdks/web-component/src/lib/types.ts
```diff
@@ -141,6 +141,11 @@ export type CustomScreenState = Omit<
     type: ScreenState['errorType'];
   };
   action?: string;
+  inboundAppApproveScopes?: {
```
asafshen, line 144: I'm a bit concerned about this type, why do we need this? maybe we can have it something more generic? like `& Record<string, any>`?

asafshen (approved):
looks great, small comments

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: If it's empty we do not need to do anything
This is for custom screens
we want to expose an API that includes the updated names so we are mapping it to the "old" names

packages/sdks/web-component/src/lib/types.ts
nirgur: I'm not sure I understand
This is the type we currently have
Why should we make it more generic?

packages/sdks/web-component/src/lib/helpers/helpers.ts
asafshen: ok

packages/sdks/web-component/src/lib/types.ts
asafshen: it seems a bit obsecure to me that `CustomScreenState` has a very specific property which is `inboundAppApproveScopes`

nirgur: We agreed that the custom screen state will be an object with a flat structure, so I'm following this structure
Regarding the name, this is what they need to send back when calling next, so we decided to use the same name

What do you have in mind?

PR #1119, opened 2025-05-18
fix: oauth race condition

asafshen:
## Related Issues
https://github.com/descope/etc/issues/9119

## Description

fix oauth race condition in case of closing window

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -734,6 +734,7 @@ class DescopeWc extends BaseDescopeWc {
         window.location.origin,
       );
       window.close();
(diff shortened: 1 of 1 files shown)


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -734,6 +734,7 @@ class DescopeWc extends BaseDescopeWc {
         window.location.origin,
       );
       window.close();
+      return;
```
nirgur, line 737: I thought that since we’re closing the window, there would be no need to return here
but I guess it’s needed after all

asafshen: Yes there seems to be a race there

PR #1151, opened 2025-07-07
fix: Issue 11222 - retry on 521 and 524

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/11222

Added a retry mechanism for handling specific HTTP status codes (`521` and `524`) by retrying the request once

### DIFF
--- a/packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
+++ b/packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
@@ -77,7 +77,7 @@ const buildRequestLog = (args: Parameters<Fetch>) =>
 // we should retry once in case we got these status codes:
 // 521: Web Server Is Down (Cloudflare error)
 // 524: A Timeout Occurred (Cloudflare error)
-const retryStatusCode = [521, 524];
+const retryStatusCodes = [521, 524];
 
 /** Log the response object */
 const buildResponseLog = async (resp: Response & { retries?: number }) => {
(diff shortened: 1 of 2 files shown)


packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
```diff
@@ -97,28 +109,33 @@ const fetchWrapper =
     return resp;
   };
 
+// we should retry once in case we got these status codes:
+// 521: Web Server Is Down (Cloudflare error)
+// 524: A Timeout Occurred (Cloudflare error)
+const retryStatusCode = [521, 524];
```
dorsha, line 115: It should be defined before the actual use no?

nirgur: It's not that important in this case but I moved it

dorsha:
Looks good to me, @orius123 please review as well

packages/sdks/core-js-sdk/src/httpClient/helpers/createFetchLogger.ts
```diff
@@ -78,13 +89,19 @@ const buildResponseLog = async (resp: Response) => {
...
     .build();
 };
 
 const fetchWrapper =
   (fetch: Fetch) =>
   async (...args: Parameters<Fetch>) => {
-    const resp = await fetch(...args);
+    let resp: Response & { retries?: number } = await fetch(...args);
+
+    if (retryStatusCodes.includes(resp.status)) {
```
orius123, line 101: Where is the retries limit?

nirgur: There is only one retry

orius123: true

PR #1164, opened 2025-07-16
fix: oauth with popup

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/11404

## Description
handle exchange code error as well as code

image

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -735,9 +735,14 @@ class DescopeWc extends BaseDescopeWc {
       }
     }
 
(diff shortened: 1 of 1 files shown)


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -825,6 +830,7 @@ class DescopeWc extends BaseDescopeWc {
 
             this.flowState.update({
               code: data.code,
+              exchangeError: data.exchangeError,
```
asafshen, line 833: this means that we may override code and exchange error together, we can split into two events as well but I think that is not required at the moment, because there is no case there will be both `exchangeError` and `code`

image


@nirgur lmk if you think otherwise

nirgur (approved):

PR #1168, opened 2025-07-25
fix: restore components state when page is shown from cache RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/11493


dorsha (approved):
Any way to add a test?

nirgur:
@dorsha added test

dorsha (approved):

PR #1182, opened 2025-08-09
feat: try refresh API on init

asafshen:
## Related Issues
https://github.com/descope/etc/issues/2862

## Description
This pull request introduces support for a new "try refresh" authentication endpoint across the JS SDKs, allowing for more flexible session refresh logic that does not fail when the token is missing, invalid, or expired. The main changes include updating the API paths, enhancing the `refresh` method across SDK layers to optionally use the new endpoint, and updating the React AuthProvider to leverage this behavior.

**Authentication refresh enhancements:**

* Added a new `tryRefresh` endpoint to `apiPaths`, enabling a non-failing session refresh option.
* Updated the core SDK's `refresh` method to accept a `tryRefresh` boolean parameter, which determines whether to use the standard or new endpoint. The method now also documents the new parameter in its JSDoc. [[1]](diffhunk://#diff-b54ba820e510c7d454f01a60518f72e5733d0b0080845ca8876fec6f13747c41R64-R65) [[2]](diffhunk://#diff-b54ba820e510c7d454f01a60518f72e5733d0b0080845ca8876fec6f13747c41R74-R82)
* Updated the web SDK's `refresh` method to accept and forward the `tryRefresh` parameter to the core SDK, ensuring consistent behavior across SDK layers. [[1]](diffhunk://#diff-ae711197d7d2a9a89b857b679df66e4ac741a0f52dac4da7e49240f7fc40d03fL21-R24) [[2]](diffhunk://#diff-ae711197d7d2a9a89b857b679df66e4ac741a0f52dac4da7e49240f7fc40d03fR61)

### DIFF
--- a/packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
+++ b/packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
@@ -120,7 +120,7 @@ const AuthProvider: FC<IAuthProviderProps> = ({
     isSessionFetched.current = true;
 
     setIsSessionLoading(true);
-    withValidation(sdk?.refresh)().then(() => {
+    withValidation(sdk?.refresh)(undefined, true).then(() => {
       setIsSessionLoading(false);
     });
   }, [sdk]);
--- a/packages/sdks/core-js-sdk/src/constants/apiPaths.ts
+++ b/packages/sdks/core-js-sdk/src/constants/apiPaths.ts
@@ -88,6 +88,7 @@ export default {
     policy: '/v1/auth/password/policy',
   },
   refresh: '/v1/auth/refresh',
+  tryRefresh: '/v1/auth/try-refresh',
   selectTenant: '/v1/auth/tenant/select',
   logout: '/v1/auth/logout',
   logoutAll: '/v1/auth/logoutall',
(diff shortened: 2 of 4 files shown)


packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
```diff
@@ -120,7 +120,7 @@ const AuthProvider: FC<IAuthProviderProps> = ({
     isSessionFetched.current = true;
 
     setIsSessionLoading(true);
-    withValidation(sdk?.refresh)().then(() => {
+    withValidation(sdk?.refresh)(undefined, true).then(() => {
```
nirgur, line 123: Why do we pass undefined?

packages/sdks/vue-sdk/src/hooks.ts
```diff
@@ -21,7 +21,7 @@ export const useSession = () => {
   const { session } = injectDescope();
 
   if (session.isFetchSessionWasNeverCalled.value) {
-    session.fetchSession();
+    session.fetchSession(true);
```
nirgur, line 24: Are we always using try refresh in the framework SDKs?
Don't we want the descoper to configure it?

asafshen: why should they want/need to configure?

I don't think this is a Descoper decision rather to use the `refresh` or the `try-refresh` api
in the context of the SDK behavior they are the same, its just that when token is missing/invalid/expire - one return 401 and the other 200

packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
nirgur: This API is a bit confusing IMO 
but I don't have any other suggestions ATM

PR #1183, opened 2025-08-10
fix: reload sdk scripts when restoring components state RELEASE

nirgur:


dorsha (approved):

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1648,13 +1649,20 @@ class DescopeWc extends BaseDescopeWc {
       ),
     ).filter((ele) => ele !== submitter);
 
-    const restoreComponentsState = () => {
+    const restoreComponentsState = async () => {
```
orius123, line 1652: Do we need to use awaut now when calling it?

nirgur: I don't think so
It's just for getting the config and it's already loaded

PR #1216, opened 2025-09-17
feat: add storagePrefix option to react sdk

gustavofabro-i2p:
## Related PRs

https://github.com/descope/descope-js/pull/265 

## Description

Add storagePrefix option to react SDK

## Must

- [x] Tests

### DIFF
--- a/packages/sdks/react-sdk/src/components/AuthProvider/useSdk.ts
+++ b/packages/sdks/react-sdk/src/components/AuthProvider/useSdk.ts
@@ -51,7 +51,13 @@ export default ({
   }, [
     projectId,
     baseUrl,
-    sessionTokenViaCookie,
+    // NOTE: Avoid creating another instance of the SDK if the consumer of this
+    // component forgot to `useMemo` the object.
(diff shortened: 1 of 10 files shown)


packages/sdks/react-sdk/src/components/AuthProvider/useSdk.ts
```diff
@@ -44,5 +46,6 @@ export default ({
       storeLastAuthenticatedUser,
       keepLastAuthenticatedUserAfterLogout,
       getExternalToken,
+      storagePrefix,
     });
   }, [projectId, baseUrl, sessionTokenViaCookie, getExternalToken]);
```
asafshen, line 51: storage prefix should also be in the deps array

asafshen (commented):
very neat! thanks for the contribution!

there are other usage of the storage prefixes that I'm not sure that are considered, such as
 - Descope component (`descope-wc` inside web-compoent package)  also creates an sdk
 - Widgets as well

we need to address those as well

PR #1219, opened 2025-09-18
feat: Issue12168 - allow custom storage all the way down to web-js-sdk RELEASE

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/12168

### DIFF
--- a/packages/sdks/web-component/jest.config.js
+++ b/packages/sdks/web-component/jest.config.js
@@ -9,7 +9,7 @@ module.exports = {
   collectCoverageFrom: ['src/lib/**/*.ts'],
   coverageThreshold: {
     global: {
-      branches: 79,
+      branches: 78.5,
       functions: 86,
       lines: 88,
       statements: 88,


packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
```diff
@@ -41,6 +41,12 @@ interface IAuthProviderProps {
   refreshCookieName?: string;
   // Function to get external token, for seamless migration from external system
   getExternalToken?: () => Promise<string>;
+  // Custom storage configuration for tokens and user data
+  customStorage?: {
```
asafshen, line 45: why don't we take this type from web-js and duplicate it? 

we also borrow this type from web-js
image

packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
```diff
@@ -41,6 +41,12 @@ interface IAuthProviderProps {
   refreshCookieName?: string;
   // Function to get external token, for seamless migration from external system
   getExternalToken?: () => Promise<string>;
+  // Custom storage configuration for tokens and user data
```
asafshen, line 44: 1. I would mention that the default is local storage 
2. I would not mention tokens and user data, there are stuff that may stored not for those two purposes, for devise for example. maybe something more generic?

```suggestion
  // Custom storage object for authentication related data 
```

packages/sdks/react-sdk/src/types.ts
```diff
@@ -109,6 +109,11 @@ export interface IContext {
   storeLastAuthenticatedUser?: boolean;
   keepLastAuthenticatedUserAfterLogout?: boolean;
   refreshCookieName?: string;
+  customStorage?: {
```
asafshen, line 112: this type is duplicated also here, I would export from web-js

packages/sdks/vue-sdk/src/types.ts
```diff
@@ -13,6 +13,12 @@ export type Options = {
   sessionTokenViaCookie?: CookieConfig;
   // Default is true. If true, last authenticated user will be stored on local storage and can accessed with getUser function
   storeLastAuthenticatedUser?: boolean;
+  // Custom storage configuration for tokens and user data
+  customStorage?: {
```
asafshen, line 17: duplicated here as well

packages/sdks/web-component/src/lib/types.ts
```diff
@@ -294,3 +294,9 @@ export interface ProjectConfiguration {
 }
 
 export type FlowStatus = 'loading' | 'error' | 'success' | 'ready' | 'initial';
+
+export type CustomStorage = {
```
asafshen, line 298: type s duplicated here as well :)

packages/sdks/web-js-sdk/src/types.ts
```diff
@@ -29,6 +29,7 @@ export type CoreSdkConfig = Head<Parameters<CreateCoreSdk>>;
 export type WebSdkConfig = CoreSdkConfig & {
   oidcConfig?: OidcConfig;
   getExternalToken?: () => Promise<string>;
+  customStorage?: CustomStorage;
```
asafshen, line 32: naming wise, I think that the prefix `custom` is a bit mouthful and redundant, and not provide too much value here (on a broad level, inputs are configuration)

maybe just storage? or even store?
- this term is also consisent with MDN
image
- FYI oidc-ts package uses `stateStorage` - https://github.com/authts/oidc-client-ts/blob/c7765577862f9f4ed337f29cd37b448214dc48ca/src/StateStore.ts#L4


I think that storage is the best, for all inputs, and also for type

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
```diff
@@ -116,11 +126,16 @@ export const getUserFromResponse = async (
 export const isDescopeBridge = () =>
   typeof window !== 'undefined' && !!window['descopeBridge'];
 
-export const isLocalStorage = typeof localStorage !== 'undefined';
+export const isLocalStorage =
+  typeof customStorage !== 'undefined' || typeof localStorage !== 'undefined';
```
asafshen, line 130: just double checking - if custom storage was provided, we want to expire nonce?
image

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
```diff
@@ -116,11 +126,16 @@ export const getUserFromResponse = async (
...
 
 export const setLocalStorage = (key: string, value: string) =>
-  isLocalStorage && localStorage?.setItem(key, value);
+  (customStorage || localStorage)?.setItem?.(key, value);
 export const getLocalStorage = (key: string) =>
-  isLocalStorage && localStorage?.getItem(key);
+  (customStorage || localStorage)?.getItem?.(key);
 export const removeLocalStorage = (key: string) =>
```
asafshen, line 136: there are other places that need to be considered
image


maybe I missed it, but I didn't see that those places also uses the custom storage

asafshen (commented):
looks good! most comments are small, but I think we have some places that we missed and need to also use this custom storage 

also lets not foget to add readme (we can do it after of course)

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
nirgur: According to the code, if there is no storage, we skip nonce cleanup
I think that's ok

packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
nirgur: Done

packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
nirgur: Done

packages/sdks/vue-sdk/src/types.ts
nirgur: Done

packages/sdks/web-js-sdk/src/types.ts
nirgur: `Storage` is the type of the object
I think `customStorage` gives good indication that it's optional, and it allows to override the default storage

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
nirgur: Done

PR #1261, opened 2025-11-08
feat: Add usersAndRoles to flow data from user mgmt widget

OfekAvergil:
## Related Issues

https://github.com/descope/etc/issues/12598

## Related PRs
https://github.com/descope/orchestrationservice/pull/4714

## Description

Send a map of userIds to roles in generic flow button - user management widget


dorsha (commented):
Looks good
I think you can add test for that

dorsha (approved):
But we must have tests

PR #1263, opened 2025-11-13
feat: add formProps to Descope component

IamShobe:
## Description

Suggested solution to allow props propagation to `form` element - for BYOS flows.
This allows integration with libs like tanstack form etc.. - and control the onSubmit events etc..

## Must

- [ ] Tests


nirgur:
> Looks like the binding issue with input is related to `#handleKeyPress` on BaseDescopeWc.
> @nirgur 


Thanks, I noticed that as well
I’ll disable it for the custom screens
We’ll still see the warning about nested forms, but it shouldn’t affect anything

PR #1264, opened 2025-11-13
fix: Tenant widget updates custom attributes modal's input box after edit RELEASE

yoavgodelnik:
## Related Issues
Related to: https://github.com/descope/etc/issues/12285
Fixes: https://github.com/descope/etc/issues/12747 (part 2 of the bug)

## Description
Fixes strange behavior in the custom attributes modals' input box's, where value only updates after clicking "edit" at the second time.

Checked that the test fails before the fix

## Must
- [X] Tests

### DIFF
--- a/packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initTenantCustomAttributesMixin.ts
+++ b/packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initTenantCustomAttributesMixin.ts
@@ -49,12 +49,8 @@ export const initTenantCustomAttributesMixin = createSingletonMixin(
 
       #deleteFlows: Record<string, FlowDriver> = {};
 
-      #initEditModalContent(
-        flowId: string,
-        type: string,
-        attName: string,
-        value: any,
-      ) {
+      #initEditModalContent(flowId: string, type: string, attName: string) {
+        const value = getTenantCustomAttributes(this.state)[attName];
         const customAttributeValue =
           type === AttributeTypeName.ARRAY ? (value || []).join(',') : value;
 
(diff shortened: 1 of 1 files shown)


packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initTenantCustomAttributesMixin.ts
```diff
@@ -176,26 +178,28 @@ export const initTenantCustomAttributesMixin = createSingletonMixin(
       ) {
         const deleteFlowId = nodeEle.getAttribute('delete-flow-id');
         if (deleteFlowId) {
-          this.#deleteModals[deleteFlowId] = this.createModal({
-            'data-id': `delete-${customAttrName}`,
-          });
+          // Only create modal and flow if they don't exist yet
+          if (!this.#deleteModals[deleteFlowId]) {
```
nirgur: This fn should run once, on which scenario its already created?

packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initTenantCustomAttributesMixin.ts
```diff
@@ -144,28 +143,31 @@ export const initTenantCustomAttributesMixin = createSingletonMixin(
...
-          this.#editModals[editFlowId].beforeOpen =
-            this.#initEditModalContent.bind(this, editFlowId);
+            this.#editFlows[editFlowId] = new FlowDriver(
+              () =>
+                this.#editModals[editFlowId]?.ele?.querySelector('descope-wc'),
+              { logger: this.logger },
+            );
+            this.#editModals[editFlowId].beforeOpen = () => {
+              const currentVal = getTenantCustomAttributes(this.state)[attName];
+              this.#initEditModalContent(editFlowId, type, attName, currentVal);
```
nirgur: Maybe it will be more consistent to get the current val inside initEditModalContent?

yoavgodelnik: Isn't beforeOpen is responsible for getting data, and initEditModalContent responsible for rendering? It's more separated in this way for my opinion

nirgur (approved):
I must say I'm not fully understand why some init fns run several times but I guess there is a reason :)

packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initTenantCustomAttributesMixin.ts
yoavgodelnik: Basically not unless changed in another instance of the widget or from another settings page, which supposed to require refresh anyway

PR #1274, opened 2025-11-27
feat: Skip session handling when running in native flows

shilgapira:
## Related Issues
Fixes https://github.com/descope/etc/issues/13000

## Description
- Disable all refresh calls by frontend SDKs when they detect they're running in a native flow. This is already done today for autorefresh in react-sdk for the component itself, but not for other SDKs and not for unrelated code.
- Missing the relevant bits in the Angular SDK if we want to add there as well.

## Must
- [X] Tests

### DIFF
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/utils/constants.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/utils/constants.ts
@@ -6,6 +6,7 @@ export const baseHeaders = {
 };
 
 // Detect if running in a native flow (e.g., mobile app with Descope bridge in a webview)
-export const isDescopeBridge = () => !!(window as any)?.descopeBridge;
+export const isDescopeBridge = () =>
+  typeof window !== 'undefined' && !!(window as any)['descopeBridge'];
 
 export const isBrowser = () => typeof window !== 'undefined';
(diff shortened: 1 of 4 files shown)


packages/sdks/angular-sdk/projects/angular-sdk/src/lib/utils/constants.ts
```diff
@@ -5,4 +5,7 @@ export const baseHeaders = {
   'x-descope-sdk-version': environment.buildVersion
 };
 
+// Detect if running in a native flow (e.g., mobile app with Descope bridge in a webview)
+export const isDescopeBridge = () => !!(window as any)?.descopeBridge;
```
asafshen: though the logic is minimal, is the same in all sdks, I think it worth to put it in `web-js-sdk`  and export it from there

edit - I actually see there is such one, lets export it and use it instead

asafshen (commented):
cool! one comment, consider adding tests, should be feasible in react sdk

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/utils/constants.ts
shilgapira: Done

asafshen (approved):

PR #1291, opened 2025-12-18
feat: Add Descope prefix to default logger

nirgur:
## Related Issues

Fixes https://github.com/descope/etc/issues/13303

In case you do not want the `Descope` prefix in the log messages, you can provide a custom logger
See more details [here](https://github.com/descope/descope-js/blob/main/packages/sdks/web-component/README.md#logger---an-object-that-defines-how-to-log-error-warning-and-info-defaults-to-consoleerror-consolewarn-and-consoleinfo-respectively)


asafshen:
@nirgur this looks good, 
this will solve the logs from different web components, but what about other logs from other layers?

specifically, web-js (Which we have logger) and react (which we currently don't have logger, but we can add)

asafshen (approved):

PR #1300, opened 2026-01-01
feat: session inactivity improvement 

asafshen:
## Related Issues

https://github.com/descope/etc/issues/13001

## Description
improve session inactivity to get `nextRefreshSeconds` from the API


packages/sdks/web-js-sdk/examples/loadEnv.js
nirgur: Much better 💪

asafshen: thanks for the direction :)

nirgur (approved):

PR #1303, opened 2026-01-06
fix: improve session invalidation

asafshen:
## Related Issues

Fixes https://github.com/descope/etc/issues/13535

## Description
Instead considering 401 from all functions - consider only subset of routes

### DIFF
--- a/packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
+++ b/packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
@@ -170,7 +170,7 @@ export const isInvalidSessionResponse = (
   const is4xx = res?.status >= 400 && res?.status < 500;
   if (!is4xx) return false;
   const path = req?.path || '';
-  return SESSION_VALIDATION_ROUTES.some((route) => path.includes(route));
+  return SESSION_VALIDATION_ROUTES.includes(path);
 };
 
 export const isLocalStorage =


packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
```diff
@@ -148,6 +148,30 @@ export const getUserFromResponse = async (
...
+/**
+ * Check if a failed response indicates an invalid/expired session
+ * Only specific routes should trigger logout/clear tokens behavior on failure
+ * Other routes (like OTP verify) may fail for invalid input
+ */
+export const isInvalidSessionResponse = (
+  req: { path?: string },
+  res: Response | undefined,
+): boolean => {
+  if (res?.ok) return false;
```
nirgur: I think it's better to check for 4XX status instead of ok

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
```diff
@@ -148,6 +148,30 @@ export const getUserFromResponse = async (
...
+  const path = req?.path || '';
+  return SESSION_VALIDATION_ROUTES.some((route) => path.endsWith(route));
```
nirgur: Why `endsWith` and not `includes`?

nirgur (commented):
Looks good 👍 
Added a few comments

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
asafshen: fixed 👍

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
asafshen: adjusted to include

packages/sdks/web-js-sdk/src/enhancers/helpers/index.ts
```diff
@@ -148,6 +158,21 @@ export const getUserFromResponse = async (
...
+ * Other routes (like OTP verify) may fail for invalid input
+ */
+export const isInvalidSessionResponse = (
+  req: { path?: string },
+  res: Response | undefined,
+): boolean => {
+  const is4xx = res?.status >= 400 && res?.status < 500;
+  if (!is4xx) return false;
+  const path = req?.path || '';
+  return SESSION_VALIDATION_ROUTES.some((route) => path.includes(route));
```
nirgur:
```suggestion
  return SESSION_VALIDATION_ROUTES.includes(path)
```

PR #1325, opened 2026-01-31
fix: Prevent auto-focus in hidden inputs RELEASE

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/13975

## Description

Prevent `auto-focus` API from focusing on hidden inputs (`type="hidden"` or `aria-hidden`), or inputs that explicitly defined `auto-focus="false"`.

## Must

- [x] Tests


packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -507,6 +507,9 @@ export const withMemCache = <I extends any[], O>(fn: (...args: I) => O) => {
   );
 };
 
+export const FOCUSABLE_INPUTS_SELECTOR =
+  '*[name]:not([auto-focus="false"]):not([type="hidden"]):not([aria-hidden="true"])';
```
nirgur: what is type hidden?
do we have something like this?

nirgur (approved):

PR #1342, opened 2026-03-09
feat(web-component): Track last auth per screen and populate LastAuthBadge RELEASE

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/14017

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| WCUI | [Link to PR](https://github.com/descope/web-components-ui/pull/920) |
| ConsoleApp | [Link to PR](https://github.com/descope/console-app/pull/4977) |

## Description

The last-auth badge feature surfaces a visual indicator on a button that the user previously used during their last successful login. The SDK tracks which interactions occurred, persists that state across page navigations, and wire the badge element to its anchor at render time.

### Approach: 2-key localStorage design

Tracking uses two keys to ensure `dls_last_auth` is only written on a **completed** authentication — abandoned or partial flows never pollute it.

| Key | Lifetime | Contents |
|---|---|---|
| `dls_last_auth_in_flight` | During flow (cleared on completion or new flow start) | `Record<screenId, elementId>` — opt-in clicks accumulated during the current flow |
| `dls_last_auth` | Persistent | `{ loginId, authMethod, lastUsedPerScreen }` — written only on `status === 'completed'` |

### Opt-in tracking

Only elements with `opt-in-last-used="true"` are tracked — an explicit contract for flow designers to mark badge-eligible buttons.

Tracking happens in `#handleSubmit` (after form validation), immediately before `next()` is called:

```ts
if (submitterId && screenId && submitter.getAttribute('opt-in-last-used') === 'true') {
  updateLastUsedPerScreen(screenId, submitterId);
}
```

Because the write happens to localStorage before any navigation, it survives OAuth full-page redirects, magic links, and any other navigation mechanism — no special per-mechanism handling needed.

### Flow completion

When `status === 'completed'`, the in-flight data is merged into `dls_last_auth` and then cleared:

```ts
setLastAuth({ ...lastAuth, lastUsedPerScreen: getInFlightLastUsedPerScreen() }, false);
clearInFlightLastAuth();
```

### Badge anchoring (`#applyLastAuthBadge`)

Called on every screen render. Reads `lastUsedPerScreen[currentScreenId]` from `dls_last_auth` and sets `badgeEl.anchor = targetEl`. No opt-in check here — if an ID is in `lastUsedPerScreen`, it was already gated by `#handleSubmit` at tracking time.

## Must

- [x] Tests


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1876,12 +1899,21 @@ class DescopeWc extends BaseDescopeWc {
...
       if (
         submitter.getAttribute('formnovalidate') === 'true' ||
         this.#validateInputs()
       ) {
         const submitterId = submitter?.getAttribute('id');
+
+        if (
+          submitterId &&
+          screenId &&
+          submitter.getAttribute('opt-in-last-used') === 'true'
```
nirgur: I think it will be better to extract it into a const

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1876,12 +1899,21 @@ class DescopeWc extends BaseDescopeWc {
...
+        ) {
+          updateLastUsedPerScreen(screenId, submitterId);
+        }
```
nirgur: Maybe we should extract this logic into a fn

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1976,20 +2008,43 @@ class DescopeWc extends BaseDescopeWc {
...
           }
         });
       });
   }
 
-  #hydrate(next: NextFn) {
+  #applyLastAuthBadge(screenId: string) {
+    const loginId = this.sdk.getLastUserLoginId();
+    const lastAuth = getLastAuth(loginId);
+    const lastUsedPerScreen = (lastAuth as any)?.lastUsedPerScreen as
```
nirgur: Lets fix the types here

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1976,20 +2008,43 @@ class DescopeWc extends BaseDescopeWc {
...
+      | Record<string, string>
+      | undefined;
+    if (!lastUsedPerScreen) return;
+
+    const componentId = lastUsedPerScreen[screenId];
+    if (!componentId) return;
```
nirgur: You can shorten by using optional chaining

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1976,20 +2008,43 @@ class DescopeWc extends BaseDescopeWc {
...
+
+    const badgeEl = this.contentRootElement.querySelector(
+      DESCOPE_LAST_AUTH_BADGE_COMPONENT,
+    ) as any;
+    if (!badgeEl) return;
+
+    const targetEl = this.contentRootElement.querySelector(`#${componentId}`);
+    if (!targetEl) return;
+
+    targetEl.parentElement?.insertBefore(badgeEl, targetEl);
```
nirgur: Consider using replaceWith

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
```diff
@@ -32,3 +37,43 @@ export function setLastAuth(
...
+    const stored = JSON.parse(
+      getStorageItem(DESCOPE_LAST_AUTH_IN_FLIGHT_LOCAL_STORAGE_KEY) || '{}',
+    );
+    stored[screenId] = elementId;
+    setStorageItem(
+      DESCOPE_LAST_AUTH_IN_FLIGHT_LOCAL_STORAGE_KEY,
+      JSON.stringify(stored),
+    );
+  } catch (e) {
+    /* empty */
```
nirgur: Maybe we should log something in case of an error

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
```diff
@@ -32,3 +37,43 @@ export function setLastAuth(
...
 }
+
+// Track which opt-in element was last used per screen during an in-progress flow.
+// Written on every qualifying click so the data survives any page navigation
+// (OAuth redirect, magic link, etc.) without special per-mechanism handling.
+// This key is separate from dls_last_auth so partial/abandoned flows never
+// pollute the last authenticated user record.
+export function updateLastUsedPerScreen(screenId: string, elementId: string) {
+  try {
+    const stored = JSON.parse(
```
nirgur: Can we reuse getInFlightLastUsedPerScreen?

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
```diff
@@ -32,3 +37,43 @@ export function setLastAuth(
...
+    return {};
+  }
+}
+
+// Clear in-flight state. Called on flow completion and on new flow start.
+export function clearInFlightLastAuth() {
+  try {
+    removeStorageItem(DESCOPE_LAST_AUTH_IN_FLIGHT_LOCAL_STORAGE_KEY);
+  } catch (e) {
+    /* empty */
```
nirgur: log error?

packages/sdks/web-component/test/descope-wc.lastAuth.test.ts
```diff
@@ -0,0 +1,383 @@
...
+          );
+          expect(lastAuth.lastUsedPerScreen?.['screen-1']).toBe('button-b');
+        },
+        { timeout: WAIT_TIMEOUT },
+      );
+    });
+  });
+
+  describe('#applyLastAuthBadge', () => {
+    it('should nest the last used button inside the badge for the current screen', async () => {
```
nirgur: Should replace the button with the badge where the button is now a child of it

packages/sdks/web-component/test/descope-wc.lastAuth.test.ts
nirgur, line 1: Let's add a test for the clear logic

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: done

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: done

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: added console.warn

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: fixed

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: done

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: added console.warn

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: done

packages/sdks/web-component/test/descope-wc.lastAuth.test.ts
tomerlichtash: changed

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: changed

packages/sdks/web-component/test/descope-wc.lastAuth.test.ts
tomerlichtash: done

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1969,20 +1999,42 @@ class DescopeWc extends BaseDescopeWc {
...
+
+    const componentId = lastUsedPerScreen[screenId];
+    if (!componentId) return;
+
+    const badgeEl = this.contentRootElement.querySelector(
+      DESCOPE_LAST_AUTH_BADGE_COMPONENT,
+    ) as any;
+    if (!badgeEl) return;
+
+    const targetEl = this.contentRootElement.querySelector(`#${componentId}`);
```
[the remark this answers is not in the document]
tomerlichtash: escaped

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
```diff
@@ -1,34 +1,74 @@
...
   }
   return lastAuth;
 }
 
 // save last auth to local storage
-export function setLastAuth(
-  lastAuth: NextFnReturnPromiseValue['data']['lastAuth'],
-  forceLoginId?: boolean,
-) {
+export function setLastAuth(lastAuth: LastAuthState, forceLoginId?: boolean) {
```
nirgur: Why the forceLoginId is needed?

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
```diff
@@ -1,34 +1,74 @@
...
     return;
   }
   setStorageItem(DESCOPE_LAST_AUTH_LOCAL_STORAGE_KEY, JSON.stringify(lastAuth));
 }
+
+// Read the in-flight lastUsedPerScreen to be merged into dls_last_auth on completion.
+export function getInFlightLastUsedPerScreen(): Record<string, string> {
+  try {
+    return JSON.parse(
+      getStorageItem(DESCOPE_LAST_AUTH_IN_FLIGHT_LOCAL_STORAGE_KEY) || '{}',
```
nirgur, line 46: why do we need the `|| '{}'`?

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
nirgur, line 1: You should use the logger instead of console in this file

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: Bad naming. This is to prevent saving last auth data when no login ID is present during in-flight.
Changing param name and adding a comment.

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: `getStorageItem` returns `null` when the key doesn't exist.

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: Replaced logs with logger

packages/sdks/web-component/src/lib/helpers/lastAuth.ts
tomerlichtash: Indeed, as we discussed: This is part of the handling of an in-flight auth state. I added a comment that explains where lastAuth runs and why, and explains the use case that this call is needed for, as well as why it won't write on every response, only when in in-flight:

```
    // setLastAuth writes to dls_last_auth from two places:
    //
    //   - Completion (above): saves once the user finishes logging in.
    //
    //   - Here (every non-completed response): a mid-flow snapshot so
    //     a future visit can pre-fill {authMethod, loginId} even if
    //     the user abandons this flow.
    //
    // Use case:
    //   user submits email@example.com → server responds with
    //   {loginId: 'email@example.com', authMethod: 'magicLink'} →
    //   user closes the tab. On the next visit the form pre-fills
    //   the email and shows magicLink, even though the flow never
    //   completed.
    //
    // The helper decides whether to write. requireLoginId=true tells
    // it to skip when loginId isn't set yet — without that guard, an
    // early-screen response could overwrite a prior user's saved
    // record with an auth-method-only entry that can't pre-fill
    // anything.
```

nirgur (approved):

PR #1365, opened 2026-04-12
fix(core-js-sdk): use set-cookie-parser for correct multi-cookie parsing

itaihanski:
## Related Issues
Fixes https://github.com/descope/etc/issues/15018

## Description
- Replace cookie extraction logic with a more robust version, since in the old logic the DSR would've been lost when both DS and DSR were being sent
- Add dedicated `transformSetCookie` tests covering single cookies, multiple cookies (DS + DSR), cookies with `Expires` dates containing commas, custom cookie names, and empty input
- Fix existing `httpClient` test mock to use correct comma-separated cookie format

## Must
- [x] Tests

### DIFF
--- a/packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
+++ b/packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
@@ -1,13 +1,91 @@
-import setCookieParser from 'set-cookie-parser';
+/**
+ * Split a combined Set-Cookie header string into individual cookie strings.
+ *
+ * Set-Cookie values are sometimes comma-joined into one string. This splits
+ * them without choking on commas that appear inside a single cookie's
+ * attributes (e.g. the Expires date "Sun, 10 May 2026 12:00:00 GMT").
+ *
+ * Based on https://github.com/nfriedly/set-cookie-parser (MIT).
+ */
+function splitCookiesString(cookiesString: string): string[] {
+  if (!cookiesString) return [];
 
-export { default as createFetchLogger } from './createFetchLogger';
-export { getClientSessionId } from './getClientSessionId';
+  const result: string[] = [];
+  let pos = 0;
+
+  function skipWhitespace() {
+    while (pos < cookiesString.length && /\s/.test(cookiesString.charAt(pos))) {
+      pos += 1;
+    }
+    return pos < cookiesString.length;
+  }
+
+  function notSpecialChar() {
+    const ch = cookiesString.charAt(pos);
+    return ch !== '=' && ch !== ';' && ch !== ',';
+  }
+
+  while (pos < cookiesString.length) {
+    const start = pos;
+    let cookiesSeparatorFound = false;
+    let lastComma: number;
+    let nextStart: number;
+
+    while (skipWhitespace()) {
+      const ch = cookiesString.charAt(pos);
+      if (ch === ',') {
+        // ',' is a cookie separator if we later find '=' before ';' or ','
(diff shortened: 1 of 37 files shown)


packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
```diff
@@ -1,15 +1,13 @@
...
+import setCookieParser from 'set-cookie-parser';
 
-  return cookiesString.reduce((acc, cookie) => {
-    const [key, value] = cookie.split('=');
-    return {
-      ...acc,
-      [key.trim()]: value,
-    };
-  }, {});
+export function transformSetCookie(
```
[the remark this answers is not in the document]
itaihanski: Fixed — moved the import above the re-exports in 465f59c9.

pnpm-lock.yaml
```diff
@@ -1587,7 +1591,7 @@ importers:
     devDependencies:
       '@descope/web-components-ui':
         specifier: latest
-        version: 2.2.58
+        version: 3.1.5
```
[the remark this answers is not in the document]
itaihanski: The lockfile churn is from the pre-commit hook running `pnpm i`, which re-resolves the full lockfile. This is expected behavior in this repo.

pnpm-lock.yaml
```diff
@@ -1411,7 +1415,7 @@ importers:
         specifier: ^5.0.2
         version: 5.4.5
 
-  packages/sdks/solid-sdk/dist/cjs: {}
+  packages/sdks/react-sdk/dist/cjs: {}
```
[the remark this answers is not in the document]
itaihanski: Same as above — the pre-commit hook regenerates the lockfile, causing these workspace importer changes. Not a manual change.

packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
```diff
@@ -1,15 +1,13 @@
+import setCookieParser from 'set-cookie-parser';
```
asafshen: good catch, and I get what you mean by set-cookie-parser is a zero dependency package 
but if we can have our own implementation that can solves this - I would prefer to implement,and not add another dep

we are aiming to even remove jwt-decode and build our own, so I think that the overhead here worth to implement our own `setCookieParser` logic or copy from lib, 

if we decide to keep it, I would tell renovate not to bump it

itaihanski: Good point - I'll extract the essence into our code

packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
```diff
@@ -1,15 +1,91 @@
...
+        return [
+          nameValue.substring(0, eqIdx).trim(),
+          nameValue.substring(eqIdx + 1).trim(),
+        ];
+      })
+      .filter(Boolean),
+  );
 }
+
+/* istanbul ignore next -- re-exports */
```
asafshen, line 88: not critical, just curious - those lines just moved from above, why do we need this istanbul notation now?

asafshen (approved):
excellent work!

packages/sdks/core-js-sdk/src/httpClient/helpers/index.ts
itaihanski: Leftover from my attempts at 100% coverage. Removing

PR #1376, opened 2026-05-03
feat(widgets): improve toast notifications with detail field and onToast callback

LioriE:
## Related Issues
Required for:
https://github.com/descope/etc/issues/15355

## Description
Adds a `detail` field to notification payloads across user, role, and access-key management widgets so toasts can carry supplementary context alongside the main message. Exposes a typed `onToast` callback on the `UserManagement` React SDK component, allowing consumers to call `event.preventDefault()` to suppress the built-in toast and handle it themselves. Also splits the combined `useEffect` for event listeners into two separate effects so each one only re-runs when its own dependency changes.

## Must
- [x] Tests


packages/sdks/react-sdk/test/components/DescopeWidgets.test.tsx
```diff
@@ -254,6 +254,36 @@ describe('Descope Widgets', () => {
...
+        cancelable: true,
+        detail: { message: 'hi', severity: 'success' },
+      }),
+    );
+    expect(onToastMock).toHaveBeenCalledTimes(1);
+    expect(onToastMock.mock.calls[0][0].detail).toEqual({
+      message: 'hi',
+      severity: 'success',
+    });
+  });
```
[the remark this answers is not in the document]
LioriE: fix this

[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Added `test/initNotificationsMixin.test.ts` with three widget-level unit tests: (1) default behavior shows the notification, (2) calling `preventDefault()` on the dispatched `toast` event suppresses `createNotification()`/`show()`, (3) the event carries the expected `{ message, detail, severity }` payload with `cancelable: true`.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -56,11 +58,18 @@ export const initNotificationsMixin = createSingletonMixin(
...
       #onNotificationsUpdate = withMemCache((notifications: Notifications) => {
         if (notifications.length) {
-          notifications.forEach(({ type, msg }) => {
+          notifications.forEach(({ type, msg, detail }) => {
+            const toastEvent = new CustomEvent('toast', {
+              cancelable: true,
+              detail: { message: msg, detail, severity: type },
+            });
+            this.dispatchEvent(toastEvent);
+            if (toastEvent.defaultPrevented) return;
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Added `test/initNotificationsMixin.test.ts` covering both the event dispatch with correct detail payload and the `preventDefault()` suppression contract.

packages/widgets/role-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -56,11 +58,11 @@ export const initNotificationsMixin = createSingletonMixin(
 
       #onNotificationsUpdate = withMemCache((notifications: Notifications) => {
         if (notifications.length) {
-          notifications.forEach(({ type, msg }) => {
+          notifications.forEach(({ type, msg, detail }) => {
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] The `toast` CustomEvent dispatch and `onToast` prop are intentionally scoped to `user-management-widget` in this PR. Extending to `role-management-widget` and `access-key-management-widget` is out of scope here. Leaving for human review to decide whether to expand in a follow-up.

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -56,11 +58,11 @@ export const initNotificationsMixin = createSingletonMixin(
 
       #onNotificationsUpdate = withMemCache((notifications: Notifications) => {
         if (notifications.length) {
-          notifications.forEach(({ type, msg }) => {
+          notifications.forEach(({ type, msg, detail }) => {
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Same as the role-management comment — `onToast` is intentionally scoped to `user-management-widget` in this PR. Leaving for human review to decide on follow-up expansion.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -56,11 +58,18 @@ export const initNotificationsMixin = createSingletonMixin(
 
       #onNotificationsUpdate = withMemCache((notifications: Notifications) => {
         if (notifications.length) {
-          notifications.forEach(({ type, msg }) => {
+          notifications.forEach(({ type, msg, detail }) => {
+            const toastEvent = new CustomEvent('toast', {
+              cancelable: true,
+              detail: { message: msg, detail, severity: type },
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Valid naming concern — `event.detail.detail` is awkward. However renaming `detail` to `description` on a new public API is a meaningful breaking change. Leaving for human review to decide before this ships.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -34,12 +34,14 @@ export const initNotificationsMixin = createSingletonMixin(
...
-      #createNotificationContent({ type, msg }: Notification) {
+      #createNotificationContent({ type, msg, detail }: Notification) {
         const icon = notificationTypesIcons[type];
 
         const closeIcon = Object.assign(close(), { slot: 'close' });
 
-        return `${icon?.outerHTML || ''}${msg}${closeIcon.outerHTML}`;
+        const body = detail ? `<div><div>${msg}</div>${detail}</div>` : msg;
+
+        return `${icon?.outerHTML || ''}${body}${closeIcon.outerHTML}`;
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Pre-existing behavior — `setContent` via `innerHTML` was there before this PR. Escaping `msg`/`detail` would close the gap but is out of scope for this change. Leaving for human review.

packages/widgets/role-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -34,12 +34,14 @@ export const initNotificationsMixin = createSingletonMixin(
...
-      #createNotificationContent({ type, msg }: Notification) {
+      #createNotificationContent({ type, msg, detail }: Notification) {
         const icon = notificationTypesIcons[type];
 
         const closeIcon = Object.assign(close(), { slot: 'close' });
 
-        return `${icon?.outerHTML || ''}${msg}${closeIcon.outerHTML}`;
+        const body = detail ? `<div><div>${msg}</div>${detail}</div>` : msg;
+
+        return `${icon?.outerHTML || ''}${body}${closeIcon.outerHTML}`;
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Same pre-existing XSS behavior as the user-management mixin. Out of scope for this PR. Leaving for human review.

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -34,12 +34,14 @@ export const initNotificationsMixin = createSingletonMixin(
...
-      #createNotificationContent({ type, msg }: Notification) {
+      #createNotificationContent({ type, msg, detail }: Notification) {
         const icon = notificationTypesIcons[type];
 
         const closeIcon = Object.assign(close(), { slot: 'close' });
 
-        return `${icon?.outerHTML || ''}${msg}${closeIcon.outerHTML}`;
+        const body = detail ? `<div><div>${msg}</div>${detail}</div>` : msg;
+
+        return `${icon?.outerHTML || ''}${body}${closeIcon.outerHTML}`;
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Same pre-existing XSS behavior as the other mixins. Out of scope for this PR. Leaving for human review.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
```diff
@@ -34,12 +34,14 @@ export const initNotificationsMixin = createSingletonMixin(
...
     )(superclass) {
       // eslint-disable-next-line class-methods-use-this
-      #createNotificationContent({ type, msg }: Notification) {
+      #createNotificationContent({ type, msg, detail }: Notification) {
         const icon = notificationTypesIcons[type];
 
         const closeIcon = Object.assign(close(), { slot: 'close' });
 
-        return `${icon?.outerHTML || ''}${msg}${closeIcon.outerHTML}`;
+        const body = detail ? `<div><div>${msg}</div>${detail}</div>` : msg;
```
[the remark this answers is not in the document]
LioriE: [CLAUDE 🤖] Acknowledged — pre-existing behavior. Out of scope for this PR. Leaving for human review to decide on escaping.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+  type: 'success' | 'error';
+  msg: string;
+  detail?: string;
+};
+
+export type ToastEventDetail = {
+  message: string;
+  detail?: string;
+  severity: ToastNotification['type'];
+};
```
nirgur: Do we need 2 different types here?
Can we choose one?

packages/sdks/react-sdk/src/components/AccessKeyManagement.tsx
```diff
@@ -25,45 +25,59 @@ const AccessKeyManagementWC = lazy(async () => {
...
-  }, [innerRef, onReady]);
+      return () => {
+        if (onReady) ele?.removeEventListener('ready', onReady);
+      };
+    }, [innerRef, onReady]);
 
-  return (
+    useEffect(() => {
+      const ele = innerRef;
+      if (onToast) ele?.addEventListener('toast', onToast as EventListener);
```
nirgur: Why do we need to cast `onToast`?

packages/sdks/react-sdk/src/components/RoleManagement.tsx
```diff
@@ -40,6 +43,15 @@ const RoleManagement = React.forwardRef<HTMLElement, RoleManagementProps>(
...
 
+    useEffect(() => {
+      const ele = innerRef;
+      if (onToast) ele?.addEventListener('toast', onToast as EventListener);
+      return () => {
+        if (onToast)
+          ele?.removeEventListener('toast', onToast as EventListener);
+      };
+    }, [innerRef, onToast]);
+
```
nirgur: I see this block repeats several times
WDYT about creating a custom hook for it?
something like
```useEventListener(ele, 'toast', onToast)```

File: packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts [hunk 13]
```diff
@@ -1,84 +1,15 @@
...
-        await super.init?.();
-
-        this.subscribe(
-          this.#onNotificationsUpdate.bind(this),
-          getNotifications,
-        );
-      }
-    },
+export const initNotificationsMixin = compose(
+  stateManagementMixin,
```
nirgur, line 6: Is this needed here?
I don't see we are using any of its fns

packages/widgets/access-key-management-widget/src/lib/widget/state/asyncActions/helpers.ts
nirgur, line 1: Is this file needed?

packages/widgets/access-key-management-widget/test/initNotificationsMixin.test.ts
nirgur, line 1: From what I understand, all the logic was moved to the mixin
So can we move this test file to the sdk-mixins pkg and remove its duplications?

packages/widgets/access-key-management-widget/src/lib/widget/state/types.ts
```diff
@@ -49,4 +49,5 @@ export type Builder = ActionReducerMapBuilder<State>;
 type Notification = {
```
nirgur: Do we need this type here?
can it be taken from the mixin?

packages/widgets/user-management-widget/test/config/fileMock.js
```diff
@@ -1 +1 @@
-module.exports = () => {};
+module.exports = () => document.createElement('span');
```
nirgur: Why we need to change this mock?

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+      )(superclass) {
+        declare subscribe: <SelectorR>(
+          cb: (state: SelectorR) => void,
+          selector?: (state: any) => SelectorR,
+        ) => () => void;
+
+        declare actions: { clearNotifications: () => void };
+
+        // eslint-disable-next-line class-methods-use-this
+        #createNotificationContent({ type, msg, detail }: ToastNotification) {
```
nirgur, line 73: What do you think about simplifying this fn a bit?
```
#createNotificationContent({ type, msg, detail }: ToastNotification) {
  const typeIcon = createIcon(type);
  const closeIcon = createIcon('close');
  const body = detail ? `<div><div>${msg}</div>${detail}</div>` : msg;

  return `${typeIcon}${body}${closeIcon}`;
}
```

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+    <T extends CustomElementConstructor>(superclass: T) =>
+      class ToastEventsMixinClass extends compose(
+        loggerMixin,
+        notificationsMixin,
+        initLifecycleMixin,
+      )(superclass) {
+        declare subscribe: <SelectorR>(
+          cb: (state: SelectorR) => void,
+          selector?: (state: any) => SelectorR,
+        ) => () => void;
```
nirgur, line 68: You should use the stateMixin instead of declaring `subscribe`

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+  compose,
+  createSingletonMixin,
+  withMemCache,
+} from '@descope/sdk-helpers';
+import { initLifecycleMixin } from '../initLifecycleMixin';
+import { loggerMixin } from '../loggerMixin';
+import { notificationsMixin } from '../notificationsMixin';
+
+const SUCCESS_NOTIFICATION_DURATION = 3000;
+const ERROR_NOTIFICATION_DURATION = 0;
```
nirgur:
```suggestion
const DEFAULT_SUCCESS_NOTIFICATION_DURATION = 3000;
const DEFAULT_ERROR_NOTIFICATION_DURATION = 0;
```

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+    close?: () => HTMLElement;
+  };
+  eventName?: string;
+};
+
+export const createToastEventsMixin = <S>(
+  config: ToastEventsMixinConfig<S>,
+) => {
+  const {
+    selector,
```
nirgur: I think we can provide also a default selector
WDYT?

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+    successDuration = SUCCESS_NOTIFICATION_DURATION,
+    errorDuration = ERROR_NOTIFICATION_DURATION,
+    position = 'bottom-start',
+    size = 'sm',
+    icons = {},
```
nirgur: I'm not sure we should allow to customize the icons ATM
WDYT?

nirgur (commented):
Well done 💪 
I added some comments

In addition, we need to update also the other SDKs with this change (Vue & Angular)
And update the relevant docs

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Done — removed `ToastEventDetail` (it was dead code, never imported). `ToastNotification` is now the single type.

packages/sdks/react-sdk/src/components/AccessKeyManagement.tsx
LioriE: The cast is gone — centralised inside `useEventListener<T extends Event>` which handles the `as EventListener` cast internally.

packages/sdks/react-sdk/src/components/RoleManagement.tsx
LioriE: Done — `useEventListener` hook created (`src/hooks/useEventListener.ts`) and used across all components. Also converted the remaining inline `logout` `useEffect` in `UserProfile.tsx`.

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
LioriE: `stateManagementMixin` provides `subscribe` (called in `init()` to watch the notification queue) and `actions.clearNotifications()` (called after processing toasts). Both are required by `createToastEventsMixin` at runtime.

packages/widgets/access-key-management-widget/src/lib/widget/state/asyncActions/helpers.ts
LioriE: Yes — `buildAsyncReducer` and `withRequestStatus` are imported by 6+ async action files. The `withNotifications` / `NotificationContent` re-exports are also consumed locally.

packages/widgets/access-key-management-widget/test/initNotificationsMixin.test.ts
LioriE: Done — deleted all 4 widget-level `initNotificationsMixin.test.ts` files. The sdk-mixins `toastEventsMixin.test.ts` already covers `createToastEventsMixin` behaviour end-to-end.

packages/widgets/access-key-management-widget/src/lib/widget/state/types.ts
LioriE: Done — replaced the local `Notification` type with `ToastNotification` imported from `@descope/sdk-mixins` in all 4 widget `state/types.ts` files.

packages/widgets/user-management-widget/test/config/fileMock.js
LioriE: The new `initNotificationsMixin.ts` imports SVG files for icons. The previous static-string mock caused a TDZ error in CI when those imports were evaluated. Returning `() => document.createElement('span')` makes the mock a factory function, satisfying `expect.any(Function)` in icon assertions.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Done — simplified to use separate `typeIcon`, `closeIcon`, and `body` variables as suggested.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: `createSingletonMixin` constrains its argument to `Mixin = (sc: CustomElementConstructor) => CustomElementConstructor`, so we can't narrow `T` further without breaking that contract. The `declare` pattern is the idiomatic workaround — added a comment documenting that `subscribe` and `actions` come from the widget's `stateManagementMixin`.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Done — renamed to `DEFAULT_SUCCESS_NOTIFICATION_DURATION` and `DEFAULT_ERROR_NOTIFICATION_DURATION`.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Done — `selector` is now optional, defaulting to `(state: any) => state.notifications`. All 4 widget `initNotificationsMixin.ts` files drop the explicit selector.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Done — removed `icons` from `ToastEventsMixinConfig`. Default SVG icons are now baked into `sdk-mixins/src/mixins/toastEventsMixin/icons.ts` as inline strings, so widget callers no longer import or pass SVG files.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/icons.ts
```diff
@@ -0,0 +1,20 @@
+const svgIcon = (svg: string) => () => {
+  const div = document.createElement('div');
+  div.innerHTML = svg;
+  return div.firstElementChild as HTMLElement;
+};
+
+const checkmarkSvg =
+  '<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M10.0485 15.6396C9.5799 15.171 8.8201 15.171 8.35147 15.6396C7.88284 16.1082 7.88284 16.868 8.35147 17.3367L12.7072 21.6924C13.1759 22.1611 13.9357 22.1611 14.4043 21.6924L24.7152 11.3815C25.1838 10.9129 25.1838 10.1531 24.7152 9.68448C24.2466 9.21585 23.4868 9.21585 23.0181 9.68448L13.5558 19.1469L10.0485 15.6396Z" fill="currentcolor"/></svg>';
```
nirgur, line 8: Why not to use svg files?

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        // eslint-disable-next-line class-methods-use-this
+        #createNotificationContent({ type, msg, detail }: ToastNotification) {
+          const typeIcon = defaultIcons[type]();
+          const closeIcon = Object.assign(defaultIcons.close(), {
+            slot: 'close',
+          });
+          const body = detail
+            ? `<div><div>${escapeHtml(msg)}</div>${escapeHtml(detail)}</div>`
+            : escapeHtml(msg);
```
nirgur: Use textContent instead of escapeHtml

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/withToastNotifications.ts
```diff
@@ -0,0 +1,32 @@
...
+}: {
+  // getErrorMsg receives the rejected action; typed as any to avoid inference collision
+  // when the return value is used alongside buildAsyncReducer's FulfilledAction-typed handlers
+  getErrorMsg?: (action: any) => NotificationContent | undefined;
+  getSuccessMsg?: (action: A) => NotificationContent | undefined;
+}) => ({
+  onFulfilled: (state: any, action: A) => {
+    if (!getSuccessMsg) return;
+    const c = normalize(getSuccessMsg(action));
+    if (c.msg) state.notifications.push({ type: 'success', ...c });
```
nirgur: I think we should put the `type` last to avoid override
Same for `type: 'error'`

packages/sdks/react-sdk/src/types.ts
```diff
@@ -45,14 +47,17 @@ type WidgetProps = {
   onReady?: CustomEventCb<{}>;
 };
 
+type WithToastProps = {
```
nirgur:
```suggestion
type ToastProps = {
```

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
```diff
@@ -0,0 +1,129 @@
...
+      class ToastEventsMixinClass extends compose(
+        loggerMixin,
+        notificationsMixin,
+        initLifecycleMixin,
+      )(superclass) {
+        // subscribe and actions are provided at runtime by the widget's stateManagementMixin
+        declare subscribe: <SelectorR>(
+          cb: (state: SelectorR) => void,
+          selector?: (state: any) => SelectorR,
+        ) => () => void;
```
nirgur: This is problematic because it creates an indirect dependency 
anyone who wants to use this mixin must also compose the state mixin 
If they don’t, it will only fail at runtime

I think we should find an alternative way to make this dependency explicit.

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
nirgur: This is problematic, see my other comment about it

For next time, it’s better to leave comments that haven’t been fixed unresolved, so the reviewer can see your replies without needing to expand every comment

packages/widgets/access-key-management-widget/src/lib/widget/state/asyncActions/helpers.ts
```diff
@@ -85,21 +85,5 @@ export const withRequestStatus = (
...
-    if (getSuccessMsg)
-      state.notifications.push({ type: 'success', msg: getSuccessMsg(action) });
-  },
-  onRejected: (state, action) => {
-    if (getErrorMsg)
-      state.notifications.push({ type: 'error', msg: getErrorMsg(action) });
-  },
-});
+export { withToastNotifications as withNotifications } from '@descope/sdk-mixins';
+export type { NotificationContent } from '@descope/sdk-mixins';
```
nirgur, line 89: Why do we need to export it here?

packages/widgets/user-management-widget/test/config/fileMock.js
nirgur: I don't get it, `() => {}` is also a fn
why it's not satisfying `expect.any(Function)`?

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: @nirgur we can address this with `WithStateMixin` type constraint on `T` in `createToastEventsMixin`:

```ts
type WithStateMixin = abstract new (
  ...args: any[]
) => {
  subscribe: <SelectorR>(
    cb: (state: SelectorR) => void,
    selector?: (state: any) => SelectorR,
  ) => () => void;
  actions: { clearNotifications: () => void };
};

// Before:
<T extends CustomElementConstructor>(superclass: T) => ...

// After:
<T extends CustomElementConstructor & WithStateMixin>(superclass: T) => ...
```
 wdyt

LioriE: Fixed in latest commit — added the `WithStateMixin` type constraint on `T` so anyone composing `createToastEventsMixin` without providing a `stateManagementMixin` gets a compile-time error rather than a runtime failure. Also dropped the stale `// subscribe and actions are provided at runtime...` comment since the constraint makes that self-evident.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/withToastNotifications.ts
LioriE: Fixed — both push calls now use `{ ...c, type: 'success' }` and `{ ...c, type: 'error' }` so `type` always wins regardless of what's in the spread.

packages/sdks/react-sdk/src/types.ts
LioriE: Done — renamed `WithToastProps` → `ToastProps` everywhere in `react-sdk/src/types.ts`.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/toastEventsMixin.ts
LioriE: Fixed — `#createNotificationContent` now builds an `HTMLTemplateElement` using DOM APIs. `msg` and `detail` are set via `textContent` / `document.createTextNode`, so they can never be interpreted as HTML. Dropped the `ESC` map and `escapeHtml` helper. Updated the two "escapes HTML" tests to assert on `HTMLTemplateElement` + `textContent`.contains() rather than checking for entity-encoded strings.

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initNotificationsMixin.ts
LioriE: Yes, it's required:  provides  and , which are called inside  at runtime. This dependency is now made explicit at compile time via the  type constraint on  in the latest commit — if you compose  without a state mixin, TypeScript will error.

packages/libs/sdk-mixins/src/mixins/toastEventsMixin/icons.ts
LioriE: Kept inline strings intentionally. `sdk-mixins` is a shared library with no svg build infrastructure (no rollup svg plugin, no `*.svg` type declaration, no jest moduleNameMapper). Converting 3 small one-off icons to `.svg` file imports would require adding all that tooling to the shared lib for no reuse benefit — the icons are not shared elsewhere. Inline strings keep the lib self-contained.

packages/widgets/user-management-widget/test/config/fileMock.js
LioriE: Reverted to `() => {}` per your suggestion.

To answer why the element-returning mock was there: the toast code calls `.outerHTML` and `Object.assign(icon, {slot})` on icon factory return values — if those came through the file mock, getting `undefined` would crash. In practice the icons live in `icons.ts` (TypeScript source), not asset files, so the fileMock never actually touches them. The simpler `() => {}` is fine.

Note: the user-management-widget test suite has a pre-existing TS compile error at `initMixin.ts:36` (`super.init?.()` type error) that predates this PR, so I couldn't fully verify by running the tests.

packages/widgets/access-key-management-widget/src/lib/widget/state/asyncActions/helpers.ts
LioriE: It's a convenience barrel — all sibling action files (`createAccessKey.ts`, `deleteAccessKey.ts`, `activateAccessKey.ts`, `deactivateAccessKey.ts`) import `buildAsyncReducer`, `withRequestStatus`, and `withNotifications` from this single local path rather than mixing a local import with a deep `@descope/sdk-mixins` import. Happy to remove and update those 4 files to import directly from `@descope/sdk-mixins` if you'd prefer.

PR #1383, opened 2026-05-12
feat(web-js-sdk): skip tryRefresh when no login indicator in localStorage

asafshen:
## Summary

- Adds a new `withLoggedInIndicator` enhancer that maintains a `DSLI` localStorage key as a lightweight signal for whether this browser has an active authenticated session
- `sdk.refresh(undefined, true)` (the `tryRefresh` path used by `AuthProvider` on mount) now short-circuits to `{ ok: true }` without a network round-trip when no login indicator is present — eliminating a redundant `/v1/auth/try-refresh` call for anonymous visitors
- Bootstrap fallback: if `DSLI` is absent but `dls_last_user_login_id` exists (written by `withLastLoggedInUser`), the refresh is still attempted — so users who authenticated under a previous SDK version aren't wrongly skipped on their first page load after upgrade

## How it works

**`withLoggedInIndicator` enhancer:**
- Hooks into every API response via `afterRequest`
- Writes `DSLI` to localStorage when `sessionExpiration` is present in the response body (the reliable auth-success signal — JWTs may be HttpOnly cookies, but `sessionExpiration` is always in the body)
- Clears `DSLI` on logout, logoutAll, oidc.logout, and 4xx responses from session-validation routes (`/v1/auth/refresh`, `/v1/auth/try-refresh`)
- Always-on — no opt-in flag needed since this is pure client-side with no server dependency
- `DSLI` key is intentionally unprefixed, matching the `dls_last_user_login_id` convention

**Skip guard in `sdk.refresh`:**
```ts
if (tryRefresh && !hasLoginIndicator()) {
  return Promise.resolve({ ok: true });
}
```

## Test plan

- [ ] `withLoggedInIndicator` test suite (10 tests): writes/clears DSLI correctly across auth, logout, error, prefix, and nested-response scenarios
- [ ] `sdk.test.ts` tryRefresh skip optimization suite (5 tests): verifies skip behavior with/without DSLI and lastUser keys
- [ ] `persistTokens.test.ts` hasLoginIndicator suite (4 tests): unit tests for the indicator helper
- [ ] All 238 web-js-sdk tests pass

### DIFF
--- a/packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
+++ b/packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
@@ -8,8 +8,6 @@ import {
   SESSION_TOKEN_KEY,
   TRUSTED_DEVICE_TOKEN_KEY,
 } from './constants';
-import { LOGGED_IN_INDICATOR_KEY } from '../withLoggedInIndicator/constants';
-import { LOCAL_STORAGE_LAST_USER_LOGIN_ID } from '../withLastLoggedInUser/constants';
 import {
   getLocalStorage,
   removeLocalStorage,
@@ -230,28 +228,6 @@ export function getTrustedDeviceToken(prefix: string = ''): string {
 /** Return the server-returned refresh cookie name from localStorage, if available */
 export function getStoredRefreshCookieName(prefix: string = ''): string | null {
   return getLocalStorage(`${prefix}${REFRESH_COOKIE_NAME_KEY}`);
-}
-
-/**
- * Returns true if the browser shows any sign of an authenticated session.
- *
- * Primary signal: the DSLI key, written by `withLoggedInIndicator` after every
- * successful auth response and cleared on logout / invalid-session.
- *
- * Bootstrap fallback: the lastUser key (`dls_last_user_login_id`) written by
- * `withLastLoggedInUser`. This exists so users authenticated under a previous
- * SDK version (before DSLI existed) aren't wrongly treated as anonymous on the
- * first page load after upgrade. Once any successful auth has written DSLI,
- * the fallback is no longer load-bearing and can be dropped in a future cleanup.
- *
- * Both keys are intentionally unprefixed — `storagePrefix` is consumed by
- * `withPersistTokens` and not propagated to other enhancers.
- */
-export function hasLoginIndicator(): boolean {
-  return (
-    !!getLocalStorage(LOGGED_IN_INDICATOR_KEY) ||
-    !!getLocalStorage(LOCAL_STORAGE_LAST_USER_LOGIN_ID)
-  );
 }
 
 /** Remove auth tokens from localStorage (refresh JWT, session JWT, ID token, server-returned refresh cookie name)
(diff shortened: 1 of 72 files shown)


packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -230,6 +232,28 @@ export function getStoredRefreshCookieName(prefix: string = ''): string | null {
...
+ * Bootstrap fallback: the lastUser key (`dls_last_user_login_id`) written by
+ * `withLastLoggedInUser`. This exists so users authenticated under a previous
+ * SDK version (before DSLI existed) aren't wrongly treated as anonymous on the
+ * first page load after upgrade. Once any successful auth has written DSLI,
+ * the fallback is no longer load-bearing and can be dropped in a future cleanup.
+ *
+ * Both keys are intentionally unprefixed — `storagePrefix` is consumed by
+ * `withPersistTokens` and not propagated to other enhancers.
+ */
+export function hasLoginIndicator(): boolean {
```
nirgur: Why did you decide to put this here instead of under withLoggedInIndicator?

packages/sdks/web-js-sdk/src/sdk/index.ts
```diff
@@ -53,6 +56,22 @@ const createSdk = (config: WebSdkConfig) => {
...
+
+      // Skip the up-front /try-refresh round-trip when localStorage has no sign
+      // of a prior authenticated session. `withLoggedInIndicator` writes DSLI
+      // on every successful auth and clears it on logout / invalid session.
+      // DSLI_DISABLED overrides the skip — escape hatch for apps that hit the
+      // storeLastAuthenticatedUser=false edge case after upgrading.
+      // OIDC mode is handled above and bypasses this optimization, since OIDC
+      // sessions live under a separate `oidc-client-ts` key and `oidc.refreshToken`
+      // performs its own no-session short-circuit.
+      if (
```
nirgur, line 68: Maybe this should be under withAutoRefresh?

nirgur (commented):
Looks good
added 2 small comments

packages/sdks/web-js-sdk/src/sdk/index.ts
asafshen: I don't think its related to auto refresh, (it happens outside the auto-refresh loop, and should work when auto refresh is disabled)

packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
asafshen: absolutely 👍

nirgur (approved):

royair:
Hi @asafshen — filed [#1393](https://github.com/descope/descope-js/issues/1393) reporting that this PR causes useSession() to hang at isSessionLoading=true for fresh users in @descope/react-sdk@2.28.0. Mind taking a look?

asafshen:
Hi @royair , checking

asafshen:
@royair thanks for the feedback, react-sdk version 2.28.2 has a fix

please let me know if something else comes up

PR #1397, opened 2026-05-25
feature: Add tenant style sheet

OfekAvergil:
## Related Issues
https://github.com/descope/etc/issues/5290

## Related PRs
https://github.com/descope/descope-js/pull/1397
https://github.com/descope/backend/pull/1015
https://github.com/descope/console-app/pull/5276

## Description
Adding tenant style sheet and tenant components

## Must

- [ ] 📱 Responsiveness (mobile/XL resolutions)
- [X] 🧪 Tests


packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
         );
       }
 
+      async #fetchTenantTheme() {
+        const tenantId = this.getAttribute('tenant');
+        if (!tenantId) return undefined;
+        if (!/^[A-Za-z0-9_-]+$/.test(tenantId)) {
```
nirgur: We have a mixin for attribute validation, lets use it

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
...
+          this.logger.error(
+            'Invalid tenant attribute: must contain only alphanumeric characters, hyphens, or underscores',
+          );
+          return undefined;
+        }
+        try {
+          const { body: fetchedTenantTheme } = await this.fetchStaticResource(
+            `${tenantId}/theme.json`,
```
nirgur, line 184: Do we support multiple styles for tenant like in the project?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
...
+        } catch (e) {
+          this.logger.error(
+            'Cannot fetch tenant theme file',
+            'make sure that your tenantId, projectId & flowId are correct',
+          );
+        }
+      }
+
+      async #loadTenantStyle() {
+        const projectTheme = await this.#themeResource;
```
nirgur: Why do we need the projectTheme in loadTenantStyle?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
...
+        const tenantTheme = await this.#fetchTenantTheme();
+        const descopeUi = await this.descopeUi;
+
+        if (!tenantTheme) {
+          this.#tenantStyle?.replaceSync('');
+          if (descopeUi?.componentsThemeManager && projectTheme) {
+            descopeUi.componentsThemeManager.themes = {
```
nirgur: Why this is needed? it should already be handled

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
...
+        this.#tenantStyle.replaceSync(
+          (tenantTheme?.light?.globals || '') +
+            (tenantTheme?.dark?.globals || ''),
+        );
+        if (descopeUi?.componentsThemeManager) {
+          const project = projectTheme as Record<string, any> | undefined;
+          descopeUi.componentsThemeManager.themes = {
+            light: {
+              ...project?.light?.components,
+              ...tenantTheme?.light?.components,
```
nirgur: This won't work

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +168,67 @@ export const themeMixin = createSingletonMixin(
...
+          return fetchedTenantTheme;
+        } catch (e) {
+          this.logger.error(
+            'Cannot fetch tenant theme file',
+            'make sure that your tenantId, projectId & flowId are correct',
+          );
+        }
+      }
+
+      async #loadTenantStyle() {
```
nirgur: Let's split this fn to smaller fns like we have for the project theme
In addition, lets reuse the existing fns where possible

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
OfekAvergil: currently not. so reading <teantaId>/theme.json directly

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
         );
       }
 
+      async #fetchTenantTheme() {
+        const tenantId = this.getAttribute('tenant');
```
tomerlichtash: Don't we already have some mixin or utility to get tenant ID?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -18,12 +18,42 @@ const themeValidation = (_: string, theme: string | null) =>
   theme !== 'dark' &&
   'Supported theme values are "light", "dark", or leave empty for using the OS theme';
 
+const tenantValidation = (_: string, tenant: string | null) =>
+  tenant !== null &&
+  !/^[A-Za-z0-9_-]+$/.test(tenant) &&
+  'Invalid tenant attribute: must contain only alphanumeric characters, hyphens, or underscores';
```
tomerlichtash: Probably not the right place to validate tenant name; we need to fix it in a lower level.

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -18,12 +18,42 @@ const themeValidation = (_: string, theme: string | null) =>
...
+  return merged;
+}
+
 export const themeMixin = createSingletonMixin(
   <T extends CustomElementConstructor>(superclass: T) => {
     const BaseClass = compose(
-      createValidateAttributesMixin({ theme: themeValidation }),
+      createValidateAttributesMixin({
+        theme: themeValidation,
+        tenant: tenantValidation,
```
tomerlichtash: Probably not needed

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+          ),
+          dark: this.#mergeComponentThemes(
+            projectThemes?.dark?.components,
+            tenantTheme?.dark?.components,
+          ),
+        };
+      }
+
+      async #loadTenantStyle() {
+        const tenantTheme = await this.#fetchTenantTheme();
```
tomerlichtash: if tenantTheme does not exist, no point in continuing this execution

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
         );
       }
 
+      async #fetchTenantTheme() {
+        const tenantId = this.getAttribute('tenant');
+        if (!tenantId) return undefined;
+        if (tenantValidation('tenant', tenantId)) return undefined;
```
tomerlichtash: Validation probably not necessary.

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+          );
+        }
+      }
+
+      #mergeComponentThemes(
+        base?: Record<string, any>,
+        override?: Record<string, any>,
+      ): Record<string, any> | undefined {
+        if (!base && !override) return undefined;
+        const merged = { ...(base || {}) };
```
tomerlichtash: If we're going to change it, use `let`

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+            'make sure that your tenantId, projectId & flowId are correct',
+          );
+        }
+      }
+
+      #mergeComponentThemes(
+        base?: Record<string, any>,
+        override?: Record<string, any>,
+      ): Record<string, any> | undefined {
+        if (!base && !override) return undefined;
```
tomerlichtash: If no override supplied - no need to execute merge

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+          return fetchedTenantTheme;
+        } catch (e) {
+          this.logger.error(
+            'Cannot fetch tenant theme file',
+            'make sure that your tenantId, projectId & flowId are correct',
+          );
+        }
+      }
+
+      #mergeComponentThemes(
```
tomerlichtash: Can we combine this logic as another nesting level in deepMergeNonEmpty?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+        if (!base && !override) return undefined;
+        const merged = { ...(base || {}) };
+        for (const [component, value] of Object.entries(override || {})) {
+          if (!value || Object.keys(value).length === 0) continue;
+          merged[component] = deepMergeNonEmpty(merged[component] || {}, value);
+        }
+        return merged;
+      }
+
+      async #loadTenantGlobalStyle(tenantTheme?: Record<string, any>) {
```
tomerlichtash: This logic that clears/injects styles into custom stylesheets is a shared behavior of custom theme and tenant theme; would it better if we had a custom stylesheet mixin at this point? @nirgur, wdyt?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+        }
+        this.#tenantStyle.replaceSync(
+          (tenantTheme?.light?.globals || '') +
+            (tenantTheme?.dark?.globals || ''),
+        );
+      }
+
+      async #loadTenantComponentsStyle(tenantTheme?: Record<string, any>) {
+        if (!tenantTheme) {
+          await this.#loadComponentsStyle();
```
tomerlichtash: Looks like we're re-loading components style if tenant theme not exists; looks like a redundant call.

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +197,80 @@ export const themeMixin = createSingletonMixin(
...
+          return;
+        }
+
+        const descopeUi = await this.descopeUi;
```
tomerlichtash: Do we need these validations for descopeUi at this point?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -237,7 +341,6 @@ export const themeMixin = createSingletonMixin(
         this.#applyTheme();
       }
 
-      // add or remove os theme change listener
```
tomerlichtash, line 240: why remove this comment? :)

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -253,17 +356,25 @@ export const themeMixin = createSingletonMixin(
...
 
-        this.observeAttributes(['theme-override'], () => this.#loadCustomStyle());
+        this.observeAttributes(['theme-override'], () =>
+          this.#loadCustomStyle(),
+        );
+
+        this.observeAttributes(['tenant'], () => this.#loadTenantStyle());
 
-        this.observeAttributes(['style-id'], () => {
+        this.observeAttributes(['style-id'], async () => {
```
tomerlichtash: Why async? Why not run this load fns in sync?

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -55,6 +61,10 @@ export const themeMixin = createSingletonMixin(
         return this.getAttribute('style-id') || DEFAULT_STYLE_ID;
       }
 
+      get tenant(): string | null {
```
nirgur: I think it will be better to have a `tenantIdMixin`, just like `projectIdMixin`
we should use the `createValidateAttributesMixin` and make sure the tenantId contains only valid characters (not sure which exactly)

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +176,65 @@ export const themeMixin = createSingletonMixin(
...
+          this.logger.error(
+            'Cannot fetch tenant theme file',
+            'make sure that your tenantId, projectId & flowId are correct',
+          );
+        }
+      }
+
+      async #loadTenantGlobalStyle(tenantTheme: Record<string, any>) {
+        if (!this.#tenantStyle) {
+          this.#tenantStyle = this.injectStyle('');
```
nirgur: There is no reason to create an empty stylesheet and then use replace sync,
let's create it with the wanted styles

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -167,6 +176,65 @@ export const themeMixin = createSingletonMixin(
...
+            (tenantTheme.dark?.globals || ''),
+        );
+      }
+
+      async #loadTenantComponentsStyle(tenantTheme: Record<string, any>) {
+        const descopeUi = await this.descopeUi;
+        const projectThemes = await this.#themeResource;
+        if (!projectThemes) return;
+        if (descopeUi?.componentsThemeManager) {
+          descopeUi.componentsThemeManager.themes = {
```
nirgur: We are reloading the components styles twice, once for the project only and once with the tenant style

Let's update the loadComponentsStyle fn to handle also the tenant components styles

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -253,17 +321,23 @@ export const themeMixin = createSingletonMixin(
           this.#loadGlobalStyle(),
           this.#loadComponentsStyle(),
         ]);
+        await this.#loadTenantStyle();
         await this.#loadCustomStyle();
```
nirgur: This is not really async, no need to wait for it, should remain last

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -253,17 +321,23 @@ export const themeMixin = createSingletonMixin(
           this.#loadGlobalStyle(),
           this.#loadComponentsStyle(),
         ]);
+        await this.#loadTenantStyle();
```
nirgur: Here we are really waiting for the tenant theme to be fetched, 
but instead of calling it only after the project theme is fetched, we can inject the project theme style with `{prepend: true}` so it will be injected before the tenant theme even when the tenant theme is fetched before

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -253,17 +321,23 @@ export const themeMixin = createSingletonMixin(
...
         await this.#loadCustomStyle();
 
         this.observeAttributes(['theme'], this.#onThemeChange);
 
-        this.observeAttributes(['theme-override'], () => this.#loadCustomStyle());
+        this.observeAttributes(['theme-override'], () =>
+          this.#loadCustomStyle(),
+        );
+
+        this.observeAttributes(['tenant'], () => this.#loadTenantStyle());
```
nirgur: Now, when the tenant changes we should load component styles & tenant styles

packages/libs/sdk-mixins/src/mixins/themeMixin/themeMixin.ts
```diff
@@ -253,17 +321,23 @@ export const themeMixin = createSingletonMixin(
...
 
         this.observeAttributes(['style-id'], () => {
           this.#_themeResource = null;
           this.#loadFonts();
           this.#loadGlobalStyle();
           this.#loadComponentsStyle();
+          this.#loadTenantStyle();
```
nirgur: There is no need to call it here anymore

nirgur (approved):

PR #1400, opened 2026-05-28
feat: tenant multi sso

LioriE:
## Summary

Adds multi-SSO configuration support to the `tenant-profile-widget`.

- Integrates the `descope-multi-sso` component — displays the tenant's SSO configurations as a list with create and delete actions via configurable flow modals
- `getTenantAdminLinkSSO` now calls a batch endpoint (`/batch`), accepts the tenant's SSO config IDs, and returns a `defaultLink` + a per-SSO `ssoIdToLink` map instead of a single link
- SSO configurations (default + additional) are derived in a `getSsoConfigurations` selector that merges tenant data with the per-SSO admin links, and pushed to the component as `data`
- `getTenantAdminLinkSSO` is called sequentially after `getTenant` so SSO IDs from the tenant response are available when constructing the batch request

image


packages/widgets/tenant-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initMultiSsoConfigurationsMixin.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+      }
+
+      async onWidgetRootReady() {
+        await super.onWidgetRootReady?.();
+
+        const ele = this.shadowRoot?.querySelector('descope-multi-sso');
+        if (!ele) return;
+
+        this.#wireComponent(ele);
```
nirgur: Let's create a driver for it like we have for the other components and move the relevant component logic to it

packages/widgets/tenant-profile-widget/src/lib/widget/state/selectors.ts
```diff
@@ -36,3 +39,32 @@ export const getTenantEnforceSSO = createSelector(
   getTenant,
   (tenant) => tenant.enforceSSO || false,
 );
+
+export const getSsoConfigurations = createSelector(
+  getTenant,
+  (state: State) => state.tenantAdminLinkSSO.data,
```
nirgur: Let's use the relevant selector

nirgur (approved):

PR #1403, opened 2026-06-02
feat(widgets): add widget localization

asafshen:
## Description

Adds widget localization to the SDK so widgets render in the user's locale, consuming the per-locale screen artifacts (`<page>-<locale>.html`) and the `config.json` `widgets[].targetLocales` published by the backend. Pairs with the backend and console-app widget-localization work.

- **sdk-mixins**: new `localeMixin` (`locale` / `resolvedLocale`, mirroring the web-component `getUserLocale` resolution so widget locale matches the flow runtime) and `widgetConfigMixin.isLocaleAvailable(locale)` backed by `config.json` `widgets[].targetLocales` (case-insensitive). `WidgetConfig` gains `version?` / `targetLocales?`.
- **all 8 widgets** (`fetchWidgetPagesMixin`): fetch `<page>-<locale>.html` when the widget publishes that locale, falling back to the default page on miss or fetch error. Unified onto `compose(staticResourcesMixin, widgetIdMixin, widgetConfigMixin, localeMixin)`.
- **flow-embedding widgets** (user-profile, tenant-profile, user-management, outbound-applications): propagate an explicit widget `locale` attribute into embedded `descope-wc` flows via `createFlowTemplate`. When no explicit `locale` is set, the embedded flow keeps its own `navigator.language` fallback.

## Tests
- `localeMixin.test.ts`, `widgetConfigMixin.test.ts` (sdk-mixins)
- `fetchWidgetPagesMixin.test.ts` (user-profile-widget): localized fetch, case-insensitive match, fallback when locale unavailable, fallback when localized fetch throws

## Must
- [x] Tests

### DIFF
--- a/packages/libs/sdk-helpers/src/locale.ts
+++ b/packages/libs/sdk-helpers/src/locale.ts
@@ -4,13 +4,13 @@ export type Locale = {
 };
 
 /**
- * Resolves the user-facing locale. Mirrors web-component's getUserLocale exactly so widget
- * locale resolution matches the flow runtime byte-for-byte.
+ * Resolves the user-facing locale, shared by the web-component flow runtime and the widget mixins
+ * so both resolve locale identically.
  *
  * Order:
- *  1. Explicit locale argument (lowercased)
+ *  1. Explicit locale argument (lowercased), used as its own fallback
  *  2. navigator.language (lowercased), with the language part as fallback for 'xx-YY' forms
- *  3. Empty string
+ *  3. Empty strings
  */
 export function getUserLocale(locale: string): Locale {
   if (locale) {
--- a/packages/libs/sdk-mixins/src/mixins/fetchWidgetPagesMixin/createFetchWidgetPagesMixin.ts
+++ b/packages/libs/sdk-mixins/src/mixins/fetchWidgetPagesMixin/createFetchWidgetPagesMixin.ts
@@ -1,0 +1,68 @@
+import { compose, createSingletonMixin } from '@descope/sdk-helpers';
+import { localeMixin } from '../localeMixin';
+import { staticResourcesMixin } from '../staticResourcesMixin';
+import { widgetConfigMixin } from '../widgetConfigMixin';
+import { widgetIdMixin } from '../widgetIdMixin';
+
+/**
+ * Builds a widget's fetchWidgetPagesMixin. Every widget shares identical logic and differs only by
+ * its published base directory (the widget type/name under which its screens live in S3), so each
+ * widget calls this with its own dir:
+ *
+ *   export const fetchWidgetPagesMixin = createFetchWidgetPagesMixin('user-profile-widget');
+ *
+ * `fetchWidgetPage` fetches `<base>/<widgetId>/<file>`, transparently preferring a localized
+ * `<file>-<locale>.html` variant when the widget publishes that locale (config.json targetLocales),
+ * and falling back to the default page otherwise.
+ */
+export const createFetchWidgetPagesMixin = (widgetPagesBaseDir: string) =>
+  createSingletonMixin(<T extends CustomElementConstructor>(superclass: T) => {
+    const BaseClass = compose(
+      staticResourcesMixin,
+      widgetIdMixin,
+      widgetConfigMixin,
+      localeMixin,
+    )(superclass);
+    return class FetchWidgetPagesMixinClass extends BaseClass {
+      // Keep `init` typed as a method across the package boundary (mirrors widgetConfigMixin):
+      // without it, the emitted .d.ts widens the inherited `init` to a property, and a widget
+      // composing this alongside another init-providing mixin (e.g. initLifecycleMixin) hits TS2425.
+      async init() {
+        await super.init?.();
+      }
+
+      async fetchWidgetPage(filename: string) {
+        const localized = await this.#tryFetchLocalized(filename);
+        if (localized !== undefined) return localized;
+
+        const res = await this.fetchStaticResource(
+          `${widgetPagesBaseDir}/${this.widgetId}/${filename}`,
+          'text',
+        );
+        return res?.body;
+      }
+
+      // Fetch <page>-<locale>.html when the widget publishes that locale (config.json
+      // targetLocales); otherwise return undefined so the caller falls back to the default page.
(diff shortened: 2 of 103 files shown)


packages/libs/sdk-helpers/src/locale.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!nl) {
+    return { locale: '', fallback: '' };
+  }
+
+  if (nl.includes('-')) {
+    return {
+      locale: nl.toLowerCase(),
+      fallback: nl.split('-')[0].toLowerCase(),
+    };
+  }
```
nirgur, line 29: Why do we treat the received locale differently from navigator.language? 
Is this how it works today?

packages/libs/sdk-mixins/src/mixins/widgetConfigMixin/widgetConfigMixin.ts
```diff
@@ -20,6 +20,20 @@ export const widgetConfigMixin = createSingletonMixin(
         const config = await this.config;
         return config?.projectConfig?.widgets?.[this.widgetId];
       }
+
+      // The first candidate locale for which the widget has published localized screens
+      // (case-insensitive), per the widget's targetLocales in config.json. Returns '' when none
+      // match. Callers pass candidates most-specific first (e.g. ['en-us', 'en']) so a region
+      // locale falls back to its language, mirroring the web-component's resolution.
+      async firstAvailableLocale(candidates: string[]): Promise<string> {
```
nirgur: I think this fn should be on localeMixin and not here

packages/libs/sdk-mixins/src/mixins/localeMixin/localeMixin.ts
```diff
@@ -0,0 +1,24 @@
+import { createSingletonMixin, getUserLocale } from '@descope/sdk-helpers';
```
nirgur, line 1: If we call `getUserLocale` only here, maybe it should be a private member of this mixin class?

packages/libs/sdk-mixins/src/mixins/fetchWidgetPagesMixin/createFetchWidgetPagesMixin.ts
```diff
@@ -0,0 +1,68 @@
...
+ * its published base directory (the widget type/name under which its screens live in S3), so each
+ * widget calls this with its own dir:
+ *
+ *   export const fetchWidgetPagesMixin = createFetchWidgetPagesMixin('user-profile-widget');
+ *
+ * `fetchWidgetPage` fetches `<base>/<widgetId>/<file>`, transparently preferring a localized
+ * `<file>-<locale>.html` variant when the widget publishes that locale (config.json targetLocales),
+ * and falling back to the default page otherwise.
+ */
+export const createFetchWidgetPagesMixin = (widgetPagesBaseDir: string) =>
```
nirgur, line 18: Can we make it even more generic and use it also for flows?

packages/widgets/access-key-management-widget/src/lib/widget/mixins/fetchWidgetPagesMixin.ts
nirgur, line 1: Do we really need this file? (relevant for other similar files as well)

packages/widgets/access-key-management-widget/src/lib/widget/mixins/fetchWidgetPagesMixin.ts
```diff
@@ -1,31 +1,5 @@
...
-        const res = await this.fetchStaticResource(
-          `${WIDGET_PAGES_BASE_DIR}/${this.widgetId}/${filename}`,
-          'text',
-        );
-        return res?.body;
-      }
-    };
-  },
+export const fetchWidgetPagesMixin = createFetchWidgetPagesMixin(
+  'access-key-management-widget',
```
nirgur: Let's put it in a const (relevant also for the other widgets)

nirgur (commented):
Looks good 👍 
One question, don't we need to update the FE sdks to expose the new locale attribute?
e.g. in React we are listing the attributes and `locale` is not one of them

packages/libs/sdk-helpers/src/locale.ts
asafshen: yes, this is a refactor so we can use the same code in both descope-wc and in widgets 
 I didn't reason about it too much, just copied it from here:
Image

packages/libs/sdk-mixins/src/mixins/fetchWidgetPagesMixin/createFetchWidgetPagesMixin.ts
asafshen: I can try 👍

packages/widgets/access-key-management-widget/src/lib/widget/mixins/fetchWidgetPagesMixin.ts
asafshen: we can use inline instead 👍

packages/widgets/access-key-management-widget/src/lib/widget/mixins/fetchWidgetPagesMixin.ts
asafshen: will change to const WIDGET_PAGES_BASE_DIR = '....'

packages/libs/sdk-mixins/src/mixins/fetchWidgetPagesMixin/createFetchWidgetPagesMixin.ts
asafshen: it was a bit more complicated to do then I thought, leaving this aside

packages/libs/sdk-mixins/src/mixins/localeMixin/localeMixin.ts
nirgur: Don't you want to move it here?

asafshen: sorry, forgot to communicte that its also used in web-component 
image

nirgur (approved):

PR #1410, opened 2026-06-09
fix(core-js-sdk): declare tslib as a dependency

itaihanski:
## Related Issues
Required for:
https://github.com/descope/etc/issues/16068

### Related PRs
- descope/descope-js#1408 — earlier core-js fix for the same issue (async-arrow Metro bug)
- descope/descope-react-native#172 — consumes the fixes

## In a Nutshell
- Declare `tslib` in `core-js-sdk` `dependencies` (used via TS `importHelpers`, never declared)
- Fixes strict installs (bun / pnpm) that don't resolve undeclared transitive deps

## Description
The compiled output imports `__rest` from `tslib` (TypeScript's `importHelpers`), but `tslib` was never listed in `dependencies` — only `jwt-decode` was. Loose, hoisted installers (npm / yarn-classic) resolve it transitively, so it went unnoticed; strict installers (**bun**, pnpm) don't, forcing consumers to `bun add tslib` by hand. This was the second of the two requirements in the customer report (the first — the Metro async-arrow bundling bug — shipped in 2.62.1 / #1408).

## Must
- [ ] Tests

### DIFF
--- a/packages/sdks/core-js-sdk/package.json
+++ b/packages/sdks/core-js-sdk/package.json
@@ -82,7 +82,8 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
-    "jwt-decode": "4.0.0"
+    "jwt-decode": "4.0.0",
+    "tslib": "2.8.1"
   },
   "overrides": {
     "terser": "^5.14.2"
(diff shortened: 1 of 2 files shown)


packages/sdks/core-js-sdk/package.json
```diff
@@ -82,7 +82,8 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
-    "jwt-decode": "4.0.0"
+    "jwt-decode": "4.0.0",
+    "tslib": "2.8.1"
```
nirgur:
```suggestion
    "tslib": "^2.8.1"
```

PR #1414, opened 2026-06-17
feat(access-key-management-widget): add Rotate action RELEASE

nirgur:
## Summary

- New `rotate({id})` SDK call hitting `POST /v1/mgmt/accesskey/rotate`.
- New confirm dialog and dedicated cleartext reveal modal — both are new screen IDs (`rotate-access-key-modal`, `rotated-access-key-modal`) seeded onto the widget by the console-app screen-injection on save.
- Single-active-key gating: `getCanRotateAccessKey` selector composes the existing `canChangeStatus + isSingleSelected` and adds `status === 'active'` (which the shared selector intentionally doesn't gate on because it's used by Activate too).

## Linked

- Ticket: descope/etc#15738
- Related PRs:
  - descope/backend — https://github.com/descope/backend/pull/1331
  - descope/console-app — https://github.com/descope/console-app/pull/5410

## Changes

- `apiPaths.ts`: `rotate: '/v1/mgmt/accesskey/rotate'`.
- `createAccessKeySdk.ts`, `mocks.ts`, `types.ts`: SDK function + mock + `RotateAccessKeyConfig` type.
- `state/asyncActions/rotateAccessKey.ts`: new async thunk. `onFulfilled` replaces the matching key in the list with the BE response (defensive against any BE-side metadata updates on rotate).
- `state/selectors.ts`: `getCanRotateAccessKey`.
- `state/initialState.ts`, `state/types.ts`, `stateManagementMixin.ts`: state slice + wiring.
- `initMixin/initComponentsMixins/initRotateAccessKeyButtonMixin.ts`: new button mixin — click opens the confirm modal.
- `initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts`: new confirm dialog. Submit fires the rotate action; on success opens the reveal modal.
- `initMixin/initComponentsMixins/initRotatedAccessKeyModalMixin.ts`: new dedicated cleartext reveal — title "Access key secret rotated" (vs the create flow's "Access key was created").
- `test/access-key-management-widget.test.ts`: SDK + selector + reducer tests.
- `e2e/access-key-management-widget.spec.ts`: 5 new Playwright tests (happy path, API failure, cancel, multi-select disabled, expired disabled).

## Manual test plan

- Local: build + run widget against a local BE with the matching permission entry from the companion `descope/backend` PR.
- Save an access-key widget in console-app (so the new modal screens land on S3).
- In the widget runtime: select an active key → click Rotate → confirm modal opens with "Rotate `<name>`?" → click Rotate → API fires → reveal modal opens with "Access key secret rotated" + new cleartext → "Copy & close" puts the secret in the clipboard.
- Verify the existing Create flow still uses its own reveal modal ("Access key was created").

## Risk / review focus

- The two new mixins (`initRotateAccessKeyModalMixin`, `initRotatedAccessKeyModalMixin`) mirror existing deactivate/created mixins closely — same modal-driver patterns, same form-data plumbing.
- Reveal-modal duplication: the create and rotate reveal modals share structure but are deliberately separate so each title fits its flow.
- Cross-repo coupling: this PR is inert (rotate button no-ops) until both the console-app PR and the backend PR are merged. Merge order: backend → either FE in either order.

### DIFF
--- a/packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
+++ b/packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
@@ -17,9 +17,6 @@ import { stateManagementMixin } from '../../stateManagementMixin';
 import { initRotatedAccessKeyModalMixin } from './initRotatedAccessKeyModalMixin';
 import { initWidgetRootMixin } from './initWidgetRootMixin';
 
-// Confirm dialog for the Rotate flow — mirrors initDeactivateAccessKeysModalMixin.
-// Submit calls actions.rotateAccessKey({ id }) and, on success, opens the
-// dedicated rotated-access-key reveal modal with the new cleartext.
 export const initRotateAccessKeyModalMixin = createSingletonMixin(
   <T extends CustomElementConstructor>(superclass: T) =>
     class InitRotateAccessKeyModalMixinClass extends compose(
--- a/packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
+++ b/packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
@@ -1,0 +1,97 @@
+import {
+  ButtonDriver,
+  ModalDriver,
+  TextDriver,
+} from '@descope/sdk-component-drivers';
+import {
+  compose,
+  createSingletonMixin,
+  createTemplate,
+} from '@descope/sdk-helpers';
+import { formMixin, loggerMixin, modalMixin } from '@descope/sdk-mixins';
+import {
+  getSelectedAccessKeys,
+  getSelectedAccessKeysDetailsForDisplay,
+} from '../../../state/selectors';
+import { stateManagementMixin } from '../../stateManagementMixin';
+import { initRotatedAccessKeyModalMixin } from './initRotatedAccessKeyModalMixin';
+import { initWidgetRootMixin } from './initWidgetRootMixin';
+
+// Confirm dialog for the Rotate flow — mirrors initDeactivateAccessKeysModalMixin.
+// Submit calls actions.rotateAccessKey({ id }) and, on success, opens the
+// dedicated rotated-access-key reveal modal with the new cleartext.
+export const initRotateAccessKeyModalMixin = createSingletonMixin(
+  <T extends CustomElementConstructor>(superclass: T) =>
+    class InitRotateAccessKeyModalMixinClass extends compose(
+      stateManagementMixin,
+      modalMixin,
+      formMixin,
+      loggerMixin,
+      initWidgetRootMixin,
+      initRotatedAccessKeyModalMixin,
+    )(superclass) {
+      rotateAccessKeyModal: ModalDriver;
+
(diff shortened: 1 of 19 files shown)


packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
```diff
@@ -0,0 +1,97 @@
...
+  getSelectedAccessKeys,
+  getSelectedAccessKeysDetailsForDisplay,
+} from '../../../state/selectors';
+import { stateManagementMixin } from '../../stateManagementMixin';
+import { initRotatedAccessKeyModalMixin } from './initRotatedAccessKeyModalMixin';
+import { initWidgetRootMixin } from './initWidgetRootMixin';
+
+// Confirm dialog for the Rotate flow — mirrors initDeactivateAccessKeysModalMixin.
+// Submit calls actions.rotateAccessKey({ id }) and, on success, opens the
+// dedicated rotated-access-key reveal modal with the new cleartext.
```
nirgur: Is this comment needed?

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
```diff
@@ -0,0 +1,97 @@
...
+          if (!selected) {
+            this.rotateAccessKeyModal.close();
+            return;
+          }
+          const res: Record<string, any> = await this.actions.rotateAccessKey({
+            id: selected.id,
+          });
+          this.rotateAccessKeyModal.close();
+
+          if (res?.payload?.cleartext) {
```
nirgur, line 67: Isn't it better to take it from the state using a selector?

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
nirgur: Dropped — the class name and body already said it. Fixed in `b2fef5b0`. (auto via /pr-comments)

packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initRotateAccessKeyModalMixin.ts
nirgur: Two reasons I would lean toward keeping as-is:

1. **Mirrors the existing pattern** — `initCreateAccessKeyModalMixin` reads `res?.payload?.cleartext` from the awaited thunk the same way ([lines 59-75](https://github.com/descope/descope-js/blob/main/packages/widgets/access-key-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initCreateAccessKeyModalMixin.ts)). Matching the pattern means we keep one shape across the widget.
2. **Scope of the secret** — the cleartext is a one-time secret. Reading it from the resolved action keeps it in a local variable that goes out of scope when the handler returns. Storing it in Redux state would mean it lives there until something clears it (modal close, reset, etc.) — bigger surface area.

Happy to move both flows to a selector-based approach if you would rather have the Redux-idiomatic shape — I just would not want two different patterns side by side. WDYT?

tomerlichtash (approved):

PR #1415, opened 2026-06-17
feat(widgets/user-profile-widget): native bridge for mobile SDKs

itaihanski:
## Related Issues
https://github.com/descope/etc/issues/15522

## Description
- Add `nativeBridgeMixin` that owns all mobile-bridge concerns: detects `window.descopeWidgetBridge`, performs the lazy-init handshake (native calls `widget.lazyInit()` to release), and delegates logout to native when present
- `nativeBridgeMixin` integrates via `super` overrides on `init()` and `handleLogout()` — `initWidgetRootMixin` and `initLogoutMixin` stay bridge-agnostic; `handleLogout()` is the named extraction point on the logout mixin that the bridge wraps
- Bump `<descope-wc>` to `bridgeVersion = 3` with opt-in `descopeBridge.register(this)` / `unregister(handle)` so multi-component native bridges can address each instance — single-flow bridges that don't implement these are unaffected (optional chain no-ops)
- Init failures dispatch a standard `error` CustomEvent on the widget element; native and host pages both listen on the same surface

## Must
- [x] Tests


packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initLogoutMixin.ts
```diff
@@ -22,6 +26,11 @@ export const initLogoutMixin = createSingletonMixin(
         );
 
         this.logout.onClick(async () => {
+          // In native mode the host SDK owns logout
+          if (isNativeBridgeAvailable()) {
+            requestNativeLogout();
+            return;
+          }
```
nirgur: I think it's better to override this fn from the bridge and not iject mobile related logic into the widget

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initWidgetRootMixin.ts
```diff
@@ -39,9 +41,32 @@ export const initWidgetRootMixin = createSingletonMixin(
 
       async init() {
         await super.init?.();
-        await this.actions.getMe();
-        await this.#initWidgetRoot();
-        await this.onWidgetRootReady();
+        // When running inside a native mobile host, we defer the init to allow native SDK
```
nirgur: Instead of changing the initWidgetRootMixin, we can keep this logic in the nativeBridgeMixin, by overriding the init fn, inject the logic we need, and call super.init()

packages/widgets/user-profile-widget/test/native-bridge.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  afterEach(() => {
+    document.body.innerHTML = '';
+    delete (window as any).descopeWidgetBridge;
+  });
+
+  describe('waitForNativeBridgeIfNeeded', () => {
+    it('resolves immediately when no bridge is installed (web mode)', async () => {
+      await expect(
+        element.waitForNativeBridgeIfNeeded(),
+      ).resolves.toBeUndefined();
```
nirgur: It will always resolve as undefined, maybe we should think about a different way to test it?

nirgur (dismissed):

itaihanski:
@shuni final pass after all changes

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur, line 1: Can we move the mobile related logic to the nativeBridgeMixin?

itaihanski: What do you mean? It's all there

nirgur: what we added in this file

itaihanski: The mixin is for the user profile, this change the the wc is to replace the polling init with a registration. Regardless of userprofilewidget, we're changing how init works, just slightly.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -410,6 +415,9 @@ class DescopeWc extends BaseDescopeWc {
     }
     // eslint-disable-next-line no-underscore-dangle
     (this as any).lazyInit = this._init;
+    // from bridge version 3 onwards, wc registers itself with the descopeBridge instead
+    // of the previous polling mechanism, which still works for backwards compat
+    this.#bridgeKey = (window as any).descopeBridge.registerFlow?.(this);
     return undefined;
   }
```
nirgur:
```suggestion
    this.#bridgeKey = (window as any)?.descopeBridge?.registerFlow?.(this);
return undefined;
}
```

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: 👍 Sorry for the confusion

nirgur (approved):

Issue #1393, opened 2026-05-24
useSession() hangs with isSessionLoading=true for fresh users after #1383 (web-js-sdk@1.49.0)


royair:
## Bug: `AuthProvider` hangs with `isSessionLoading=true` for fresh users after upgrade to
  `@descope/web-js-sdk@1.49.0`

  ### Versions

  | Package | Broken | Last working |
  |---|---|---|
  | `@descope/react-sdk` | `2.28.0` | `2.27.10` |
  | `@descope/web-js-sdk` (transitive) | `1.49.0` | `1.48.2` |

  ### Symptom

  When a user with **no Descope localStorage state** loads an app wrapped in `<AuthProvider>`:

  - `useSession()` returns `{ isAuthenticated: false, isSessionLoading: true }` and **never flips
  `isSessionLoading` to `false`**.
  - Any component gated on `!isAuthenticated && !isSessionLoading` (e.g., a "redirect to /login" guard) is
  stuck — the redirect never fires, and the app sits on whatever route the user landed on with an empty
  container.
  - Affects **production builds** of consumer apps. In dev mode (e.g., `vite` dev server), the bug is often
  masked.
  - Affects **fresh storage state** — cleared site data, first-time visitors, E2E test users between runs.
  Returning users with cached localStorage are unaffected.

  ### Minimal reproduction

  A standard `AuthProvider` + `useSession`-gated route, identical to the pattern in the README:

  ```tsx
  // App.tsx
  import { AuthProvider } from "@descope/react-sdk";
  import { BrowserRouter, Routes, Route } from "react-router-dom";

  export default function App() {
    return (
      <AuthProvider projectId="YOUR_PROJECT_ID">
        <BrowserRouter>
          <Routes>
            <Route path="/" element={<ProtectedRoute><Home /></ProtectedRoute>} />
            <Route path="/login" element={<LoginPage />} />
          </Routes>
        </BrowserRouter>
      </AuthProvider>
    );
  }

  // ProtectedRoute.tsx
  import { useSession } from "@descope/react-sdk";
  import { useEffect } from "react";
  import { useNavigate } from "react-router-dom";

  export const ProtectedRoute = ({ children }) => {
    const { isAuthenticated, isSessionLoading } = useSession();
    const navigate = useNavigate();

    useEffect(() => {
      if (!isAuthenticated && !isSessionLoading) {
        navigate("/login");
      }
    }, [isAuthenticated, isSessionLoading, navigate]);

[...]

PR #1416, opened 2026-06-18
fix(react-sdk): clear loading state when AuthProvider refresh/me reject

ophirt:
## Problem
`AuthProvider`'s `fetchSession` and `fetchUser` attach only an `onFulfilled`
handler to `sdk.refresh()` / `sdk.me()`:

```ts
withValidation(sdk?.refresh)(undefined, true).then(() => {
  /* ... setIsSessionLoading(false) ... */
});
```

`withValidation` only invokes the underlying SDK function — it does not catch.
So when `refresh()` (or `me()`) **rejects** (e.g. a transient network failure
during the proactive refresh on mount), the `.then` callback never runs and
`setIsSessionLoading(false)` / `setIsUserLoading(false)` is never called.

Since `isSessionFetched` / `isUserFetched` are set `true` *before* the call,
`fetchSession` / `fetchUser` never re-run — so `useSession().isSessionLoading`
(and `useUser().isUserLoading`) stay `true` **forever**, pinning the app on its
loading state until a full remount, even when a valid session cookie/token is
still present (the rejected call was only a background refresh).

## Reproduction
1. Render an app with `AuthProvider` and a valid session.
2. Make the initial `POST /v1/auth/try-refresh` fail at mount (offline, or a
   proxy that drops the cross-origin request).
3. `useSession().isSessionLoading` stays `true` indefinitely.

## Fix
Run the loading-state cleanup on **both** fulfilment and rejection
(`.then(onDone, onDone)`). Providing the rejection handler also means the
rejection is handled (no unhandled promise rejection).

- Minimal; preserves the existing `#1393` deferred-`setIsSessionLoading(false)`.
- Does not change the "fetch once" semantics — a failed proactive refresh falls
  back to the existing session state and `autoRefresh` retries on schedule.

Happy to add a unit test if you point me at the preferred place / mocking setup.

### DIFF
--- a/packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
+++ b/packages/sdks/react-sdk/src/components/AuthProvider/AuthProvider.tsx
@@ -149,7 +149,7 @@ const AuthProvider: FC<IAuthProviderProps> = ({
     isSessionFetched.current = true;
 
     setIsSessionLoading(true);
-    withValidation(sdk?.refresh)(undefined, true).then(() => {
+    const stopSessionLoading = () => {
       // Defer to a separate render pass so React doesn't batch this with
       // the setIsSessionLoading(true) above when refresh() short-circuits
       // synchronously - downstream consumers (e.g. useSession) need to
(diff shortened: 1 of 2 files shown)


asafshen:
@ophirt thank you for the contribution 🙏 we are looking into that

asafshen:
@ophirt I reviewed your PR and it looks great (kudos for the clean implementation)

We also want to add test for this 
Do you want to do it? If you prefer I can do it )

ophirt:
Done 🙏 added `test/components/AuthProviderSessionLoading.test.tsx` covering both paths:

- `isSessionLoading` is cleared when the initial `refresh()` rejects
- `isUserLoading` is cleared when `me()` rejects

I mirrored the existing `App.test.tsx` `createSdk` mock harness. Both tests **fail against `main`** (loading stays stuck `true`) and **pass with the fix**, so they guard against a regression. Happy to adjust naming/placement if you'd prefer them folded into `AuthProvider.test.tsx`.

asafshen:
Hey @ophirt - I've copied your PR over to #1429 because the build fails on fork PRs (a permissions issue we're fixing separately), and I wanted to make sure your contribution makes it in as soon as possible. Your commits are preserved as-is, so the authorship stays yours. Huge thanks for the fix!

PR #1422, opened 2026-06-21
feat(widgets): Support UserMgmt Filter

tomerlichtash:
## Related Issues

Fixes https://github.com/descope/etc/issues/14302

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| Console App | [Link to PR](https://github.com/descope/console-app/pull/5433) |

## Description

Support Filter component in user management widget. Full doc, architecture an how-to here: https://github.com/descope/etc/issues/14302#issuecomment-5178241391

## Must

- [ ] Tests


packages/widgets/user-management-widget/src/lib/widget/state/filterToSearchParams.ts
```diff
@@ -0,0 +1,94 @@
...
+// Seeds all touched fields with `undefined` so a removed row clears its
+// value instead of being retained by the searchUsers thunk merge.
+export const filterToSearchParams = (
+  rows: FilterRow[],
+): Partial<SearchUsersConfig> => {
+  const params: Partial<SearchUsersConfig> = {};
+  [
+    ...Object.values(ARRAY_FIELDS),
+    ...Object.values(EXACT_FIELDS),
+    'text' as const,
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in f99bc01d — stopped seeding `text` in filterToSearchParams. It's co-owned by the standalone search input, so the filter no longer clears the user's typed query on apply; it only sets `text` for its own full-text rows. Regression test added.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,94 @@
...
+};
+
+// Translate descope-filter `filter-apply` event detail.value into native
+// SearchUsersRequest fields. Negation (`not-any-of`/`not-equal`/etc) is
+// unsupported by the endpoint and dropped — restrict via column `operators`
+// allowlist in console-app metadata to prevent users from picking it.
+//
+// Seeds all touched fields with `undefined` so a removed row clears its
+// value instead of being retained by the searchUsers thunk merge.
+export const filterToSearchParams = (
```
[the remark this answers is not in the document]
tomerlichtash: Addressed — filterToSearchParams.test.ts now covers the active→enabled translation, not-* drop, and equal-vs-fulltext/searchFields routing, plus CA parsing and edge cases. Refactored into pure per-category rules in c44cb581; suite is at 100% statements/branches/functions/lines (45 tests).

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
```diff
@@ -0,0 +1,188 @@
...
+import { filterToSearchParams } from '../../../state/filterToSearchParams';
+import { CustomAttributeTypeMap } from '../../../api/types';
+
+const ROLES_COLUMN_ID = 'roles';
+const CA_COL_PREFIX = 'customAttributes.';
+
+// Map CustomAttribute proto `type` → filter inputType + operator allowlist.
+// v0: equality only across the board (date intentionally omitted — no BE
+// op support yet). Widen when BE confirms richer ops on `customAttributes`.
+const CA_TYPE_TO_FILTER: Record<
```
tomerlichtash: why is this part of mixin?

packages/widgets/user-management-widget/src/lib/widget/state/filterToSearchParams.ts
```diff
@@ -0,0 +1,259 @@
...
+        field: LIKE_FIELD_MAP[row.column],
+        valStr: `${row.prefix ?? ''}${v}${row.suffix ?? ''}`,
+        ...(row.operator.startsWith('not-') ? { negative: true } : {}),
+      },
+    ],
+  };
+};
+
+// Any remaining negation op is unsupported on the flat path — drop the row.
+const dropUnsupportedNegation: RowRule = (row) =>
```
tomerlichtash: What does `dropUnsupportedNegation` do?

packages/widgets/user-management-widget/test/filterToSearchParams.test.ts
```diff
@@ -0,0 +1,589 @@
+import '@testing-library/jest-dom';
+import { filterToSearchParams } from '../src/lib/widget/state/filterToSearchParams';
+
+describe('filterToSearchParams', () => {
```
tomerlichtash, line 66: Revise tests, see if any overlaps, if all are essential, and if there is room for optimization.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
tomerlichtash: Extracted in 27ab6902 — the CA type map + toFilterOption + enrichment moved out of the mixin into a pure module `widget/helpers/filterColumns.ts` (`enrichCustomAttributeCols`, `applyRolesColumn`). The mixin now just wires the driver. Also dropped the CA type map entirely: console-app already bakes inputType/operators into the published column config, so the widget only attaches live options + label.

packages/widgets/user-management-widget/src/lib/widget/state/filterToSearchParams.ts
tomerlichtash: It drops any negation op that survived applySearchField — NOT-LIKE on LIKE-able columns is already handled there (searchFields `negative:true`); the flat endpoint can't express any other negation. Consuming the row prevents it falling through to the array/boolean/text rules where e.g. `status not-any-of ['enabled']` would set `statuses:['enabled']` and silently filter FOR the value. Renamed to `dropUnexpressibleNegation` with an expanded comment in 27ab6902.

packages/widgets/user-management-widget/test/filterToSearchParams.test.ts
tomerlichtash: Done in 27ab6902 — removed 3 duplicate cases (no-cols array CA, text-CA-cols passthrough, status+CA combine). filterToSearchParams down to 42 tests at 100% coverage; added filterColumns.test.ts (8 tests, 100%) for the extracted helper. 50 tests total across the two suites.

tomerlichtash:
@shuni review this pr

packages/widgets/user-management-widget/src/lib/widget/state/filterToSearchParams.ts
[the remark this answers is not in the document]
tomerlichtash: Resolved. The file moved to `helpers/filterToSearchParams.ts`, and `fieldsToClear` now deliberately excludes the flat `text` field — it only seeds `searchFields`, `customAttributes`, the array/boolean fields, and text *exactFields* (emails/loginIds/phones). `text` is never seeded to `undefined`, so applying a status/role filter no longer clobbers the search box query. The exclusion and its reason are documented on `fieldsToClear`.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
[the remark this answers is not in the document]
tomerlichtash: Resolved. Added `filterToSearchParams.test.ts` (34 tests) covering the wire-critical branches: the value-map translation (`active`→`enabled`), the `not-*` negation drop, and the exact / LIKE / full-text routing.

packages/widgets/user-management-widget/src/lib/widget/api/sdk/createUserSdk.ts
```diff
@@ -49,9 +55,12 @@ export const createUserSdk = ({
         withTestUser: false,
         customAttributes,
         statuses,
+        roleNames,
+        loginIds,
         emails,
         phones,
         text,
+        searchFields,
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 71e14908. `createUserSdk.search` now forwards the six booleans (`verifiedEmail`, `verifiedPhone`, `password`, `totp`, `webauthn`, `scim`) in the destructure, the mock call, and the POST body. Added `createUserSdk.test.ts` which asserts every SearchUsersConfig field reaches the request body (fails without the fix).

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+};
+
+// --- Request assembly ---
+
+// The request fields this filter can write, so the caller can blank them before
+// applying rows. `text` is left out on purpose: it is shared with the standalone
+// search box, and clearing it would erase what the user typed there.
+const fieldsToClear = (cols: FilterableColumn[]): Set<string> => {
+  const fields = new Set<string>(['searchFields', 'customAttributes']);
```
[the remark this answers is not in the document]
tomerlichtash: Resolved in 71e14908 (widget) + console 1d8fff1d8. The filter no longer writes the shared `text` field at all — `text` is owned solely by the standalone search box. Name-type text columns now carry a `likeField` and route to `searchFields` (their own field), which IS blanked in `fieldsToClear` on every apply, so removing the row clears it. A text column with no exact/like field is dropped rather than falling back to `text`. No stranded query.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (!this.filter?.isExists) return;
+
+        const tenantRoles = getTenantRoles(this.state);
+        const customAttrs = getCustomAttributes(this.state);
+
+        // Defer #originalCols snapshot until CAs resolve. Initial data attr
+        // may arrive without CA cols (console-app writes them asynchronously
+        // after the CA fetch). Freezing too early loses CA cols permanently.
+        if (!this.#originalCols) {
+          if (customAttrs === undefined) return;
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 71e14908. Dropped the dead `customAttrs === undefined` deferral guard — `customAttributes.data` inits to `[]`, never undefined, so it never fired. The MutationObserver already re-snapshots `#originalCols` on external `data` writes, which is the real mechanism; comment updated to say so.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
```diff
@@ -0,0 +1,62 @@
...
+  customAttrs: CustomAttr[] | undefined,
+): FilterColumn => {
+  const name = col.id.slice(CA_COL_PREFIX.length);
+  const attr = customAttrs?.find((a) => a.name === name);
+  if (!attr) return col;
+  // Combo-box options already arrive as {value,label} from the API; keep them,
+  // just backfill a missing label from the value.
+  const options =
+    Array.isArray(attr.options) && attr.options.length
+      ? attr.options.map((o) => ({ value: o.value, label: o.label || o.value }))
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 71e14908. `enrichColumn` now normalizes each CA option via a `toOption` helper that accepts either a plain string or a `{value,label}` object, so a string-typed `options` no longer produces `{value: undefined}`. (The user-profile-widget sibling type is out of scope for this PR.)

tomerlichtash:
@shuni review this pr

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,308 @@
...
+    ? appendSearchFields(params.searchFields, searchFields)
+    : params.searchFields;
+
+  const mergedCustomAttributes = customAttributes
+    ? mergeCustomAttributes(params.customAttributes, customAttributes)
+    : params.customAttributes;
+
+  return {
+    ...params,
+    ...directFields,
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 3eb1f062. Direct array fields now union instead of last-wins: `mergeDirectField` concatenates (deduped) when both sides are arrays, so `Status=Active` + `Status=Invited` sends `statuses:['enabled','invited']`; scalars still overwrite. Test added.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,308 @@
...
+
+// --- Request assembly ---
+
+// The request fields this filter can write, so the caller can blank them before
+// applying rows. The shared `text` field is never written by the filter (the
+// standalone search box owns it), so it is not cleared here either.
+const fieldsToClear = (cols: FilterableColumn[]): Set<string> => {
+  const fields = new Set<string>(['searchFields', 'customAttributes']);
+  cols.forEach((col) => {
+    const { mapping } = col;
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 3eb1f062. `fieldsToClear` reads `col?.mapping` and the `byId` map filters nullish entries, so a `null` in the cols array no longer throws (matches filterColumns.ts tolerating them). Test asserts no throw with a null entry.

pnpm-lock.yaml
```diff
@@ -1412,6 +1412,8 @@ importers:
         specifier: ^5.0.2
         version: 5.4.5
 
+  packages/sdks/react-sdk/dist/cjs: {}
```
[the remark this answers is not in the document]
tomerlichtash: Acknowledged — this is unrelated install churn (a local `dist/cjs/package.json` globbed as a workspace package; no `package.json` changed in this PR). Tracking separately to regenerate the lockfile from a clean tree; not part of the filter code changes.

packages/widgets/user-management-widget/src/lib/widget/api/types.ts
```diff
@@ -71,15 +72,58 @@ export type Role = {
...
+// flips to NOT LIKE / != any. Honored only when the BE feature flag is on.
+export type SearchField = {
+  field: string;
+  valStr?: string;
+  valArr?: string[];
+  negative?: boolean;
+  valType?: string;
+};
+
+export type FilterRow = {
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 3eb1f062. `types.ts` now re-exports `FilterRow` from `@descope/sdk-component-drivers` instead of redefining it, so the widget's row type and the driver's cannot drift.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
```diff
@@ -0,0 +1,68 @@
...
+// Fill one custom-attribute column's combo-box from the live CA schema (console
+// omits select options, since they drift after publish) and backfill its label.
+// Returns the column unchanged when its attribute is gone.
+const enrichColumn = (
+  col: FilterColumn,
+  customAttrs: CustomAttr[] | undefined,
+): FilterColumn => {
+  const name = col.id.slice(CA_COL_PREFIX.length);
+  const attr = customAttrs?.find((a) => a.name === name);
+  if (!attr) return col;
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 3eb1f062. `enrichFilterCustomAttributeColumns` now drops a custom-attribute column whose attribute no longer exists once the schema has loaded (kept while loading), matching `applyFilterRolesColumn` and the grid-column behavior. Tests added.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
```diff
@@ -0,0 +1,79 @@
...
+// the live CA schema. Non-CA columns and nullish entries pass through untouched.
+export const enrichFilterCustomAttributeColumns = (
+  cols: FilterColumn[],
+  customAttrs: CustomAttr[] | undefined,
+): FilterColumn[] =>
+  cols
+    // Drop a custom-attribute column whose attribute no longer exists, so users
+    // cannot filter on a deleted attribute (which always returns zero results).
+    // Keep everything while the schema is still loading (customAttrs undefined).
+    .filter((c) => {
```
tomerlichtash: consistency and readability - cols, when single - col; not `c`

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
```diff
@@ -0,0 +1,79 @@
...
+// column when there are none, else fill its value combo-box with the role list.
+export const applyFilterRolesColumn = (
+  cols: FilterColumn[],
+  tenantRoles: Role[] | undefined,
+): FilterColumn[] => {
+  if (!tenantRoles?.length) {
+    return cols.filter((c) => c?.id !== ROLES_COLUMN_ID);
+  }
+  const options: FilterOption[] = tenantRoles.map((r) => ({
+    value: r.name,
```
tomerlichtash: same here - roles, when single - role, not `r`

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
tomerlichtash: Done in the latest commit: renamed `c` -> `col` (and `a` -> `candidate`) across the enrich callbacks.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterColumns.ts
tomerlichtash: Done: `r` -> `role` in applyFilterRolesColumn.

packages/libs/sdk-component-drivers/src/FilterDriver.ts
```diff
@@ -0,0 +1,91 @@
...
+    return () => this.ele?.removeEventListener('filter-apply', handler);
+  }
+
+  onClear(cb: (detail: FilterEventDetail) => void) {
+    const handler = (e: Event) => cb((e as CustomEvent).detail);
+    this.ele?.addEventListener('filter-clear', handler);
+    return () => this.ele?.removeEventListener('filter-clear', handler);
+  }
+
+  onCancel(cb: (detail: FilterEventDetail) => void) {
```
nirgur: Do we need onCancel? I don't see we are using it

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,325 @@
...
+const searchField = (
+  field: string,
+  valStr: string,
+  negative: boolean,
+): Params => ({
+  searchFields: [{ field, valStr, ...(negative ? { negative: true } : {}) }],
+});
+
+const customAttr = (name: string, value: unknown): Params =>
+  ({ customAttributes: { [name]: value } }) as Params;
```
nirgur: Why the `as Params` cast here?

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,325 @@
...
+  cols: FilterableColumn[] = [],
+): Params => {
+  // The search remembers the previous params and merges the new ones on top, so
+  // a filter the user removed would otherwise stay applied. Start by setting
+  // every filterable field to undefined to clear it; the active rows below then
+  // set the ones still in use.
+  const cleared: Params = {};
+
+  fieldsToClear(cols).forEach((field) => {
+    (cleared as any)[field] = undefined;
```
nirgur: Can we avoid `as any` here?

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
```diff
@@ -0,0 +1,325 @@
...
+      return null;
+  }
+};
+
+// --- Request assembly ---
+
+// The request fields this filter can write, so the caller can blank them before
+// applying rows. The shared `text` field is never written by the filter (the
+// standalone search box owns it), so it is not cleared here either.
+const fieldsToClear = (cols: FilterableColumn[]): Set<string> => {
```
nirgur, line 226: fieldsToClear only clears the current cols' fields. If a col is dropped while its filter is applied (roles emptied / console republishes fewer cols), its field stays stale in searchParams. Edge case - should we clear removed cols too? WDYT

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+          });
+          observer.observe(filterEle, {
+            attributes: true,
+            attributeFilter: ['data'],
+          });
+        }
+
+        this.#syncColumns();
+        this.subscribe(this.#syncColumns, getTenantRoles);
+        this.subscribe(this.#syncColumns, getCustomAttributes);
```
nirgur, line 121: This runs on every state change, not only when roles/CAs change (subscribe doesn't gate on the selector). Can we memoize it with withMemCache like the other mixins?

packages/libs/sdk-component-drivers/src/FilterDriver.ts
tomerlichtash: You're right. Left over, removing.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
tomerlichtash: Narrowed to `as SearchUsersConfig['customAttributes']`.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
tomerlichtash: Removed `any`, typed as `Record<string, undefined>`.

packages/widgets/user-management-widget/src/lib/widget/helpers/filterToSearchParams.ts
tomerlichtash: Fixed. The clear list is now built from `#originalCols` (the full snapshot of the filter's columns), not the runtime-trimmed `filter.data`, so a column hidden at runtime (e.g. Roles when the tenant has none) still gets its field blanked; and when the `data` attribute changes, `onDataChange` re-syncs from it.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
tomerlichtash: Fixed. `#syncColumns` now memcached.

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
```diff
@@ -0,0 +1,121 @@
...
+          ...filterToSearchParams([], this.#columnsForRequest()),
+          page: 0,
+        });
+      };
+
+      // subscribe() fires on every state change, so memoize on (roles, CAs) plus
+      // #originalCols. onDataChange nulls #originalCols to force a re-snapshot,
+      // which busts the cache (the key changes); unrelated state changes keep the
+      // same refs and skip.
+      #updateColumns = withMemCache(
```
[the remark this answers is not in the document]
tomerlichtash: Fixed. #originalCols is re-snapshotted to a fresh frozen array on each external data change (via #captureAndSync), so it's a valid memo key that busts per change instead of collapsing to a constant [null, roles, cas].

packages/widgets/user-management-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initFilterMixin.ts
```diff
@@ -0,0 +1,121 @@
...
+          customAttrs: ReturnType<typeof getCustomAttributes>,
+          originalCols: readonly FilterColumn[] | null,
+        ) => {
+          // Snapshot the published columns once. If console-app later rewrites
+          // the `data` attribute (e.g. adds CA columns after its fetch), the
+          // driver's onDataChange handler resets #originalCols so we re-snapshot.
+          if (!originalCols) {
+            this.#originalCols = Object.freeze(this.filter.data.slice());
+          }
+          this.filter.data = this.#resolveColumns(tenantRoles, customAttrs);
```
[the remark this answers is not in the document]
tomerlichtash: Fixed. FilterDriver.set data now skips the write when the attribute already holds the value, so a redundant identical update no longer fires attributeChangedCallback and can't reset applied rows.

packages/widgets/user-management-widget/src/lib/widget/api/sdk/createUserSdk.ts
```diff
@@ -24,9 +24,18 @@ export const createUserSdk = ({
...
     emails,
     phones,
     text,
+    searchFields,
+    verifiedEmail,
+    verifiedPhone,
+    password,
+    totp,
+    webauthn,
+    scim,
```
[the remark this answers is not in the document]
tomerlichtash: Fixed. search() now spreads the config into the request body instead of enumerating a fixed field list, so any mapping.field the console publishes reaches the backend without an SDK release.

tomerlichtash:
@shuni resolve or finalize this pr review

nirgur (approved):

PR #1424, opened 2026-06-23
feat(user-profile-widget): forward client/form flow inputs into widget flows

asafshen:
## Summary

Widget flows are started internally with **empty inputs** (`input: {}`, `options.client: {}`), so any flow step configured with a dynamic `{{form.*}}` / `{{client.*}}` reference silently resolves to **empty** - no error, no warning. The motivating case: an End step's `refreshCookieName = {{form.cookieName}}` is dropped inside the widget, so the refresh cookie comes back under the default name `DSR` instead of the caller's value.

This wires caller-supplied **`client`/`form` inputs** through the widget into its internal `<descope-wc>` flow-start call - the same way they already flow for a directly-embedded `<Descope>` flow. No per-setting special-casing: it fixes every dynamic-config use case at once.

## For Humans
 - expose `client` and `form` inputs from client for User Profile Widget (currently only one widget* and only for react sdk**) 
 - refactored ` createFlowTemplate({...})` into a `buildFlowTemplate` mixin which also handled 
   - getting the client and form inputs 
   -  merge them with the things we already put inside those
   
*  if we agree on the solution - I'll propagate it to other widgets **and high level sdks
 
## Approach

- New **`flowInputMixin`** exposes `client`/`form` getters (read from the widget's attributes) and a `buildFlowTemplate({ flowId, client?, form? })` helper that injects the widget's shared flow config and forwards the inputs.
- All **21 `createFlowTemplate(...)` call sites** in `user-profile-widget` now go through `this.buildFlowTemplate(...)`, so every flow the widget runs (edit / delete / generic-flow-button / redirect) receives the inputs from one place. This also removes the boilerplate each call site duplicated (net **-86 lines**).
- Per-flow form values the widget supplies internally (e.g. passkey/device ids) merge **over** the caller's `form` on key conflict.
- React: `client`/`form` added to `WidgetProps` and forwarded as `.attr` in `UserProfile.tsx`, mirroring `<Descope>`.

```tsx
<UserProfile
  form={{ cookieName: 'DSR_wellsense' }}
  client={{ tenantId: 't1' }}
/>
```

## Scope

This is **Part 1** - `user-profile-widget` + React. The `client`/`form` types live on the shared `WidgetProps`, so **Part 2** just wires the `.attr` forwarding into the remaining flow-running widgets (`user-management`, `tenant-profile`, `outbound-applications`) and the Angular/Vue wrappers.

Part of descope/etc#16443


packages/widgets/user-profile-widget/src/lib/widget/mixins/flowInputMixin.ts
```diff
@@ -0,0 +1,81 @@
...
+      // caller-supplied form inputs (raw JSON string), forwarded as flow inputs.
+      // kept as a string - createFlowTemplate stringifies and <descope-wc>
+      // parses it back via transformFlowInputFormData, matching <Descope>.
+      get formInput(): string | undefined {
+        return this.getAttribute('form') || undefined;
+      }
+
+      // merge a per-flow form (e.g. passkey ids the widget supplies) over the
+      // caller-supplied form inputs, with the per-flow values taking precedence.
+      #resolveForm(
```
asafshen: facts:
 - the form input is Record<string, string> (what we already use)
 - the new attribute ( formInput - is a string | undefined)

this function relies on `createFlowTemplate` implementation that calls `stringifyValue` - we can either always retrun string or always return object - but I don't think that matters much

packages/libs/sdk-mixins/src/mixins/flowInputMixin/flowInputMixin.ts
```diff
@@ -0,0 +1,79 @@
...
+      // merge a per-flow form (e.g. ids the widget supplies) over the
+      // caller-supplied form inputs, with the per-flow values taking precedence
+      #resolveForm(
+        form?: Record<string, any>,
+      ): Record<string, any> | string | undefined {
+        const callerForm = this.formInput;
+        if (!form) return callerForm;
+        if (!callerForm) return form;
+        try {
+          return { ...JSON.parse(callerForm), ...form };
```
tomerlichtash, line 51: Not sure I follow something here - all JSON.stringify calls were removed to be handled in mixin, but mixin calls JSON.parse. Also, mixin now always receives an object, not a string. How does this work? Doesn't the widget DOM eventually needs a stringified value?

asafshen: I was thinking the same thing when looked at it, and commented here

https://github.com/descope/descope-js/pull/1424#discussion_r3460403482

in short - no, helper createFlowTemplate string objects anyway, this "keeps" old behavior that used to pass object values (it was done for internal use)

tomerlichtash (approved):

PR #1431, opened 2026-06-29
feat(web-component): warn on multiple same-flow components on a page RELEASE

nirgur:
## Summary
- Rendering more than one `<descope-wc>` with the same `project-id` + `flow-id` on a page is unsupported and can make the flow misbehave (e.g. the sign-in screen flickering after a successful login — the customer report in #14040).
- Adds a **dev warning** (via the SDK logger) that detects this on mount and on runtime `project-id`/`flow-id` changes, and points the developer at the fix.
- **Warning only — no behavior change.** Single instances and instances running *different* flows are unaffected.

## Linked
- Ticket: descope/etc#14040
- Related PRs: none (single-repo change)

## Changes
- New `duplicateFlowWarningMixin` (`src/lib/mixins/`): queries the DOM for `<descope-wc>` siblings sharing the same project-id + flow-id and warns **once per identity**, from the non-first instance. Hooks `connectedCallback` (kept out of the flow `init()` chain on purpose) and `attributeChangedCallback`.
- `BaseDescopeWc`: composes the mixin and forwards `super.attributeChangedCallback?.()` (mirrors the existing `super.disconnectedCallback?.()` forward) so mixins can react to attribute changes.
- Tests: pure-detection unit tests + a component-mount test covering the dedup (incl. `formMountMixin` re-parent double-fire) and the on-change path.

## Manual test plan
- Render two `<descope-wc>` with the same project-id + flow-id → exactly **one** console warning.
- Render a single component → **no** warning.
- Render two with **different** flow-ids → **no** warning.
- Change a live component's `flow-id` to collide with another → warns.

## Risk / review focus
- `BaseDescopeWc.attributeChangedCallback` now calls `super.attributeChangedCallback?.()`. Verified only this mixin defines one in the web-component chain today, so no other behavior changes; it's placed inside the post-init observed-change block so it doesn't double-fire on mount.
- Light-DOM only: a `<descope-wc>` nested in another component's shadow root is intentionally not detected (acceptable for a dev warning).

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -28,7 +28,11 @@ import {
 } from '../helpers/flowInputs';
 import { setCustomStorage } from '../helpers/storage';
 import { IsChanged } from '../helpers/state';
-import { formMountMixin, componentConditionsMixin } from '../mixins';
+import {
+  formMountMixin,
+  componentConditionsMixin,
+  duplicateFlowWarningMixin,
+} from '../mixins';
 import {
   AutoFocusOptions,
   CustomStorage,
@@ -51,6 +55,7 @@ const BaseClass = compose(
   staticResourcesMixin,
   formMountMixin,
   componentConditionsMixin,
+  duplicateFlowWarningMixin,
   injectStyleMixin,
 )(HTMLElement);
 
@@ -705,6 +710,11 @@ class BaseDescopeWc extends BaseClass {
       });
 
       this.#debugState.update({ isDebug: this.debug });
+
+      // Forward to the composed mixin chain (e.g. duplicateFlowWarningMixin) so
+      // mixins can react to project-id/flow-id changes. Only reached for real
+      // post-init changes, so it does not double-fire during initial mount.
+      super.attributeChangedCallback?.(attrName, oldValue, newValue);
     }
   }
 }
--- a/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -681,6 +681,10 @@ class BaseDescopeWc extends BaseClass {
     oldValue: string,
     newValue: string,
   ) {
+    // Forward to the composed mixin chain first so other mixins always receive
+    // the callback, regardless of the early return below.
+    super.attributeChangedCallback?.(attrName, oldValue, newValue);
+
     if (!this.shadowRoot.isConnected || !this.#init) return;
 
     if (
@@ -710,11 +714,6 @@ class BaseDescopeWc extends BaseClass {
       });
 
       this.#debugState.update({ isDebug: this.debug });
-
-      // Forward to the composed mixin chain (e.g. duplicateFlowWarningMixin) so
-      // mixins can react to project-id/flow-id changes. Only reached for real
-      // post-init changes, so it does not double-fire during initial mount.
-      super.attributeChangedCallback?.(attrName, oldValue, newValue);
     }
   }
 }
--- a/packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
+++ b/packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
@@ -1,0 +1,102 @@
+/* eslint-disable import/prefer-default-export */
+import { compose, createSingletonMixin } from '@descope/sdk-helpers';
+import { loggerMixin } from '@descope/sdk-mixins';
+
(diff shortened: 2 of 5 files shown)


packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
```diff
@@ -0,0 +1,102 @@
...
+const DESCOPE_WC_TAG = 'descope-wc';
+const PROJECT_ID_ATTR = 'project-id';
+const FLOW_ID_ATTR = 'flow-id';
+
+/**
+ * All `<descope-wc>` elements in `el`'s document that run the same project-id
+ * and flow-id as `el` (including `el` itself), in document order. Empty when
+ * `el` is missing either attribute.
+ */
+export const getSameFlowInstances = (el: Element): Element[] => {
```
asafshen, line 14: nit: can remove the export, because I think its not used outside this module 

(also isDuplicateSameFlowInstance)

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -705,6 +710,11 @@ class BaseDescopeWc extends BaseClass {
       });
 
       this.#debugState.update({ isDebug: this.debug });
+
+      // Forward to the composed mixin chain (e.g. duplicateFlowWarningMixin) so
+      // mixins can react to project-id/flow-id changes. Only reached for real
+      // post-init changes, so it does not double-fire during initial mount.
+      super.attributeChangedCallback?.(attrName, oldValue, newValue);
```
asafshen: isn't it more commmon to call super.<...> in the start of the func?
(avoining side effect / returning early)
Image

packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
```diff
@@ -0,0 +1,102 @@
...
+          if (
+            this.#warnedForKey === key ||
+            !isDuplicateSameFlowInstance(this)
+          ) {
+            return;
+          }
+          this.#warnedForKey = key;
+          this.logger.warn(
+            'Multiple Descope flow components detected on the same page',
+            `More than one <descope-wc> is running the same project ("${projectId}") and flow ("${flowId}") on this page. Running the same flow in more than one component at once is not supported and may cause the flow to behave unexpectedly. Render a single Descope flow component per page.`,
```
asafshen: the wording of `Running the same flow in more than one component at once is not supported` sounds a bit confusing, maybe `Running the same flow in more than once in the same page is not supported`?


also, I am a bit concern that a high level user (e.g. react) will not necessarily know what is descope-js, but the last sentence clarify it a a bit and  I don't have a better idea

packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
```diff
@@ -0,0 +1,102 @@
...
+      // The project|flow key this element has already warned for, so we warn at
+      // most once per identity. connectedCallback can fire multiple times for
+      // the same element (e.g. formMountMixin re-parents it into a <form>), so a
+      // plain per-call warning would fire repeatedly.
+      #warnedForKey?: string;
+
+      #scheduleDuplicateFlowCheck() {
+        // Defer so sibling instances mounted in the same tick are attached to
+        // the DOM before we query it.
+        queueMicrotask(() => {
```
asafshen, line 58: out of curiosity, if we remove the `queueMicrotask` and run it immediately, what will happen?

asafshen (dismissed):
very nice, few questions

packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
nirgur: they're exported for the unit test 
we can remove the export and test only through the mounted component but it will be a bit more complex

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
nirgur: moved super call to the start

packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
nirgur: updated the warn text

packages/sdks/web-component/src/lib/mixins/duplicateFlowWarningMixin.ts
nirgur: if we run it immediately, a second component added at the same time might be missed

asafshen (approved):
👍👍

PR #1434, opened 2026-07-10
feat(web-js-sdk): support well-known issuer URLs and resource-scoped tokens for OIDC

joshwertheim:
## Related Issues

Fixes https://github.com/descope/etc/issues/14584

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| service a PR | Link to PR |
| service b PR | Link to PR |

## Description

This branch adds the complete client-to-server story for OIDC with Inbound Apps; sign in against any Descope Inbound App by pasting its well-known URL, optionally request a resource-scoped access token, and have nextjs-sdk's server-side middleware validate that token's audience correctly (including the audience-mismatch edge case).

Manually verified end-to-end against a live project; Federated-App (applicationId) OIDC path is untouched.

## Must

- [x] Tests (new ones added)
- [x] Documentation (updated)

### DIFF
--- a/packages/sdks/react-sdk/README.md
+++ b/packages/sdks/react-sdk/README.md
@@ -914,7 +914,34 @@ To use an inbound app as an OIDC provider, you must provide both the `issuer` an
 </AuthProvider>
 ```
 
+`issuer` also accepts the full well-known URL as-is - you don't need to strip it yourself:
+
+```js
+oidcConfig={{
+  issuer:
+    'https://api.descope.com/v1/apps/<Project ID>/.well-known/openid-configuration',
+  clientId: 'your-inbound-app-client-id',
+}}
+```
+
+Federated apps (`applicationId`) and inbound apps (`issuer` + `clientId`) are distinct OIDC paths with distinct default scopes - the default scope is never inferred from the shape of the `issuer` URL, only from which fields you supplied.
+
 When using a custom `issuer` (including inbound apps), the default scope is `'openid'` instead of the full Descope scope. You can override this by providing a custom `scope` value.
+
+##### Requesting a resource-scoped token (`resource`)
+
+If your inbound app is scoped to a specific resource, pass `resource` (per [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707)) to request an access token bound to that resource at sign-in. It's applied both when redirecting to sign in and when refreshing the token, and accepts either a single resource or an array:
+
+```js
+oidcConfig={{
+  issuer: 'https://api.descope.com/v1/apps/<Project ID>',
+  clientId: 'your-inbound-app-client-id',
+  resource: 'https://api.my-app.com',
+  // or: resource: ['https://api.my-app.com', 'https://other-api.my-app.com'],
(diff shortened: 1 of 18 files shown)


packages/sdks/react-sdk/README.md
```diff
@@ -912,8 +912,35 @@ To use an inbound app as an OIDC provider, you must provide both the `issuer` an
...
+
+```js
+oidcConfig={{
+  issuer:
+    'https://api.descope.com/v1/apps/<Project ID>/.well-known/openid-configuration',
+  clientId: 'your-inbound-app-client-id',
+}}
+```
+
+Federated apps (`applicationId`) and inbound apps (`issuer` + `clientId`) are distinct OIDC paths with distinct default scopes - the default scope is never inferred from the shape of the `issuer` URL, only from which fields you supplied.
```
gaokevin1: Federated apps, also have Client IDs potentially. So I think you need to be able to override the Issuer with either a Federated App URL or Inbound App URL.

You should also be able to override the Client ID as you've done, but if you provide a Federated App URL, you can use the Project ID if no client ID is provided as the client ID.

The user shouldn't need to know this information about Federated vs Inbound, we should just tell them to take the Discovery URL of the Inbound or Federated App they want to use with this OIDC client, and put it here.

packages/sdks/react-sdk/README.md
```diff
@@ -912,8 +912,35 @@ To use an inbound app as an OIDC provider, you must provide both the `issuer` an
...
+
 When using a custom `issuer` (including inbound apps), the default scope is `'openid'` instead of the full Descope scope. You can override this by providing a custom `scope` value.
```
gaokevin1: This is based on the project level configuration, or a specific resource right?

Can you give an example of passing in that custom `scope` value as a code snippet?

For Federated Apps, scopes aren't really that relevant tbh.

packages/sdks/react-sdk/README.md
```diff
@@ -912,8 +912,35 @@ To use an inbound app as an OIDC provider, you must provide both the `issuer` an
...
 
+##### Requesting a resource-scoped token (`resource`)
```
gaokevin1: Does it make sense to support passing a resource parameter both when you do the initial config of the SDK (globally), but also when you make specific login requests? 

In most cases it'll probably be at the login request, but I guess there might be a scenario when the app only talks to when resource where you define it in the initialization of the SDK.

packages/sdks/react-sdk/README.md
```diff
@@ -912,8 +912,35 @@ To use an inbound app as an OIDC provider, you must provide both the `issuer` an
...
+
+If your inbound app is scoped to a specific resource, pass `resource` (per [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707)) to request an access token bound to that resource at sign-in. It's applied both when redirecting to sign in and when refreshing the token, and accepts either a single resource or an array:
```
gaokevin1: We should support both `audience` and `resource`, not just `resource`.

PR #1442, opened 2026-07-20
fix(web-component): anchor username for password managers on password screens

tomerlichtash:
## Problem

Screens with a password input but no username input (step-up password change, password step of multi-step login) give password managers nothing to associate the credential with. The password is saved without a username — Apple Passwords on iOS/macOS creates a duplicate entry instead of updating the existing one (descope/etc#16712).

## Solution

Inject a hidden `autocomplete="username"` input pre-filled with the user's identifier, so password managers save "username + password" as on a normal form.

1. Run on screens that contain `password` or `new-password` components only.
2. Skip if the screen already has an `autocomplete="username"` input, or no identifier resolves.
3. Resolve the identifier through a fallback chain (first non-empty wins):
   - **3a. Flow state** — `screenState.user` email/loginIds sent by the backend. Often missing: e.g. after magic link auth the state carries only the user's name.
   - **3b. Identifier submitted earlier in this flow** — captured from form data on every submit (`externalId` → `email` → `phone`, same precedence as the backend's `getLoginID`). Covers the magic-link case: the email typed on screen 1 anchors the password on screen 2. Kept in sessionStorage so it survives mid-flow reloads, and stored with the flow execution id — a value from a different execution is ignored, so it never leaks into another flow or another user in the same tab.
   - **3c. Last authenticated user** — `sdk.getLastUserLoginId()` (localStorage, written by the SDK on successful auth). Covers step-up: the user logged in earlier on this browser, so their login id is already persisted.
4. Append the hidden username `<input>` as a direct child of the `<descope-wc>` element (light DOM) — password manager save heuristics do not scan shadow DOM (verified manually on Safari; this is the load-bearing constraint).
5. Hide the input by setting styles through the CSSOM (`input.style.x = ...`). The two simpler alternatives don't work: `type="hidden"` inputs are ignored by password managers, and a `style` attribute is blocked under strict CSP `style-src`.
6. Removed automatically between screens by its own cleanup: the anchor carries a dedicated `data-username-anchor` marker, swept by `clearUsernameAnchor` on every screen render (independent of the external-input mechanism).
7. Pass the same identifier to `navigator.credentials.store(PasswordCredential)` — fixes the Chromium path, which previously stored nothing on these screens.

## Other

- Verified on macOS Safari: save prompt shows the email, existing Apple Passwords entry is updated, no duplicate.
- 11 jest tests in a dedicated `descope-wc.password-managers.test.ts` suite.
- Pending: on-device iOS verification of the original report.

Closes descope/etc#16712


packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -2008,6 +2075,16 @@ class DescopeWc extends BaseDescopeWc {
...
         const eleDescopeAttrs = getElementDescopeAttributes(submitter);
 
+        // same identifier hierarchy as #handleStoreCredentials
+        const submittedLoginId = getFirstNonEmptyValue(formData, [
+          'externalId',
+          'email',
+          'phone',
+        ]);
+        if (submittedLoginId) {
+          setLastSubmittedLoginId(submittedLoginId);
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in b9f7feac — the stored value is now `{executionId, loginId}` and the getter only returns it when it matches the current flow execution, so a later flow (or another user) in the same tab never inherits a stale identifier. Reload persistence within the same execution is preserved (resumed flows keep their execution id). Added a test covering the stale-execution case.

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -820,6 +820,90 @@ export const clearPreviousExternalInputs = () => {
...
+    hostEle.querySelector('input[autocomplete="username"]') ||
+    contentRootEle.querySelector('input[autocomplete="username"]')
+  ) {
+    logger?.debug(
+      'Username anchor: screen already has an autocomplete="username" input',
+    );
+    return;
+  }
+
+  logger?.debug(`Username anchor: injecting hidden input with value ${email}`);
```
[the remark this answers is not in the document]
tomerlichtash: Already addressed before this review landed — the `Screen user resolved` and `Store credentials` debug logs (and the anchor value log) were removed in an earlier commit on this branch; the flagged lines come from the cumulative PR diff. Current logs are reason-only ("no user identifier available" etc.) and carry no identifiers.

tomerlichtash:
@shuni review this pr

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -2008,6 +2080,16 @@ class DescopeWc extends BaseDescopeWc {
...
         const eleDescopeAttrs = getElementDescopeAttributes(submitter);
 
+        // same identifier hierarchy as #handleStoreCredentials
+        const submittedLoginId = getFirstNonEmptyValue(formData, [
+          'externalId',
+          'email',
+          'phone',
+        ]);
+        if (submittedLoginId) {
+          setLastSubmittedLoginId(submittedLoginId, this.#executionId);
```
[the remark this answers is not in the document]
tomerlichtash: Fixed in 8bec31f9 — the identifier is now captured after `next()` resolves, using the server-assigned `executionId` from the response (falling back to the current step state's id), so first-screen-optimized flows store it under the real execution id. Both sessionStorage helpers also guard against a missing execution id: it's never written and never matches, eliminating the `null === null` sharing path.

tomerlichtash:
@shuni review this pr

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -80,6 +83,47 @@ import {
...
+// (e.g. the email entered before a magic link was sent), used when neither
+// the flow state nor a previous authentication provides one.
+// kept in sessionStorage (tab scoped, cleared when the tab closes) so it
+// survives page reloads during the flow, and scoped to the flow execution so
+// a stale identifier is never carried into another flow (or another user)
+// running in the same tab
+const getLastSubmittedLoginId = (executionId: string): string => {
+  // a missing execution id must never match anything, otherwise two flows
+  // without one would share the stored identifier
+  if (!executionId) {
```
nirgur: What if the login id is on the first page, and start was not called yet?

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -820,6 +820,77 @@ export const clearPreviousExternalInputs = () => {
...
+  contentRootEle.querySelector('descope-new-password') ||
+  contentRootEle.querySelector('descope-password');
+
+const hasUsernameInput = (roots: Element[]) =>
+  roots.some((root) => root.querySelector('input[autocomplete="username"]'));
+
+const createUsernameAnchor = (email: string) => {
+  const input = document.createElement('input');
+  input.setAttribute('autocomplete', 'username');
+  input.setAttribute('name', 'username');
```
nirgur: We need to make sure this input value is not sent to the sever on submit

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -820,6 +820,77 @@ export const clearPreviousExternalInputs = () => {
...
+    .querySelectorAll(`[${USERNAME_ANCHOR_ATTRIBUTE}="true"]`)
+    .forEach((ele) => ele.remove());
+};
+
+const findPasswordComponent = (contentRootEle: Element) =>
+  contentRootEle.querySelector('descope-new-password') ||
+  contentRootEle.querySelector('descope-password');
+
+const hasUsernameInput = (roots: Element[]) =>
+  roots.some((root) => root.querySelector('input[autocomplete="username"]'));
```
nirgur: What if the flow screen already contains a username component? (e.g. descope-email-field)

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -820,6 +820,77 @@ export const clearPreviousExternalInputs = () => {
...
+// have no identifier input, so password managers can associate the credential
+// with a user (descope/etc#16712). Appended to the host light DOM — password
+// manager save heuristics (verified on WebKit) do not scan shadow trees
+export const injectUsernameAnchor = (
+  hostEle: HTMLElement,
+  contentRootEle: Element,
+  user?: { email?: string; loginIds?: string[] },
+  logger?: { debug: (...args: string[]) => void },
+) => {
+  const email = user?.email || user?.loginIds?.[0];
```
nirgur: I think we should call it loginId

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -80,6 +83,47 @@ import {
...
+    return '';
+  }
+  try {
+    const raw = window.sessionStorage?.getItem(
```
nirgur: Will it work for SSR frameworks, like NextJS?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1679,10 +1724,29 @@ class DescopeWc extends BaseDescopeWc {
...
     this.loggerWrapper.debug('Rendering a flow screen');
 
+    this.#executionId = executionId;
+
+    // flow state does not always carry the user (e.g. after magic link auth),
+    // so fall back to the last authenticated user's login id
+    const stateUser = screenState?.user as {
+      email?: string;
+      loginIds?: string[];
+    };
```
nirgur: Why do we need to cast it?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1933,12 +1998,22 @@ class DescopeWc extends BaseDescopeWc {
     );
   }
 
+  // the authenticated user of the current screen (if any), used as a fallback
+  // identifier for password manager integrations on screens that have no
+  // identifier input (e.g. step-up password change)
+  #screenUser: { email?: string; loginIds?: string[] };
+
+  #executionId: string;
```
nirgur: Why do we need it also here?
we already have it in the state

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1967,6 +2042,19 @@ class DescopeWc extends BaseDescopeWc {
     eles.forEach((ele) => this.#handleExternalInputs(ele));
   }
 
+  #updateUsernameAnchor() {
+    // clear the previous screen's anchor, so each screen gets an anchor only
+    // if it needs one
+    clearUsernameAnchor();
```
nirgur:
```suggestion
    clearUsernameAnchorIfExists();
```

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1967,6 +2042,19 @@ class DescopeWc extends BaseDescopeWc {
...
+
+    injectUsernameAnchor(
```
nirgur:
```suggestion
    injectUsernameAnchorIfNeeded(
```

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur, line 1: Can we extract this logic into a mixin so it will be in one place?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Handled — the login id is captured after `next()` resolves, taking the execution id from the response (`res.data.executionId`) when the flow state does not have one yet (flows that render their first screen before start). Covered by the "identifier submitted on a previous screen" test.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Fixed. The anchor value was never sent to the server (form data comes from the shadow root, the anchor lives in the host light DOM). It now also carries `data-exclude-field` as an explicit guard, and a test asserts the submitted form data has no `username` key.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Fixed. The skip check now also matches components collecting an identifier (`[name="externalId"/"email"/"phone"]`), which catches `descope-email-field` even though its input is hidden inside its own shadow root. Added a test.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Fixed — renamed to `loginId`.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Yes. The NextJS concern is `window` access during server-side evaluation — importing this module only defines functions and touches nothing global. `window.sessionStorage` is read only inside those functions, and they run on screen render / form submit, which only happen in the browser. Even if one were somehow called server-side, the storage access is wrapped in try/catch, so it no-ops instead of crashing.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Fixed. `ScreenState.user` was typed `unknown`, which forced the cast. It is now typed as `ScreenStateUser` and the cast is gone.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Fixed. The `#executionId` field is gone — the mixin reads it from `flowState.current`. Also reverted the `executionId` addition to `StepState`, which is no longer needed.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Fixed — renamed as suggested.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Fixed — renamed as suggested.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
tomerlichtash: Fixed. All password manager logic (username anchor, last submitted login id, screen user tracking, credential storing) now lives in `passwordManagerMixin`, composed into `BaseDescopeWc`.

nirgur (approved):

PR #1443, opened 2026-07-21
fix(web-component): refresh captcha token on abandoned OAuth popup RELEASE

nirgur:
## Summary
- Fixes the `DUPE` recaptcha error when a social-login (OAuth) popup is opened in popup mode, closed without finishing, and the user then submits again on the same screen (e.g. email/password).
- Root cause: the recaptcha token is single-use. A popup returns a *redirect* (not a screen), so the normal "reload the screen's clientScripts" step never runs, and the redirect response wipes `screenState`. The spent token stayed in context and was resent → backend `DUPE`.
- Fix: capture the screen's `clientScripts` at submit time, and on an *abandoned* popup close reload only those to mint a fresh token before the next submit. Non-captcha `sdkScripts` (forter, etc.) are not re-run.

## Linked
- Fixes: descope/etc#17057
- Related PRs: none (web-component only)

## Changes
- `DescopeWc.ts`
  - Popup-close detection now carries a per-popup `abandoned` flag on the `popupclosed` event (`receivedCode` in the popup closure).
  - `#handleComponentsLoadingState` captures `screenState.clientScripts` at click time (before the redirect wipes it).
  - `restoreComponentsState` reloads those captured scripts and gates the next submit on the fresh token (`#sdkScriptsLoading`) only when the popup was abandoned. Reload is reject-safe.
- New test `descope-wc.popupTokenRefresh.test.ts`.

## Manual test plan
- Elementor-style flow (social + email/password on one screen, OAuth in popup mode):
  1. Click the social button → popup opens.
  2. Close the popup without completing.
  3. Submit email/password on the same screen.
  - Expected: succeeds, no `DUPE`; the two `flow/next` calls send **different** `grecaptcha_riskToken` values.
- Verify a **successful** OAuth (popup completes) is unaffected — code exchange still runs, no extra delay.

## Risk / review focus
- Core flow code (`restoreComponentsState` / popup handling). Behavior change is scoped to the *abandoned* popup path; success, same-screen resubmit, and pageshow paths are unchanged.
- On abandoned popup, flow-level `sdkScripts` are intentionally **no longer** reloaded (only the screen's captcha script) — confirm no flow relies on that.
- Verified live against the real Elementor flow: two different tokens, no `DUPE`, flow advanced.

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
@@ -1883,27 +1883,17 @@ class DescopeWc extends BaseDescopeWc {
       this.loggerWrapper.debug('Restoring components state');
       this.removeEventListener('popupclosed', restoreComponentsState);
 
-      // Only a genuine web-popup-closed-without-response arms the token-refresh
-      // gate below. No-event callers (pageshow, same-screen resubmit) and
-      // non-web-popup emitters fall through to false, keeping today's behavior.
+      // Only a genuine web-popup-closed-without-response arms the refresh below;
+      // no-event callers (pageshow, same-screen) fall through to today's path.
       const abandoned = (e as CustomEvent)?.detail?.abandoned === true;
 
       if (abandoned && screenClientScripts.length) {
-        // A popup returns a *redirect*, not a screen, so the usual "reload the
-        // screen's client scripts" step (see #handleStepUpdate) never runs when
-        // the popup closes. That is why the captcha token that was already spent
-        // by the popup attempt gets reused on the next submit and the backend
-        // rejects it as a duplicate ("DUPE").
-        //
-        // Reload the screen's client scripts (captured at click time - the
-        // captcha lives there) to regenerate a fresh single-use token. Only
-        // clientScripts are reloaded, NOT sdkScripts (forter, darwinium, ...),
-        // which must not re-run just because a popup closed.
-        //
-        // Assigned to #sdkScriptsLoading before we re-enable the button, so the
-        // next submit blocks on the fresh token via `await this.#sdkScriptsLoading`.
-        // Wrapped so it can never reject - a rejected #sdkScriptsLoading would
-        // throw out of #runSdkScriptsModules and break the submit.
+        // The abandoned popup already spent the single-use captcha token, and a
+        // popup returns a redirect (not a screen) so the usual screen-scripts
+        // reload never runs - reusing that token would be rejected as "DUPE".
+        // Reload the screen's clientScripts (only these, not sdkScripts) to mint
+        // a fresh token, and gate the next submit on it via #sdkScriptsLoading
+        // (assigned before re-enabling the button; wrapped so it never rejects).
         this.#sdkScriptsLoading = (async () => {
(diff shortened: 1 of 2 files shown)


packages/sdks/web-component/test/descope-wc.popupTokenRefresh.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    expect(detail.abandoned).toBe(true);
+    jest.useRealTimers();
+  });
+
+  // When the popup is abandoned we reload the current screen's client scripts
+  // (the captcha lives there, in screenState) to mint a fresh token - NOT the
+  // sdk scripts (forter, etc.), which must not re-run just because a popup
+  // closed. The captcha is a screen-level clientScript, captured at click time
+  // because the popup redirect response wipes screenState.
+  it('reloads the screen captcha script (not sdk scripts) after an abandoned popup', async () => {
```
orius123: Don't we want to load the recaptcha als obefore the popup ?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1857,9 +1879,47 @@ class DescopeWc extends BaseDescopeWc {
...
+        //
+        // Reload the screen's client scripts (captured at click time - the
+        // captcha lives there) to regenerate a fresh single-use token. Only
+        // clientScripts are reloaded, NOT sdkScripts (forter, darwinium, ...),
+        // which must not re-run just because a popup closed.
+        //
+        // Assigned to #sdkScriptsLoading before we re-enable the button, so the
+        // next submit blocks on the fresh token via `await this.#sdkScriptsLoading`.
+        // Wrapped so it can never reject - a rejected #sdkScriptsLoading would
+        // throw out of #runSdkScriptsModules and break the submit.
```
orius123: comments are pretty verbose :)

orius123 (dismissed):
approved, with 2 minor suggestions

packages/sdks/web-component/test/descope-wc.popupTokenRefresh.test.ts
nirgur: We are already reloading the scripts after each screen render
Do you think we should reload it once more?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Trimmed the comment block down to the essentials.

packages/sdks/web-component/test/descope-wc.popupTokenRefresh.test.ts
orius123: But do we load it before? 
I want to mimic the issue in the test

nirgur: Good call - updated the test to mimic the issue end-to-end: it now mints a token before the popup, asserts the OAuth `next` carries it, then after an abandoned popup asserts the following submit carries a *different* (fresh) token. Pushed in 6a365612.

orius123 (approved):

PR #1444, opened 2026-07-21
feat(web-component): expose server screen data under context.data in onScreenUpdate

nirgur:
## What

Reshape the custom-screen (BYOS) `context` passed to `onScreenUpdate`. All
server-produced screen data now lives under a single, documented **`context.data`**
field, instead of being spread across the top level and a deeply-nested,
selector-keyed `componentsConfig` that the SDK used to hide entirely.

Motivation (issue [descope/etc#16940](https://github.com/descope/etc/issues/16940)): a
customer needed MFA `recoveryCodes`, which were only reachable through internal,
nested screen-state paths. Rather than expose one more internal path, we give
custom screens one clean, stable place to read screen data.

## How `context.data` works

- **Allow-list, not passthrough.** Only known, documented fields are exposed —
  renamed and normalized into a stable shape (see `CONFIG_FIELD_MAP` +
  `CustomScreenData`). Anything else in the raw screen state is intentionally
  dropped, so internal backend fields never leak.
- **Stability rule.** Every key on `context.data` is an SDK-owned, documented name.
  We never place arbitrary, flow-defined keys directly on `context.data`, so adding
  new named fields later is additive and non-breaking.
- **Dynamic-select options** (arbitrary flow-defined keys) are **not exposed** and
  are reserved for a future `context.data.selects` map. They were never available to
  custom screens before, so this is not a regression.

Top-level status/identity fields (`error`, `action`, `form`, `user`, `project`,
`lastAuth`) stay where they were.

Example:

```js
function onScreenUpdate(screenName, context, next, ref) {
  if (screenName === 'Display recovery codes') {
    const codes = context.data?.recoveryCodes; // string[]
    return true;
  }
  return false;
}
```

## Breaking change

Screen-data fields that used to sit directly on the `onScreenUpdate` context now
live under `context.data`:

| Before | After |
| --- | --- |
| `context.totp` / `context.notp` | `context.data.totp` / `context.data.notp` |
| `context.sentTo` | `context.data.sentTo` |
| `context.sso` / `context.selfProvisionDomains` | `context.data.sso` / `context.data.selfProvisionDomains` |
| `context.inboundAppApproveScopes` | `context.data.inboundAppApproveScopes` |

- Conventional-commit `feat!:` → triggers a **major** web-component release.
- **react-sdk / vue-sdk**: type the context loosely (`Record<string, any>`), so no
  TypeScript break.
- ⚠️ **Native mobile SDKs**: the whole `context` is serialized to native over the
  `beforeScreen` bridge (`DescopeWc.ts`). Any native SDK reading the relocated
  fields from the top level must migrate to `context.data`. **Needs mobile-team
  sign-off before release.**

## Test plan

- New unit suite `test/helpers/customScreenData.test.ts` (recovery codes, quote
  styles, empty/missing guards, outbound app, dropped selectors, rename/normalize,
  allow-list drop, dynamic-select drop, no componentsConfig leak, backend-`data`
  ignored, top-level relocation, no-config).
- Updated `descope-wc.customScreen.test.ts` to assert `inboundAppApproveScopes`
  under `context.data`.
- Typecheck clean; `customScreen` suite green (27 tests).
- Manual: run a real recovery-codes flow through the web component; smoke-test the
  native `beforeScreen` payload.

## Related issue

descope/etc#16940


nirgur:
@shilgapira @itaihanski Does this change affect the mobile in any way?

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -853,9 +971,11 @@ export const transformStepStateForCustomScreen = (
     sanitizedState.action = state.action;
   }
 
-  if (state.screenState?.componentsConfig?.thirdPartyAppApproveScopes?.data) {
-    sanitizedState.inboundAppApproveScopes =
```
asafshen, line 857: isn't moving this going to potentially break BC? 

if so, maybe we can just omit it from TS definition, but keep the actual value

packages/sdks/web-component/src/lib/constants/customScreens.ts
```diff
@@ -1,8 +1,19 @@
...
+// transformStepStateForCustomScreen, so a backend-provided value never leaks). The rest
+// are server-produced screen data relocated under `data`.
 export const EXCLUDED_STATE_KEYS = [
   'cssVars',
   'componentsConfig',
   'inputs',
   'errorText',
   'errorType',
   'clientScripts',
+  'data',
```
asafshen, line 13: I'm not 100% sure how this EXCLUDED_STATE_KEYS is used, but
1. why data is a part of it?
2. removing all extra fields like totp, notp, .... `selfProvisionDomains` - can potentially break BC?

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -833,6 +835,122 @@ const omitBy = <T extends Record<string, any>>(
...
+    // allow-list: only known fields surface; everything else is dropped.
+    // `newPassword` and `componentsDynamicAttrs` are handled separately below and are
+    // intentionally absent from CONFIG_FIELD_MAP. Dropped here too: dynamic-select
+    // options, which the backend spreads into componentsConfig under the flow's own
+    // field names (arbitrary keys). We don't put arbitrary keys on `data`; exposing
+    // them is reserved for a future `data.selects` map.
+    Object.entries(componentsConfig).forEach(([key, value]) => {
+      if (value == null) return;
+
+      const mapping = CONFIG_FIELD_MAP[key];
```
asafshen: if I read this properly, there are few ways we set data here

1. top level set (aka `set('totp', screenState.totp);`)
2. the mapping configuration 
3. newPassword / componentsDynamicAttrs

noting that this is 
(a) what confusing about this is that there are different layers of abstractions here, the `mapping` configration is relatively higher level of abstractions while the other are more explcicit, maybe we can move 1. into the mapping as well?
(b) more importantly, when we add new fields to return value (either componentConfig or other backend fields, - will they appear in the response? or by default - they won't, and we have to add them here?

packages/sdks/web-component/README.md
```diff
@@ -226,6 +226,61 @@ const descopeWcEle = document.querySelector('descope-wc');
...
+Every key on `context.data` is an SDK-owned, documented name. The SDK never puts
+arbitrary, flow-defined keys directly on `context.data`, so new fields are always added
+without breaking existing ones.
+
+**Not exposed yet:** dynamic-select options (a select component whose options are computed
+by the flow). Their keys are the flow's own field names, so they don't fit the fixed shape
+above; they are omitted for now and reserved for a future `context.data.selects` map. This
+data was not available to custom screens before, so nothing changes for existing code.
+
+> **Breaking change (v-next):** screen-data fields that used to sit directly on `context`
```
asafshen, line 278: I see about breaking backward now, I wonder if this is the place to communicate it? I think not

asafshen (commented):
looks greeat, few comments, mayb we should talk f2f also

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: This is going to break
I discussed this with Aviad, and he said he prefers to organize the data now and break if needed, rather than splitting the data across different locations and having it break later when we decide to reorganize it

packages/sdks/web-component/README.md
nirgur: We will also bump a major version in our SDKs
What else do you think we should do?

packages/sdks/web-component/src/lib/constants/customScreens.ts
nirgur: `data` is stripped so a in case the backend will send this attr, it won't leak
The other field removals are intended break

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: Good call — unified this. There is now a single `DATA_FIELDS` table where every field is defined the same way: a `data` key + a `from` reader over the screen state. `buildCustomScreenData` just loops the table and applies each reader (skipping undefined). The three ad-hoc paths (top-level sets, the rename/unwrap map, and the newPassword/componentsDynamicAttrs blocks) are gone — the per-field specifics (unwrap, password-policy normalize, dynamic-attr lookup) now live in small named readers, so everything sits at the same level. Behavior unchanged (typecheck + 27 custom-screen tests green).

packages/sdks/web-component/README.md
asafshen: creating major will create friction, and I'm not sure this worth it at this stage

i would consult with CS team after figuring out that usage, or just keep BC wihtout TS definition

asafshen (dismissed):
great work

packages/sdks/web-component/README.md
nirgur: We will bump a major version

asafshen (approved):

PR #1457, opened 2026-08-09
fix(web-component): catch rejected credentials.store promise

### LINKED ISSUE #1456: Asynchronous storing of credential throwing exception
In the referenced snippet of code, the `navigator?.credentials?.store?.(cred);` line can return a rejected Promise that the catch will not handle.

This has led to issues in our Playwright testing that uses headless Chromium.  With that browser we end up with an uncaught exception:

* name: `NotSupportedError`
* message: `The user agent does not support public key credentials.`

https://github.com/descope/descope-js/blob/6f1956fa6ba191c7a069d4292bad8abb39da45a5/packages/sdks/web-component/src/lib/mixins/passwordManagerMixin.ts#L218-L229

I think something like:

```js
void navigator.credentials.store(cred).catch((e) => {
  this.logger.error('Could not store credentials', e?.message ?? e);
});
```
or  making the outer function `async` and doing an `await` would resolve the issue

tomerlichtash:
Fixes #1456

## Problem

`navigator?.credentials?.store?.(cred)` returns a promise. The `try/catch` wrapped around it is synchronous, so it catches only what `new PasswordCredential(...)` throws immediately, never a rejection from `store()`.

Chromium rejects that promise with `NotSupportedError: The user agent does not support public key credentials.` when public key credentials are unavailable. For consumers this escapes as an uncaught exception and fails their headless Playwright runs.

## Fix

Attach a `.catch` in `passwordManagerMixin.ts` that routes the error to the existing logger, as suggested in the issue. Optional chaining on the call result keeps it safe when `store` is absent or a mock returns `undefined`. The synchronous `try/catch` stays for the `PasswordCredential` constructor.

## Why existing tests missed it

The existing tests mocked `navigator.credentials` as `{ store: jest.fn() }`. A bare `jest.fn()` returns `undefined` rather than a promise, so the rejection path did not exist under test.

The new case mocks `store` with `mockRejectedValue` and asserts the error reaches the logger. Before the fix it does not just fail an assertion, it kills the jest process with the reported error string. After, all 13 tests in `descope-wc.password-managers.test.ts` pass.

Note: no `RELEASE` keyword in the commit. Add it on merge if this should publish immediately.


asafshen (approved):
well done!

PR #1458, opened 2026-08-09
fix(web-component): handle rejected credentials.store promise

### LINKED ISSUE #1456: Asynchronous storing of credential throwing exception

nirgur:
## Summary
- `navigator.credentials.store()` returns a promise that can reject asynchronously. In headless Chromium it rejects with `NotSupportedError` ("The user agent does not support public key credentials").
- The surrounding sync `try/catch` never saw that rejection, so it surfaced as an uncaught error and broke consumers' Playwright tests (reported in #1456).
- Attach a `.catch` that logs via the existing logger, keeping the call fire-and-forget and preserving the optional-chaining guard (so it stays safe when `store` is unavailable).

## Linked
- Fixes #1456

## Changes
- `passwordManagerMixin.ts`: `void navigator?.credentials?.store?.(cred)?.catch(...)` logs the async failure through `this.logger.error`, matching the existing sync-branch message.
- Test: new case mocks `store` to reject and asserts the failure is caught and logged via the `console.error` spy.

## Manual test plan
- In a headless Chromium context (e.g. Playwright), run a password flow that triggers credential storage and confirm no uncaught `NotSupportedError` is thrown; the failure is logged instead.
- Existing password-manager tests still pass (store called with the right payload when supported).

## Risk / review focus
- Low risk; only adds rejection handling, no signature change, stays fire-and-forget.
- Worth a glance: the async catch uses `err?.message ?? err` (defensive, async rejection value may not be an `Error`) while the pre-existing sync catch uses `e.message` - intentional, sync branch is left untouched.

### DIFF
--- a/packages/sdks/web-component/src/lib/mixins/passwordManagerMixin.ts
+++ b/packages/sdks/web-component/src/lib/mixins/passwordManagerMixin.ts
@@ -225,7 +225,7 @@ export const passwordManagerMixin = createSingletonMixin(
             // store() may be missing (optional chain -> undefined) or reject
             // async (e.g. NotSupportedError in headless Chromium). catch the
             // rejection so it doesn't surface as an unhandled rejection.
-            void navigator?.credentials?.store?.(cred)?.catch((err) => {
+            navigator?.credentials?.store?.(cred)?.catch((err) => {
               this.logger.error(
                 'Could not store credentials',
                 err?.message ?? err,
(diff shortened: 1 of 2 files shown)


nirgur:
Closing as a duplicate of #1457, which already merged the same fix (catch the rejected `navigator.credentials.store()` promise) plus a regression test. Thanks @tomerlichtash.

PR #1461, opened 2026-08-12
feat(telemetry): add opt-in AWS CloudWatch RUM debug telemetry for web-component SDK RELEASE

nirgur:
## Summary

Opt-in AWS CloudWatch RUM debug telemetry for the web-component SDK. When the backend turns telemetry on (per-project, time-boxed, gated by company support-consent), the SDK lazy-loads the `@descope/debug-logs` UMD and streams console / network / navigation / scoped-DOM events to AWS CloudWatch RUM.

> Supersedes #1446. Same work, moved onto `feature/log-collector` to match the backend branch/PR (cross-repo convention), and with the config-path fix below.

## Related

- Backend PR: descope/backend#2045
- Ticket: descope/etc#13172

## Fix included on top of the original branch

`telemetryMixin` read `this.config.telemetry`, but the current `configMixin` nests the project's `config.json` under `this.config.projectConfig`. So `this.config.telemetry` was always `undefined` and telemetry silently never activated (it fell back to the disabled env default). Now reads `projectConfig?.telemetry ?? telemetry` — works with the current shape and older top-level shapes. Without this, telemetry never turns on once merged with current `descope-js`.

## Changes

- `@descope/debug-logs` — `TelemetryManager` (AWS RUM client + per-step `safe()` shutdown), console / network / navigation / DOM-mutation plugins.
- `@descope/sdk-mixins` (`telemetryMixin`) — `#getTelemetryConfig`: env-fallback merged with the BE `telemetry` block; empty-creds guard; `#resolveCapture` for shadow-root injection; the `projectConfig` read fix.
- `web-component` — dev rollup substitutes `DESCOPE_TELEMETRY_*`; prod `stubEnvInProd` plugin blocks any env-var leak / runtime `ReferenceError`.
- Tests + docs (README env-var docs, manual playground `.env` workflow).

## Manual test plan (verified locally, multi-region env)

- [x] Enable telemetry via BE config on a project -> SDK loads `@descope/debug-logs`, creates the AwsRum client (session id), reads `expiration`, and POSTs to `cognito-identity.eu-west-1.amazonaws.com` (send attempts; 400 only because the e2e config uses placeholder creds).
- [x] Disable telemetry -> no `@descope/debug-logs` request, no Cognito/RUM calls, console logs "Telemetry is disabled".
- [x] `expiration` propagates from `config.json` into the SDK ("expire in ~N hours" matches the support-set duration).
- [ ] With real RUM app + identity pool: confirm the session lands in CloudWatch RUM with `projectId` / `flowId` / SDK version.
- [ ] Prod build of `web-component`: grep the bundle for `process.env.` -> zero remaining refs.

## Risk / review focus

- Merge semantics of `#getTelemetryConfig` (BE block over env fallback, per-key `rumConfig` merge) — confirm it matches the BE contract shipped in descope/backend#2045.
- The `projectConfig` read fix — confirm it holds for every SDK entry point that builds `this.config`.
- `stubEnvInProd` catch-all in `web-component/rollup.config.mjs`.
- `pnpm-lock.yaml` churn — material change is the `dotenv` devDep in `@descope/debug-logs`; the rest is pnpm transitive re-resolution.


packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,338 @@
...
+        // config.json. Otherwise, fall back to values from
+        // packages/sdks/web-component/.env (DESCOPE_TELEMETRY_* — substituted
+        // at WC build time via rollup.config.app.mjs). We MERGE rather than
+        // replace so the backend can override individual fields (e.g. just
+        // expiration) without re-stating every credential.
+        const resolvedConfig = (await this.config) as any;
+        // config.json content is nested under `projectConfig` in the current
+        // configMixin; fall back to the top level for older shapes.
+        const beTelemetry =
+          resolvedConfig?.projectConfig?.telemetry ?? resolvedConfig?.telemetry;
```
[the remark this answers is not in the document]
nirgur: Done in c55b6c72. Moved the telemetry type onto ProjectConfiguration so projectConfig.telemetry is typed, and dropped the as any cast (await this.config is now Config). Added a test that feeds the nested projectConfig.telemetry shape.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,338 @@
...
+              capture: config.capture,
+            },
+            {
+              projectId: context.projectId,
+              flowId: context.flowId,
+              version: context.version || '1.0.0',
+            },
+            this.logger,
+          );
+          this.#telemetryInitialized = true;
```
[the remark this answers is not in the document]
nirgur: Done in c55b6c72. We now check manager.isReady() before flagging success. If the RUM client failed to init we log an error, clear the manager, and leave the init flag off so a later attempt can retry. Added a test for the failure + retry path.

packages/libs/debug-logs/src/plugins/networkPlugin/helpers.ts
```diff
@@ -0,0 +1,201 @@
...
+): HeadersMap {
+  const result: HeadersMap = {};
+  if (!headers) {
+    return result;
+  }
+
+  const applyPair = (key: string, value: string) => {
+    const normalizedKey = key.toLowerCase();
+    const normalizedValue = truncate ? truncate(value) : value;
+    result[normalizedKey] = normalizedValue;
```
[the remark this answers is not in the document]
nirgur: Intentional, and accepted. This is opt-in debug telemetry: it only runs when the customer enables support consent, it is time-boxed (auto-expires), and it streams to a Descope-owned CloudWatch RUM, not a third party. Capturing full headers (including Authorization) is the point - support needs the real request to debug. We are deliberately not adding a header denylist here.

packages/sdks/web-component/src/app/index.html
```diff
@@ -5,14 +5,14 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <!-- Uncomment next section if you want to test CSP -->
 
-    <meta
+    <!-- <meta
```
tomerlichtash: Finally :)

But is this intentional?

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+      #telemetryManager: any = null;
+      #telemetryInitialized = false;
+      #expirationTimeoutId: NodeJS.Timeout | null = null;
+
+      async #initializeTelemetry(
+        config: TelemetryConfig,
+        context: {
+          projectId: string;
+          flowId?: string;
+          version?: string;
```
tomerlichtash: version of what?

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+          this.logger.debug('Telemetry is disabled, skipping initialization');
+          return;
+        }
+
+        if (this.#telemetryInitialized) {
+          this.logger.debug('Telemetry already initialized');
+          return;
+        }
+
+        if (config.expiration && Date.now() >= config.expiration) {
```
tomerlichtash, line 68: How is `expiration` handling/verifying the configuration from the backend?
Is this user facing at any point? Would this work as UTC time?

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+          },
+          capture: this.#resolveCapture(beTelemetry?.capture),
+        };
+
+        // If enabled but a required RUM credential is missing (empty from both
+        // the BE block and the .env fallback), disable rather than hand AwsRum
+        // an empty applicationId/identityPoolId/region — that would either
+        // throw at construction or silently no-op with no session recorded.
+        if (
+          cfg.enabled &&
```
tomerlichtash: Can be a helper fn to validate rum config.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+            'Telemetry enabled but rumConfig is incomplete (applicationId/identityPoolId/region required). Skipping initialization.',
+          );
+          cfg.enabled = false;
+        }
+
+        // Safety cap: when no backend telemetry config is present at all,
+        // end the session after 5 minutes so dev sessions don't leak. Once
+        // the backend owns the config it controls expiration explicitly.
+        if (!beTelemetry) {
+          cfg.expiration = Date.now() + 5 * 60 * 1000;
```
tomerlichtash, line 266: Worth adding a logger msg here, IMO

packages/sdks/web-component/src/app/index.html
nirgur: Good catch, that was a leftover from local testing. Reverted.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
nirgur: It is the SDK version. It gets passed to AwsRum as the application version, so every recorded session is tagged with which SDK version the end-user ran. Renamed the field to `sdkVersion` to make it clear.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
nirgur: expiration is a unix timestamp in ms that the backend sends in config.json. We compare it to Date.now(), which is also unix ms in UTC, so the check is UTC-safe and needs no timezone handling. It is not user facing - it only stops the telemetry session when the time passes. If it is already past when we load, we skip init; if it is in the future, we set a timer to shut down at that time.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
nirgur: Done, moved it into a #isRumConfigComplete helper.

packages/libs/sdk-mixins/src/mixins/telemetryMixin/telemetryMixin.ts
nirgur: Added a debug log there.

tomerlichtash (approved):

PR #1463, opened 2026-08-17
feat(widgets): componentsConditionsMixin to apply server condition verdicts RELEASE

nirgur:
## What
Shared `componentsConditionsMixin` that fetches the server-computed condition verdict for the current user and applies hide / disable / read-only to the widget DOM (by `data-id`). Rolled out to all 8 widgets.

## How
- `componentsConditionsMixin` (in `sdk-mixins`) is composed **last** in each widget's `initMixin` so its `init()` wraps the widget render: it fires the fetch during init (overlapping the page/data fetch) and applies the verdict in `onWidgetRootReady` before `ready`.
- Reuses the widget's existing web-js-sdk `httpClient` (`this.api.httpClient`, narrowed to a minimal `ConditionsHttpClient` type so no web-js-sdk type leaks into a widget's public sdk surface). Each widget exposes that one line.
- **Loud on misconfig**: throws in `init()` if `httpClient` is missing (a wiring mistake). **Fail-open** on runtime fetch errors (render everything). **Skips** the fetch in `mock` mode.
- Applier hides via the `hidden` attribute + inline `display:none !important` (widgets don't inject a `.hidden` class into shadow DOM).

## Testing
- Mixin lifecycle: throw-on-missing, mock-skip, happy-path apply, fail-open.
- Applier: hide / disable / read-only / clear / empty.
- E2e (Playwright, firefox/chromium/webkit): hide, disable, and fail-open verified in real browsers through shadow DOM.

## Notes
- No dependency / `pnpm-lock.yaml` changes.
- `reevaluateComponentsState` is a public hook for the upcoming re-eval-on-mutation work - intentionally landed ahead of its callers (documented in code).

## Related PR
- backend: https://github.com/descope/backend/pull/2244
- console-app: https://github.com/descope/console-app/pull/5695

## Related issue
descope/etc#17289

### DIFF
--- a/packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
+++ b/packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
@@ -1,0 +1,79 @@
+// Applies server-computed component conditions (hide / disable / read-only) to a
+// widget's DOM. Mirrors the flow web-component applier, but targets widget
+// components by their `data-id` (widgets address components that way) and hides
+// via inline style + the `hidden` attribute rather than a `.hidden` class, which
+// widgets do not inject into their shadow DOM.
+
+export const COMPONENT_ACTIONS = ['hide', 'disable', 'read-only'] as const;
+export type ComponentAction = (typeof COMPONENT_ACTIONS)[number];
+
+const ATTR_SELECTOR_ESCAPE_PATTERN = /(["\\])/g;
+
+// CSS.escape with a narrow fallback for environments that lack it.
+function escapeSelector(value: string): string {
+  if (typeof CSS !== 'undefined' && typeof CSS.escape === 'function') {
+    return CSS.escape(value);
+  }
+  return value.replace(ATTR_SELECTOR_ESCAPE_PATTERN, '\\$1');
+}
+
+// IDs can repeat, so match all.
+function findComponents(root: ParentNode, id: string): Element[] {
+  return Array.from(root.querySelectorAll(`[data-id="${escapeSelector(id)}"]`));
+}
+
+function applyAction(el: Element, action: string): void {
+  switch (action) {
+    case 'hide':
+      el.setAttribute('hidden', '');
+      (el as HTMLElement).style?.setProperty('display', 'none', 'important');
+      break;
+    case 'disable':
+      el.setAttribute('disabled', 'true');
+      break;
+    case 'read-only':
+      el.setAttribute('readonly', 'true');
+      break;
+    default:
+    // Unknown action - ignore. A bad action shouldn't make it past the server.
+  }
+}
+
+function clearAction(el: Element, action: string): void {
+  switch (action) {
+    case 'hide':
+      el.removeAttribute('hidden');
+      (el as HTMLElement).style?.removeProperty('display');
+      break;
+    case 'disable':
+      el.removeAttribute('disabled');
+      break;
+    case 'read-only':
+      el.removeAttribute('readonly');
+      break;
+    default:
+    // no-op
+  }
+}
+
+/** Applies the componentId -> action map to the matching elements under root. */
+export function applyComponentsState(
+  root: ParentNode | null | undefined,
+  componentsState: Record<string, string> | undefined,
+): void {
+  if (!root || !componentsState) return;
+  Object.entries(componentsState).forEach(([id, action]) => {
+    findComponents(root, id).forEach((el) => applyAction(el, action));
+  });
+}
+
+/** Reverts a previously-applied componentId -> action map. */
+export function clearComponentsState(
+  root: ParentNode | null | undefined,
+  componentsState: Record<string, string> | undefined,
+): void {
+  if (!root || !componentsState) return;
+  Object.entries(componentsState).forEach(([id, action]) => {
+    findComponents(root, id).forEach((el) => clearAction(el, action));
+  });
+}
--- a/packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
+++ b/packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
@@ -4,9 +4,6 @@
 // via inline style + the `hidden` attribute rather than a `.hidden` class, which
 // widgets do not inject into their shadow DOM.
 
-export const COMPONENT_ACTIONS = ['hide', 'disable', 'read-only'] as const;
-export type ComponentAction = (typeof COMPONENT_ACTIONS)[number];
-
 const ATTR_SELECTOR_ESCAPE_PATTERN = /(["\\])/g;
 
 // CSS.escape with a narrow fallback for environments that lack it.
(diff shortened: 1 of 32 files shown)


packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/componentsConditionsMixin.ts
```diff
@@ -0,0 +1,115 @@
...
+          );
+        }
+        return httpClient;
+      }
+
+      async #fetchComponentsState(
+        httpClient: ConditionsHttpClient,
+      ): Promise<ComponentsState> {
+        try {
+          const res = await httpClient.get(COMPONENTS_STATE_PATH);
```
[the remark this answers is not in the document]
nirgur: Investigated: the widget condition allowlist is intentionally scoped to user-record keys only (enrollment/attributes, and `user.project.roles` -> `currUser.RoleNames`, tenancy from the user's own tenant list), all of which resolve by userId without a tenant. There are no tenant-scoped checks (e.g. "has role R in tenant X") in the allowlist, so no wrong-tenant risk. Threading `tenant` is deferred to the roles/permissions-parity follow-up that will add tenant-scoped keys. Leaving as-is for this PR.

nirgur:
@shuni please take another look - addressed all findings (see latest commits).

pnpm-lock.yaml
```diff
@@ -26,16 +26,16 @@ importers:
         version: 19.1.0(nx@19.8.14(@swc/core@1.7.1(@swc/helpers@0.5.15)))
       '@nrwl/eslint-plugin-nx':
         specifier: 19.8.4
-        version: 19.8.4(@babel/traverse@7.29.7)(@swc/core@1.7.1(@swc/helpers@0.5.15))(@types/node@20.17.13)(@typescript-eslint/parser@7.2.0(eslint@9.18.0(jiti@1.21.0))(typescript@5.7.3))(eslint-config-prettier@9.1.0(eslint@9.18.0(jiti@1.21.0)))(eslint@9.18.0(jiti@1.21.0))(nx@19.8.14(@swc/core@1.7.1(@swc/helpers@0.5.15)))(typescript@5.7.3)
+        version: 19.8.4(@babel/traverse@7.29.8)(@swc/core@1.7.1(@swc/helpers@0.5.15))(@types/node@20.17.13)(@typescript-eslint/parser@7.2.0(eslint@9.18.0(jiti@1.21.0))(typescript@5.7.3))(eslint-config-prettier@9.1.0(eslint@9.18.0(jiti@1.21.0)))(eslint@9.18.0(jiti@1.21.0))(nx@19.8.14(@swc/core@1.7.1(@swc/helpers@0.5.15)))(typescript@5.7.3)
```
tomerlichtash: PR description says "No dependency / pnpm-lock.yaml changes". Just FYI.

packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/componentsConditionsMixin.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.#appliedState = (await this.#statePromise) ?? {};
+        applyComponentsState(this.contentRootElement, this.#appliedState);
+      }
+
+      /**
+       * Re-fetch and re-apply after the widget mutates data the conditions read
+       * (e.g. a profile save). Public hook for the upcoming re-evaluate-on-mutation
+       * work - no widget wires it to its mutating flows yet.
+       */
+      async reevaluateComponentsState() {
```
tomerlichtash, line 107: Not wired, which is fine, but not tested as well.
Why don't we wire this?
Are you sure you want to commit this at this point?

packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
```diff
@@ -0,0 +1,76 @@
...
+    default:
+    // Unknown action - ignore. A bad action shouldn't make it past the server.
+  }
+}
+
+function clearAction(el: Element, action: string): void {
+  switch (action) {
+    case 'hide':
+      el.removeAttribute('hidden');
+      (el as HTMLElement).style?.removeProperty('display');
```
tomerlichtash, line 43: Wouldn't that break components' host display (usually `inline-block`)?

packages/widgets/user-profile-widget/e2e/user-profile-widget.spec.ts
```diff
@@ -333,6 +339,53 @@ test.describe('widget', () => {
     }
   });
 
+  test.describe('components conditions', () => {
```
tomerlichtash, line 342: Is this an acceptance test?
Is there sense in adding it in all widgets?
Not a must, but it seems arbitrary to add it just here.

packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/applier.ts
nirgur: This removes only the inline `display: none` we set on hide - the element reverts to its stylesheet display (e.g. `inline-block`), we do not force `display: block`. So show/hide restores the natural host display. Do you see a case where that would be problematic?

packages/widgets/user-profile-widget/e2e/user-profile-widget.spec.ts
nirgur: It is shared code - the mixin and the applier are identical across all widgets - so to avoid running the same behavior tests many times, we assert the apply behavior once in a real browser, here on user-profile. Do you find that problematic, and do you think we should add the same tests to all widgets?

packages/libs/sdk-mixins/src/mixins/componentsConditionsMixin/componentsConditionsMixin.ts
nirgur: Good point. We decided to keep the method but not wire it in this PR, and I added unit tests for it so it is not untested: re-fetch + clear the previous verdict, clear when the new state is empty, fail open on a re-fetch error, and throw on a missing httpClient.

We are not wiring it now because the trigger is widget-specific: re-eval only matters when a mutation inside the widget changes the current user's own identity (passkey/TOTP enrollment, attributes), which conditions read. That happens only in user-profile-widget. The other widgets manage other resources (keys, roles, other users, apps) and do not change the current user's identity, so the init-time verdict holds - only 4 of 8 widgets even call getMe. So reevaluateComponentsState stays in the shared mixin as a ready hook, and the user-profile wiring comes with that future work.

packages/widgets/user-profile-widget/e2e/user-profile-widget.spec.ts
tomerlichtash: I just think it's arbitrary, since all widgets have to compose it last, and only one widget validates its functional. If you think that would suffice for future bug hunts, then no problem.

tomerlichtash (dismissed):

pnpm-lock.yaml
nirgur: Fixed in f441773f: reset pnpm-lock.yaml to main. The feature adds no dependencies and changes no package.json, so the diff was pure pnpm re-resolution noise from a merge-conflict resolution - not a real dependency change. The lock now matches main and the PR description is accurate.

tomerlichtash (approved):

packages/widgets/user-profile-widget/e2e/user-profile-widget.spec.ts
nirgur: Agreed it is a judgment call. The shared behavior is covered by the mixin + applier unit tests, and each widget has the route stub so its compose-last wiring is exercised; the one functional browser test on user-profile proves the end-to-end apply once. I think that is enough for future bug hunts, so keeping it here. Thanks for the review.

PR #1474, opened 2026-08-31
feat: support saving and retrieving Descope Known Device (DKD) Token from local storage

OfekAvergil:
## Related Issues
Required for:
- https://github.com/descope/etc/issues/17609

## Related PRs
Depends on:
- https://github.com/descope/backend/pull/2386

## Description

Mirrors #1290 (DTD/Trusted Device Token local-storage support), adding the same treatment for the new Known Device Token (DKD, `knownDeviceJwt`) introduced by the backend PR above.

- `core-js-sdk`: `knownDeviceJwt?: string` added to `JWTResponse`, next to `trustedDeviceJwt`.
- `web-js-sdk`: new `KNOWN_DEVICE_TOKEN_KEY = 'DKD'` constant; `persistTokens` now also persists `knownDeviceJwt` to localStorage when present in the response body (unconditionally, same as DTD); new `getKnownDeviceToken`; `clearTokens` does not remove it (same rationale as DTD - it should outlive logout).

**Trusted implies known - never send both.** The backend now derives known-device state from a valid trusted-device token whenever no DKD is presented (a trusted device is a known device by definition), so `beforeRequest` only attaches `x-descope-known-device-token` when there's no DTD to send instead. DTD always takes precedence when both are present in localStorage. This is a deliberate optimization on top of the DTD precedent, not present in #1290 - covered by new tests in `beforeRequest-dkd.test.ts` (`DTD takes precedence over DKD - never sent together`).

## Must
- [x] Tests

### DIFF
--- a/packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
+++ b/packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
@@ -88,8 +88,7 @@ export const persistTokens = (
   refreshTokenViaCookie: boolean | CookieConfig = false,
 ): LastCookieOptions | undefined => {
   // persist refresh token
-  const { sessionJwt, refreshJwt, trustedDeviceJwt, knownDeviceJwt } =
-    authInfo;
+  const { sessionJwt, refreshJwt, trustedDeviceJwt, knownDeviceJwt } = authInfo;
   let cookieOptions: LastCookieOptions | undefined;
 
   if (refreshJwt) {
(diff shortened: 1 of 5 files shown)


packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
```diff
@@ -283,5 +304,18 @@ export const beforeRequest =
       };
     }
 
+    // A trusted device is a known device by definition - the backend derives known-device state
+    // from a valid trusted-device token whenever no DKD is presented, so there's no need to send
+    // both on every request. Only attach DKD when there's no DTD to send instead.
+    if (!dtd) {
```
yosiharan: I think this might be an over-optimization: while it is true that *today* if DTD is sent then DKD is redundant, that might change in the future, and then this logic would cause trouble. Given that this optimization only saves 1 small cookie from being sent, I don't think it's worth it and we should probably just send DKD each time unconditionally 🙏

yosiharan (commented):
Looks great, just 1 small comment 🙏

packages/sdks/web-js-sdk/src/enhancers/withPersistTokens/helpers.ts
OfekAvergil: [CLAUDE 🤖] Agreed - applied. Removed the DTD-gates-DKD conditional; DKD is now sent unconditionally whenever present, same as DTD.

yosiharan (approved):
Nice 🙏

PR #1477, opened 2026-09-02
fix(web-component): do not start a flow that is not visible yet

tomerlichtash:
Fixes the widget pre-render problem in descope/etc#17399.

## Why widgets pre-render

Every flow attached to a widget is rendered when the widget renders, so pressing a button opens a modal with the flow's first screen already there, with no loading gap. For a flow that starts with a screen this costs nothing: the SDK renders that screen from published config and only calls `flow/start` on the first interaction.

## The problem

A flow that starts with an action has no screen the SDK can render, so the only way to get one is to start the execution - and starting it runs the action. Pre-rendering such a flow sends the SMS, calls the connector, and so on, before the user pressed anything. A user-profile widget with five action-first auth-method flows fired all five on render.

It also re-fires after each successful flow: `onSuccess` calls `getMe()`, the state change rebuilds the preloaded modals, and the action-first flows run again.

## The fix

`descope-wc` does not start an execution while the element is not visible. When the start branch is reached and the element is hidden, it arms a `ResizeObserver` and returns. The element gets a layout box when its modal opens; the observer then re-enters `onFlowChange` through the existing `reqTimestamp` cache-buster, so the start runs with fresh flow state rather than a captured request.

Nothing is queued server side - the server never hears about the flow until it is shown. If the modal never opens, `disconnectedCallback` drops the observer and no execution ever existed.

The second commit closes the same hole for polling screens. A polling screen is the one screen that starts its own execution: rendering it fires the polling interaction (the server never volunteers `action: poll` - the client kicks polling off from the screen content), and on a start screen that interaction is what calls `flow/start`. Pre-rendered into a closed modal, it started the flow before anyone opened anything. It now waits for visibility too.

Details:

- **Native bypasses the gate.** The native layer renders screens itself, so its element is never visible.
- **Fails open.** Engines without `checkVisibility()` (Safari < 17.4) keep starting on mount, so they keep today's behavior rather than breaking.
- **No retry after a failed start.** A failed start leaves `executionId` unset, and the `flowState` update it triggers re-enters the start branch - which would re-run the action once per attempt. Deliberate restarts (components-version mismatch, attribute change) clear the guard.

Three widget mixins opened their modal only once the flow reported a page. A flow that starts on visibility cannot report a page while its modal is closed, so they now open on click - one line each, no other changes.

## Consequences to be aware of

An action-first flow's modal now opens empty and fills in when `flow/start` returns, where before it opened already rendered - because the flow had already run. Screen-first flows are unaffected and still render at widget render time. A loader in the modal is the natural follow-up and needs no SDK API.

This is not scoped to widgets: a flow embedded directly in an app inside something hidden (collapsed accordion, inactive tab) now starts when shown rather than on mount. Same correctness argument, worth a release note.

## Tests

- `test/descope-wc.deferHiddenStart.test.ts` - hidden flow does not start, starts once visible, native bypass, no retry after failure, hidden polling screen does not fire its interaction, no-`checkVisibility` fallback.
- `packages/widgets/user-profile-widget/e2e/early-execution.spec.ts` - real component in a real widget modal, counting `/v1/flow/start` at the network level: zero starts on render, exactly one per click, and screen-first flows still pre-render inside a closed modal. Green on chromium, webkit and firefox.

Full suites on the merged tree: web-component 460, user-profile e2e 33, user-management e2e 30, outbound e2e 7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ node_modules
 /connect.lock
 packages/**/coverage
 packages/**/build
+packages/**/build-preact
 /libpeerconnection.log
 npm-debug.log
 yarn-error.log
--- a/packages/libs/debug-logs/CHANGELOG.md
+++ b/packages/libs/debug-logs/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+This file was generated using [@jscutlery/semver](https://github.com/jscutlery/semver).
+
+## [0.1.1](https://github.com/descope/descope-js/compare/debug-logs-0.1.0...debug-logs-0.1.1) (2026-09-07)
+
+
+### Bug Fixes
+
+* **debug-logs:** correct types entry paths in package.json RELEASE ([#1465](https://github.com/descope/descope-js/issues/1465)) ([15b26ad](https://github.com/descope/descope-js/commit/15b26adc40510064fc2d3504658c312f3c5436b1))
--- a/packages/libs/debug-logs/package.json
+++ b/packages/libs/debug-logs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@descope/debug-logs",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "author": "Descope Team <info@descope.com>",
   "description": "CloudWatch RUM-based telemetry collector for Descope SDK",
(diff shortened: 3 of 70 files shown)


tomerlichtash:
@shuni review this pr

PR #1478, opened 2026-09-03
feat(react-sdk): support Preact via preact/compat

omercnet:
## Related Issues

Fixes: https://github.com/descope/etc/issues/18251

## In a Nutshell

- Run the **existing** react-sdk test suite a second time under `preact/compat`
- Add `test-preact/` for the compat-only checks (alias, custom-element bridge, SSR)
- Add a Preact example app + rollup config + `npm run start:preact`
- Document Preact setup and the two behavioural differences in the README
- Declare `preact` as an optional peer dependency
- Fix a real SSR crash: `customElements?.get(...)` throws on the server

## Description

`@descope/react-sdk` already worked in Preact by accident — it only uses APIs that `preact/compat` implements. This makes that official: CI now runs the SDK's tests twice, once on React and once with `react` pointed at `preact/compat`, so the same tests guard both. It also fixes a bug where the SDK crashed when rendered on a server.

## What this does

**Runs the existing test suite twice.** `jest.preact.config.js` reuses `jest.config.js` with `react`/`react-dom` mapped to `preact/compat` — the same aliasing `@preact/preset-vite` or a webpack `resolve.alias` does. The React and Preact runs share the same test files, so anything the React suite asserts is asserted under Preact too. Chained into `npm test`, so CI covers both with no pipeline change.

```
React:  10 suites, 79 tests
Preact: 13 suites, 105 tests
```

**Adds `test-preact/` for the compat-only checks** — things that are meaningless under React:
- the alias is actually in effect (`require('react') === require('preact/compat')`)
- the custom-element bridge in `withPropsMapping`: `.attr` props become kebab-case attributes, non-string values serialise to JSON rather than `[object Object]`, `.prop` props become element properties, nullish values remove the attribute
- `success` / `error` / `ready` events, the `afterRequest` hook ordering, listener teardown on unmount
- the forwarded ref resolves to the `descope-wc` element
- every widget and default flow renders its custom element
- SSR renders `<Descope />` and a widget with no DOM

**Adds `examples/preact-app`** — a Preact app (rendered with Preact's own `render`) using `AuthProvider`, `Descope`, `UserProfile`, `useSession`, `useUser`, `useDescope`, and an imperative ref. Built by `rollup.config.preact-app.mjs`; `npm run start:preact` serves it. The output bundle contains Preact and no React runtime.

**Documents it** in the README (`## Preact Support`) with bundler, package.json and tsconfig setup, and declares `preact` as an optional peer dependency.

## Behavioural differences found

Both are documented in the README. Neither blocks any SDK feature.

1. **Effects run after paint.** Preact schedules `useEffect` on the next frame; React flushes it before paint. `<Descope />` attaches its event listeners from an effect, so an event dispatched in the same tick as mount can be missed. Real flows emit these events long after mount, so this only surfaces in tests — `App.test.tsx` now waits for the listener to be attached before dispatching, via a new `testUtils/wcListeners.ts` helper that observes `addEventListener` rather than sleeping. Still green under React.

2. **SSR needs the async renderer.** `preact-render-to-string`'s sync `renderToString` throws on a suspended subtree instead of emitting the `<Suspense>` fallback, and `<Descope />` lazy-loads its web component. Consumers need `renderToStringAsync` (v6+). Pinned by a test so the README note stays honest.

## Bug fix

Writing the SSR tests surfaced a real bug:

```js
customElements?.get('descope-wc') || (await import('@descope/web-component'))
```

`customElements` is *undeclared* on the server, not merely undefined, so `?.` does not guard it — this throws `ReferenceError: customElements is not defined`, defeating the SSR-safety the surrounding `lazy()` was written for (see the comment above it). Under React the loader is `async`, so the throw becomes a rejected promise and stays latent; under Preact's async SSR it fails the render. Now reached through `globalThis`.

## Verification

- `npx nx test react-sdk --skip-nx-cache` — both suites green
- `npx rollup -c` — library builds
- `npx rollup -c rollup.config.preact-app.mjs` — example app builds
- `npm run lint` — clean
- `pnpm install --frozen-lockfile` — lockfile up to date

## Notes

- Lockfile diff is +30/-0: only `preact`, `preact-render-to-string` and `@rollup/plugin-alias`. A plain `pnpm add` also drags in a `@descope/web-components-ui` 3.19.0 -> 3.20.2 bump via the root `postinstall`; that was kept out of this PR deliberately.
- No change to the shipped runtime other than the `globalThis` fix — Preact support needs nothing from the SDK itself.
- Two CI failures during this PR were investigated and found unrelated (a `debug-logs` wall-clock assertion and `user-management-widget` e2e `waitForResponse` timeouts); both have standing-down comments below with proposed patches for their owners.

## Must

- [ ] Tests

https://claude.ai/code/session_01PeQCEtMD3vAXm1RNHN4uL7

### DIFF
--- a/packages/sdks/react-sdk/package.json
+++ b/packages/sdks/react-sdk/package.json
@@ -52,7 +52,10 @@
     "lint": "eslint '+(src|examples)/**/*.+(ts|tsx)' --fix",
     "prepublishOnly": "npm run build",
     "start": "npx nx run react-sdk:build && rollup -c rollup.config.app.mjs -w",
-    "test": "jest"
+    "start:preact": "npx nx run react-sdk:build && rollup -c rollup.config.preact-app.mjs -w",
+    "build:preact-app": "rollup -c rollup.config.preact-app.mjs",
+    "test": "jest && npm run test:preact",
+    "test:preact": "jest -c jest.preact.config.js"
   },
   "lint-staged": {
     "+(src|examples)/**/*.{js,ts,jsx,tsx}": [
@@ -60,18 +63,18 @@
     ]
   },
   "dependencies": {
-    "@descope/sdk-helpers": "workspace:*",
     "@descope/access-key-management-widget": "workspace:*",
+    "@descope/applications-portal-widget": "workspace:*",
     "@descope/audit-management-widget": "workspace:*",
+    "@descope/core-js-sdk": "workspace:*",
+    "@descope/outbound-applications-widget": "workspace:*",
     "@descope/role-management-widget": "workspace:*",
+    "@descope/sdk-helpers": "workspace:*",
+    "@descope/tenant-profile-widget": "workspace:*",
     "@descope/user-management-widget": "workspace:*",
     "@descope/user-profile-widget": "workspace:*",
-    "@descope/applications-portal-widget": "workspace:*",
-    "@descope/outbound-applications-widget": "workspace:*",
     "@descope/web-component": "workspace:*",
-    "@descope/web-js-sdk": "workspace:*",
-    "@descope/core-js-sdk": "workspace:*",
-    "@descope/tenant-profile-widget": "workspace:*"
+    "@descope/web-js-sdk": "workspace:*"
   },
   "devDependencies": {
     "@babel/core": "7.26.0",
@@ -79,6 +82,8 @@
     "@babel/preset-react": "7.26.3",
     "@babel/preset-typescript": "7.26.0",
     "@open-wc/rollup-plugin-html": "^1.2.5",
+    "@remix-run/router": "1.21.0",
+    "@rollup/plugin-alias": "5.1.1",
     "@rollup/plugin-commonjs": "^28.0.0",
     "@rollup/plugin-node-resolve": "^15.0.0",
     "@rollup/plugin-replace": "^5.0.0",
@@ -88,11 +93,12 @@
     "@testing-library/react-hooks": "8.0.1",
     "@testing-library/user-event": "14.5.2",
     "@types/jest": "^29.0.0",
-    "@types/react": "18.3.18",
     "@types/node": "^20.0.0",
+    "@types/react": "18.3.18",
     "@types/react-dom": "18.3.5",
     "@types/react-router-dom": "^5.3.3",
     "babel-jest": "29.7.0",
+    "core-js": "3.40.0",
     "eslint": "8.57.1",
     "eslint-config-airbnb": "19.0.4",
     "eslint-config-airbnb-typescript": "17.1.0",
@@ -113,13 +119,17 @@
     "eslint-plugin-react-hooks": "4.6.2",
     "eslint-plugin-testing-library": "6.2.2",
     "jest": "^29.0.0",
+    "jest-environment-jsdom": "^29.0.0",
     "jest-extended": "^4.0.0",
     "lint-staged": "^13.0.3",
+    "object-assign": "^4.1.1",
     "oidc-client-ts": "3.2.0",
+    "preact": "^10.29.8",
+    "preact-render-to-string": "^6.7.0",
     "pretty-quick": "^3.1.3",
     "react": "18.3.1",
-    "react-router": "6.28.1",
     "react-dom": "18.3.1",
+    "react-router": "6.28.1",
     "react-router-dom": "6.28.1",
     "rollup": "^4.0.0",
     "rollup-plugin-auto-external": "^2.0.0",
@@ -129,20 +139,22 @@
     "rollup-plugin-dotenv": "^0.5.0",
     "rollup-plugin-dts": "^6.1.1",
     "rollup-plugin-livereload": "^2.0.5",
+    "rollup-plugin-no-emit": "1.2.1",
     "rollup-plugin-serve": "^3.0.0",
     "rollup-plugin-terser": "^7.0.2",
+    "scheduler": "^0.25.0",
     "ts-jest": "^29.0.0",
     "ts-node": "10.9.1",
-    "typescript": "^5.0.2",
-    "object-assign": "^4.1.1",
-    "scheduler": "^0.25.0",
-    "@remix-run/router": "1.21.0",
-    "jest-environment-jsdom": "^29.0.0",
-    "core-js": "3.40.0",
-    "rollup-plugin-no-emit": "1.2.1"
+    "typescript": "^5.0.2"
   },
   "peerDependencies": {
     "@types/react": ">=16",
+    "preact": ">=10.11.0",
     "react": ">=16"
+  },
+  "peerDependenciesMeta": {
+    "preact": {
+      "optional": true
+    }
   }
 }
--- a/packages/sdks/react-sdk/test-preact/compat.test.tsx
+++ b/packages/sdks/react-sdk/test-preact/compat.test.tsx
@@ -1,0 +1,261 @@
+/**
+ * Preact-only assertions. The shared `test/` suite already runs against
+ * `preact/compat` (see `jest.preact.config.js`); this file covers the parts
+ * that are meaningless under React - that the alias is actually in effect, and
+ * that the custom-element bridge behaves identically once it is.
+ */
+import { fireEvent, render, waitFor } from '@testing-library/react';
+import React from 'react';
+import {
+  AccessKeyManagement,
+  ApplicationsPortal,
+  AuditManagement,
+  AuthProvider,
+  Descope,
+  OutboundApplications,
+  RoleManagement,
+  SignInFlow,
+  SignUpFlow,
+  SignUpOrInFlow,
+  TenantProfile,
+  UserManagement,
+  UserProfile,
+  useDescope,
+  useSession,
+  useUser,
+} from '../src';
+
+Object.defineProperty(global, 'Response', {
+  value: class {},
+  configurable: true,
+  writable: true,
+});
+
+jest.mock('@descope/web-component', () => ({ default: {} }));
+jest.mock('@descope/user-management-widget', () => ({ default: {} }));
+jest.mock('@descope/user-profile-widget', () => ({ default: {} }));
+jest.mock('@descope/role-management-widget', () => ({ default: {} }));
+jest.mock('@descope/access-key-management-widget', () => ({ default: {} }));
+jest.mock('@descope/audit-management-widget', () => ({ default: {} }));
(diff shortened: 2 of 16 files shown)


packages/sdks/react-sdk/test-preact/compat.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+// Waits for the web component to mount *and* for the SDK's `useEffect`s to
+// have run - under Preact effects are scheduled after paint, so the element
+// being in the DOM does not imply its listeners are attached yet.
+const findWc = async (container: HTMLElement, tag = 'descope-wc') => {
+  await waitFor(() => expect(container.querySelector(tag)).toBeTruthy());
+  await new Promise((resolve) => {
+    setTimeout(resolve, 0);
+  });
+  return container.querySelector(tag) as HTMLElement;
```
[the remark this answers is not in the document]
omercnet: Fixed in cfc8ab1 — the `setTimeout(0)` is gone.

New `testUtils/wcListeners.ts` records `addEventListener` calls, so a test can poll for the attachment itself rather than sleeping. The helper is now split by what each assertion actually needs:

- `findWc` waits only for the element. Attributes, properties, the forwarded ref and the widget tests are all applied from the ref callback, which runs *before* effects, so there is nothing to flush for those.
- `findFlowWc` additionally waits for the `success` listener — the one `<Descope />` registers unconditionally, so its attachment marks the whole effect as having run. Only the three event tests use it.

Worth noting your suggestion on the other thread proposed the same fixed `setTimeout(0)` this comment argues against; the shared helper resolves both.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sdks/react-sdk/test/components/App.test.tsx
```diff
@@ -99,16 +99,20 @@ describe('App', () => {
...
+    await waitFor(() => {
+      fireEvent(
+        // eslint-disable-next-line testing-library/no-container
+        container.querySelector('descope-wc'),
+        new CustomEvent('error', {}),
+      );
+
+      // ensure error is shown
+      expect(document.querySelector('.error')).not.toBeNull();
+    });
```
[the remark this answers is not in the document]
omercnet: Fixed in cfc8ab1 — the dispatch is out of the retry callback.

I did not take the suggested diff as written, because its `await new Promise(resolve => setTimeout(resolve, 0))` is the same fixed-delay race you flagged on `test-preact/compat.test.tsx`. Dispatching before the listener is attached loses the event outright (it is not deferred), so a later `waitFor` on `.error` cannot recover it — the wait has to happen *before* the dispatch, and it has to be a real condition.

So the test now waits on the listener attachment via the new `waitForListener` helper, dispatches exactly once, then asserts:

```js
const wc = container.querySelector('descope-wc');
await waitForListener(wc, 'error');

fireEvent(wc, new CustomEvent('error', {}));

await waitFor(() => expect(document.querySelector('.error')).not.toBeNull());
```

That also leaves the test safe to extend with a `toHaveBeenCalledTimes(1)` assertion, which was your underlying concern.

---
_Generated by [Claude Code](https://claude.ai/code)_

omercnet:
### CI: `debug-logs:test` failure is not this PR's

`👷 Build / Lint / Test (Node default)` failed on `dea5fd5`. The failing task is `debug-logs:test` — every other task in the run passed, including `react-sdk:test`.

**What failed** — `packages/libs/debug-logs/test/integration.test.ts:550`:

```
● TelemetryManager Integration › Performance Considerations › should not significantly delay console calls

  expect(received).toBeLessThan(expected)
  Expected: < 1
  Received:   1.5759456299999874
```

```js
const startTime = performance.now();
for (let i = 0; i < iterations; i++) console.log(`Message ${i}`);
const avgTime = (performance.now() - startTime) / iterations;

// Each call should be very fast (< 1ms on average)
expect(avgTime).toBeLessThan(1);
```

**Why it isn't this PR's.** It's a bare wall-clock threshold with no headroom, and it came in 0.58ms over on a loaded shared runner. This PR has a zero-line diff against `packages/libs/debug-logs` — `debug-logs` is only in the affected set because the change touches root `.gitignore` and `pnpm-lock.yaml`. `debug-logs:test` passes locally on this branch: 7 suites, 192 tests.

I'm not fixing it here, because that would widen a Preact-support PR into an unrelated package, and no upstream fix exists to port. The push to `cfc8ab1` re-ran the full suite; that run is the re-run.

**Proposed patch** (for whoever owns `debug-logs`) — the test's own name says "not *significantly* delay", so measure against a baseline instead of the absolute clock:

```js
const time = (fn) => {
  const start = performance.now();
  for (let i = 0; i < iterations; i++) fn(i);
  return performance.now() - start;
};

const baseline = time((i) => noopSink(`Message ${i}`)); // console with telemetry off
manager = new TelemetryManager(config, context);
const instrumented = time((i) => console.log(`Message ${i}`));

// Instrumentation should not be an order of magnitude slower than the sink.
expect(instrumented).toBeLessThan(Math.max(baseline * 10, 50));
```

That keeps the regression signal (it still catches a 10x slowdown) while being insensitive to how loaded the runner is. The one-line alternative, if a baseline is too fiddly, is to raise the bound to something with real headroom (`toBeLessThan(5)`) — it still catches a 100x regression and never fires on runner noise.

---
_Generated by [Claude Code](https://claude.ai/code)_

omercnet:
### CI: `user-management-widget:test:e2e` failure is not this PR's

`👷 Build / Lint / Test (Node 18)` failed on `cfc8ab1`. The failing task is `user-management-widget:test:e2e`; the other eight e2e projects (`web-component`, `role-management`, `user-profile`, `applications-portal`, `access-key-management`, `audit-management`, `tenant-profile`, `outbound-applications`) all passed, as did `test`, `lint`, `build` and `licenseCheck`.

**The decisive evidence:** the *same* `nx affected --target test:e2e` target on this *same* commit **passed on the sibling job** (Node default, 11m 4s). Same code, same specs, different runner.

**What failed** — 3 failed, 9 flaky, all in `e2e/user-management-widget.spec.ts`, all the same shape:

```
[firefox] › user-management-widget.spec.ts:772:7 › widget › filter users by status
[firefox] › user-management-widget.spec.ts:909:7 › widget › filter users by roles
[firefox] › user-management-widget.spec.ts:970:7 › widget › clearing the filter sends a search with no filter fields

  Test timeout of 60000ms exceeded.
  Error: page.waitForResponse: Test timeout of 60000ms exceeded.

  > 806 |     const searchResponsePromise = page.waitForResponse(
```

The 9 "flaky" entries are the same specs on chromium and webkit — they timed out and then passed on retry. Firefox happened to burn both its retries. A suite where a third of the file needs retries to go green is a runner-throughput problem, not a code one.

**Why it isn't this PR's.** Zero-line diff against `packages/widgets/user-management-widget` and against every e2e project and helper. All nine are in the affected set only because the change touches root `.gitignore` and `pnpm-lock.yaml`; `react-sdk` has no `test:e2e` target of its own. Combined with the same commit passing this target on the sibling job, there is no mechanism by which this diff produces a `waitForResponse` timeout in the user-management widget.

I have re-run the failed jobs (that is the one re-run). I am not fixing the specs here — that would widen a Preact-support PR into an unrelated widget's e2e suite, and no upstream fix exists to port.

**For whoever owns `user-management-widget`:** these specs race a 60s `page.waitForResponse` against widget bootstrap on a runner already executing nine e2e projects. The durable fix is to await the widget's ready state before arming the response promise, rather than relying on retries — the current pass rate is being bought entirely by `retries: 2`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sdks/react-sdk/package.json
```diff
@@ -52,33 +52,38 @@
     "lint": "eslint '+(src|examples)/**/*.+(ts|tsx)' --fix",
     "prepublishOnly": "npm run build",
     "start": "npx nx run react-sdk:build && rollup -c rollup.config.app.mjs -w",
-    "test": "jest"
+    "start:preact": "npx nx run react-sdk:build && rollup -c rollup.config.preact-app.mjs -w",
+    "build:preact-app": "rollup -c rollup.config.preact-app.mjs",
+    "test": "jest && npm run test:preact",
+    "test:preact": "jest -c jest.preact.config.js"
```
asafshen, line 58: ~~`test:preact` is not running in CI, I think it should~~
I was wrong - it actually does run

packages/sdks/react-sdk/package.json
```diff
@@ -52,33 +52,38 @@
     "lint": "eslint '+(src|examples)/**/*.+(ts|tsx)' --fix",
     "prepublishOnly": "npm run build",
     "start": "npx nx run react-sdk:build && rollup -c rollup.config.app.mjs -w",
-    "test": "jest"
+    "start:preact": "npx nx run react-sdk:build && rollup -c rollup.config.preact-app.mjs -w",
```
asafshen, line 55: nit: but the ports will collide (both on 3000), not sure if worth separating

asafshen (dismissed):
nice

packages/sdks/react-sdk/package.json
omercnet: Good catch, and it's a real collision rather than a nit — fixed in afa2d0e.

`rollup-plugin-browsersync` is a five-line passthrough: it hands its options object straight to `bs.init()`, and browser-sync defaults to port 3000 when none is given. Neither example config set one, so `npm start` and `npm run start:preact` both grabbed 3000 and could not run side by side — which defeats the point of having both.

The Preact example is now pinned to 3001; the React one keeps 3000 untouched:

```js
browsersync({ server: 'build-preact', single: true, port: 3001 })
```

Verified the build still passes and confirmed the passthrough by reading the plugin source. I could not stand a long-lived watcher up in this sandbox to bind-test 3001 live, so that part is from the source rather than observed.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sdks/react-sdk/package.json
omercnet: Right — `npm test` is `jest && npm run test:preact`, so the existing `nx affected --target test` picks both up with no pipeline change. That was deliberate: chaining it means nobody has to remember to add a CI step, and the Preact run can't silently stop being exercised.

You can see both in the CI log for this PR — `10 suites / 79 tests` for the React run, then `13 suites / 105 tests` for the Preact one.

---
_Generated by [Claude Code](https://claude.ai/code)_

asafshen (approved):

PR #1482, opened 2026-09-09
feat(web-component): track client-side form validation errors RELEASE

nirgur:
## Summary
- Adds `validationTrackingMixin` to the flow web-component. It captures native form
  validation failures (field, rule, message, screen) at the two existing validation
  points (submit in `#validateInputs`, and blur), batches them, and best-effort relays
  them to the new `POST /v1/flow/event` backend endpoint.
- **Off unless the flow turns it on.** The mixin starts disabled and `DescopeWc` enables
  it from `clientValidationTrackingEnabled` in the flow's `config.json` entry, so a
  missing or stale config collects nothing.
- Never blocks or fails the flow: all capture/transport errors are swallowed. Uses a
  `keepalive` fetch so abandonment (tab close) still reports.
- No public event or type is added; the mixin reads flow context via a private
  `#currentFlowContext` getter on `DescopeWc`.

## Linked
- Ticket: descope/etc#15181
- Related PRs:
  - backend (endpoint + per-flow setting): https://github.com/descope/backend/pull/2571
  - console-app (the toggle): https://github.com/descope/console-app/pull/5802
  - content (Segment connector step type): https://github.com/descope/content/pull/2323

## Changes
- NEW `mixins/validationTrackingMixin.ts` (+ tests): capture, batch, dedupe the
  blur+submit double-fire, flush (debounce / size cap / screen change / unload), bounded
  retry on non-unload failures (5xx and 429 only - a 4xx will not become a 200).
- `descope-wc/DescopeWc.ts`: private `#currentFlowContext` getter; collect invalid
  inputs in `#validateInputs`; enable from the flow config; teardown on disconnect.
- `helpers/helpers.ts`: `handleReportValidityOnBlur` gains an optional `onInvalid` callback.
- `mixins/index.ts`, `descope-wc/BaseDescopeWc.ts`: register the mixin.

## Event shape

Validation only happens on screens, so each event is located by its screen - there is no
step in this model:

```json
{"executionId": "...", "events": [
  {"id": "...", "field": "email", "rule": "format",
   "message": "Must be a valid email",
   "screenId": "sign-up-or-in/SC...", "screenName": "Welcome Screen", "ts": 1788985221608}]}
```

## Known limitation: errors before the flow starts

Events are keyed to a flow execution id, and `/v1/flow/event` only accepts a live one.

The start screen renders from `config.json` before `/v1/flow/start`, so there is no
execution id yet. Errors there are held and sent once the flow starts, keeping the screen
they were captured on (`startScreenId` / `startScreenName` from the config).

If the flow never starts, the user abandoned on the first screen and the held events are
dropped. Covering that would mean starting executions earlier or accepting events without
one, which weakens the auth model - a separate decision, tracked on descope/etc#15181.

## Test plan
- Unit: 24 mixin tests (off by default, on/off toggle, holding before an execution with
  cap + dedupe, keeping the captured screen, no double-send, retry rules, teardown).
- Component: drives the component's own submit path to check screen identity comes from
  `config.json` on the start screen and from the flow on later screens.
- Browser e2e (chromium/firefox/webkit): real clicks, real `POST /v1/flow/event`.
- End to end on a local stack: toggled in the console, verified the column in Postgres,
  the key in the served `config.json`, and the request on the wire - and that unchecking
  it sends nothing, on both the start screen and a later screen.

## Risk / review focus
- Touches the hot submit/blur path - capture is best-effort and must never throw into
  the flow. Worth a close look.
- `message` is whatever text the user saw. That is usually the flow's own message, but a
  browser default can include what the user typed (Chromium's invalid-email message quotes
  it). Deliberate, pending the discussion on the ticket.


nirgur:

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,284 @@
...
+            }
+          })
+          .catch(() => {
+            if (retriesLeft > 0) {
+              this.#scheduleRetry(url, init, retriesLeft);
+            }
+          });
+      }
+
+      #scheduleRetry(url: string, init: RequestInit, retriesLeft: number) {
```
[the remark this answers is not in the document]
nirgur: Fixed in 43f01cf - the retry timers are now tracked and cleared in `teardownValidationTracking`, same as the flush timer. Added a test that a retry waiting out its backoff never fires after disconnect.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,323 @@
...
+      }
+
+      // Drop everything buffered without sending it, and cancel anything
+      // already scheduled. Used when capture is switched off. Listeners that
+      // were attached while it was on are left in place - they only call
+      // #flush, which does nothing on an empty buffer.
+      #discard() {
+        clearTimeout(this.#flushTimer);
+        this.#retryTimers.forEach(clearTimeout);
+        this.#retryTimers.clear();
```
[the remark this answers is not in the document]
nirgur: Fixed in 05e02bf. #scheduleRetry now returns early when tracking is off, so a request already in flight when the switch flips cannot schedule further attempts. Test added, and I checked it fails without the guard.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1834,8 +1857,35 @@ class DescopeWc extends BaseDescopeWc {
...
+  // field in the model; config.json's startScreenName is the same task name.
+  get #currentFlowContext() {
+    const flow = this.flowState?.current;
+    return {
+      executionId: flow?.executionId,
+      screenId: flow?.screenId || this.#startScreenId,
+      screenName:
+        this.stepState?.current?.stepName ||
+        flow?.stepName ||
+        this.#startScreenName,
```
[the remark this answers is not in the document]
nirgur: Fixed in 05e02bf. Confirmed the premise - #handleFlowRestart clears only stepId and executionId, so flowState kept the previous screen. #currentFlowContext now returns the config start screen whenever there is no live execution, instead of preferring the stale flow state.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -696,7 +699,22 @@ class DescopeWc extends BaseDescopeWc {
...
+    }
     const flowConfig = await this.getFlowConfig();
+    // Per-flow switch from config.json. Absent means off.
+    this.setValidationTrackingEnabled(
+      !!flowConfig.clientValidationTrackingEnabled,
+    );
+    // The start screen renders before the flow starts, so validation errors
+    // there are held until an execution exists. This is that moment.
+    if (executionId && isChanged('executionId')) {
+      this.adoptPendingValidationErrors({ executionId, stepId });
```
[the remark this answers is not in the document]
nirgur: Fixed in 8636d50. The completed branch of #handleSdkResponse now hands the execution over before dispatching success, so a flow that finishes on its first submit no longer drops the errors held from the start screen. Covered by a browser e2e case - it failed against the pre-fix build and passes after.

packages/sdks/web-component/e2e/validationTracking.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+// Live-browser cover for the per-flow validation-tracking switch. The jsdom
+// tests prove the wiring; this proves it in a real engine, with real
+// ValidityState and a real request going out (or not).
+
+const componentsVersion = '1.2.3';
+
+// Must match the flow-id the demo page renders, or getFlowConfig resolves to an
+// empty object and every flag reads as off.
+const DEMO_FLOW_ID = 'sign-up-or-in';
```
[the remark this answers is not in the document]
nirgur: Fixed in 8636d50. The spec now reads the flow-id off the demo page instead of hardcoding sign-up-or-in, so it survives a demo built with a different DESCOPE_FLOW_ID. I hit exactly this failure while testing against a local stack.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -696,7 +699,25 @@ class DescopeWc extends BaseDescopeWc {
     const { outboundAppId } = this;
     const { outboundAppScopes } = this;
     const loginId = this.sdk.getLastUserLoginId();
+    // Switching flows: turn validation tracking off before we know the new
+    // flow's setting, so the previous flow's "on" can't carry over while the
+    // config resolves (or stay on for good if it never does).
+    if (isChanged('flowId')) {
+      this.setValidationTrackingEnabled(false);
+    }
```
nirgur, line 719: Don't you prefer to do it in the attribute changed callback fn?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1834,8 +1862,44 @@ class DescopeWc extends BaseDescopeWc {
     this.#handlePageSwitchTransition(injectNextPage);
   }
 
+  // The start screen is rendered from config.json before the flow starts, so
+  // the flow state has no screen id yet. config.json is the only place that
+  // knows it.
+  #startScreenId?: string;
+
+  #startScreenName?: string;
```
nirgur, line 1877: Why do we need it if we already have currentFlowContext?

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -600,7 +600,13 @@ export const handleAutoFocus = (
   }
 };
 
-export const handleReportValidityOnBlur = (rootEle: HTMLElement) => {
+export const handleReportValidityOnBlur = (
+  rootEle: HTMLElement,
+  // Best-effort callback invoked with the input when it fails validation on
+  // blur, so callers can track client-side validation errors. Never throws
+  // into the blur handler.
+  onInvalid?: (inputs: HTMLInputElement[]) => void,
```
nirgur: why inputs is an array if we always send only one input?

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,346 @@
...
+            method: 'POST',
+            headers: {
+              'Content-Type': 'application/json',
+              Authorization: `Bearer ${projectId}`,
+            },
+            body: JSON.stringify({ executionId, events }),
+          };
+          if (isUnload) {
+            // Page is going away: single keepalive shot, retry isn't possible.
+            fetch(url, { ...init, keepalive: true }).catch(() => {});
```
nirgur: I think we should add it to the low level sdks, like we already have flow.start & flow.next

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,346 @@
...
+        const executionId = this.#bufferExecutionId;
+        this.#buffer = [];
+        this.#bufferExecutionId = undefined;
+
+        const { projectId } = this;
+        // The SDK instance lives on the host (DescopeWc, a subclass in the
+        // compose() chain, so it exists at runtime).
+        const { sdk } = this as unknown as {
+          sdk?: { httpClient?: { buildUrl?: (path: string) => string } };
+        };
```
nirgur: This is an indirect dependency on the mixin consumer
if we have a better way to ensure the sdk exists lets do it

if it's too complex and require many changes, lets skip it for now

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,346 @@
...
+          })
+          .catch(() => {
+            if (retriesLeft > 0) {
+              this.#scheduleRetry(url, init, retriesLeft);
+            }
+          });
+      }
+
+      #scheduleRetry(url: string, init: RequestInit, retriesLeft: number) {
+        // A request already in flight when tracking is switched off would
```
nirgur: This comment is duplicate

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur, line 1: Is there a way to keep to footprint in DescopeWc minimal?
right now it's a bit spread

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Removed in dcd2eff. It came from a bad merge restore on my side.

packages/sdks/web-component/src/lib/helpers/helpers.ts
nirgur: Good catch - changed in dcd2eff. The blur helper now takes a single input and DescopeWc wraps it, since trackValidationErrors still takes a batch for the submit path where several fields can fail at once.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Reduced in dcd2eff. The four statements in onFlowChange are now one #syncValidationTracking(flowConfig, executionId, isChanged) call, so what DescopeWc knows about tracking is: that call, two capture call sites (submit and blur, which already existed as validation points), the #currentFlowContext getter, and teardown on disconnect. If the send moves into the low-level SDK (your other comment), the cast to reach this.sdk goes away too.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: #currentFlowContext is where they are read; these two fields are the only place they can be read from. Before the flow starts, flowState has no screenId or stepName for the config-rendered start screen - and after a restart it still holds the previous screen - so the values have to come from flowConfig. The getter is sync and sits on the submit path, while getFlowConfig() is async, hence caching them when the config resolves rather than fetching in the getter.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: I looked at that. attributeChangedCallback sees flow-id changing, but the on/off value and the start screen both come from config.json, which is fetched async - so the callback would have to kick off that fetch itself and we would end up with two places racing to set the same state. The one part that could live there is the reset on flow-id change; the rest has to wait for the config either way, which is why it sits right after the await. Happy to move the reset if you prefer it there, but it would split the wiring across two places for one line.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: If we move the send into the low-level SDK (your other comment), this cast goes away with it - the mixin would call this.sdk.flow.event(...) and never touch httpClient. So I would fold this into that change rather than invent a second mechanism. If we decide against the SDK change, the alternative is passing the sdk in explicitly when the mixin is composed, which is more plumbing than it is worth for one call site.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Agreed, that is the right place for it. Sketch of what it takes:

`core-js-sdk/src/sdk/flow/index.ts` gains an `event` alongside `start`/`next`:

```ts
event: withEventValidations(
  (executionId: string, events: FlowValidationEvent[]) =>
    transformResponse(httpClient.post(apiPaths.flow.event, { executionId, events })),
),
```

plus `apiPaths.flow.event` and the event type. The mixin then calls `this.sdk.flow.event(...)` and both the raw fetch and the `httpClient.buildUrl` reach-through disappear - which also settles your other comment about the indirect dependency.

One wrinkle: the unload path sends with `keepalive: true` so the batch survives the tab closing, and `httpClient.post` has no way to pass that today. So it needs a small addition to `sendRequest` options in core-js-sdk. Worth doing rather than keeping a raw fetch just for that case.

Want it in this PR, or as a follow-up? It touches a package the other SDKs share, so it is not a drive-by - I would rather do it deliberately than bolt it on here.

nirgur: lets do it as part of this PR
we should follow the convention

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: but we will still have indirect dependency in this.sdk right?
can we do something about it?

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Done in b455fc2a. `flow.event(executionId, events, keepalive)` now lives in core-js-sdk next to `flow.start` / `flow.next`, built the same way (`withValidations(stringNonEmpty(...))` + `transformResponse`), with the path in `apiPaths.flow.event` and a `FlowValidationEvent` type.

One thing this needed: `keepalive` had to be threaded through `httpClient` (`HttpClientReqConfig`, `RequestConfig`, `sendRequest`, `post`), because the unload send depends on it. It is only added to the `RequestInit` when true, so no existing caller changes behavior.

Verified live against a local stack, not just mocks: the real request goes to `https://<base>/v1/flow/event` with `Authorization: Bearer <projectId>` and the backend returns 200. The keepalive path was verified separately by navigating away 350ms after the error (inside the 2000ms flush debounce) and confirming the POST still reached nginx with a 200 - and a control run with the keepalive flag stripped never reached the server at all.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Fixed in b455fc2a - the indirect dependency is gone. The mixin no longer reads `this.sdk` at all. It takes a sender from the host instead:

```ts
export type ValidationSender = (
  batch: ValidationBatch,
  options: { keepalive: boolean },
) => Promise<ValidationSendResult>;
```

`DescopeWc` supplies it once in the constructor, mapping the SDK response to `{ ok, retryable }`. So the mixin owns capture and batching only, and knows nothing about URLs, tokens, or the SDK. `projectIdMixin` also dropped out of its `compose` chain, since building the Authorization header was the only reason it was there.

This also surfaced a real bug: `#flush` drained the buffer before checking it had a way to send, so a flush that ran before the sender was set silently threw the events away. It now checks first and holds. Added a test for it.

nirgur:

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1834,8 +1863,65 @@ class DescopeWc extends BaseDescopeWc {
...
+  #syncValidationTracking(
+    flowConfig: FlowConfig,
+    executionId: string,
+    isChanged: IsChanged<FlowState>,
+  ) {
+    // Per-flow switch from config.json. Absent means off.
+    this.setValidationTrackingEnabled(
+      !!flowConfig.clientValidationTrackingEnabled,
+    );
+    this.#startScreenId = flowConfig.startScreenId;
```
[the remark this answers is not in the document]
nirgur: Fixed in c983675c. Confirmed: `#syncValidationTracking` runs before the condition block, so the cache was reading `flowConfig.startScreenId`, which is absent for condition-based flows. Moved the cache update to right after the condition evaluation so it stores the resolved `startScreenId`/`startScreenName`. Also fixed the singular `condition` branch, which was dropping `startScreenName` even though `calculateCondition` returns it.

Covered by a new test using an `ELSE` condition, which I verified fails against the previous code.

packages/sdks/core-js-sdk/src/sdk/flow/index.ts
```diff
@@ -62,6 +63,25 @@ const withFlow = (httpClient: HttpClient) => ({
...
+   * nothing. `keepalive` lets a final batch survive the page going away.
+   */
+  event: withEventValidations(
+    (
+      executionId: string,
+      events: FlowValidationEvent[],
+      keepalive?: boolean,
+    ): Promise<SdkResponse<never>> =>
+      transformResponse(
+        httpClient.post(
```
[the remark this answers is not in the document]
nirgur: Fixed in c983675c. Verified the premise: `retryDelaysMs = [100, 5000, 5000]` in `createFetchLogger`, so a 503 really did cost up to 12 POSTs against an endpoint we rate limit to 10/min, and the unload send inherited those ~10s of sleeps.

Went with your first option: added a request-level `disableRetry` on `HttpClientReqConfig`/`RequestConfig` that `fetchWithRetries` honours, and set it for `flow.event`. The mixin bounded retry is now the only layer. Kept it rather than dropping our retry because the shared one only covers 503/520/521/522/524/530 - not 429 or a plain 500.

Two tests: one that `flow.event` always passes the opt-out, and one at the fetch layer proving a 503 costs one request instead of four.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+        this.#send = send;
+      }
+
+      /**
+       * The flow now has an execution. Anything captured before it started -
+       * on the start screen - can finally be attributed and sent.
+       */
+      setValidationTrackingExecution(executionId: string) {
+        if (!this.#enabled || !executionId) return;
+        this.#bufferExecutionId = executionId;
```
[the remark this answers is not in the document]
nirgur: Fixed in c983675c. Confirmed: `#flush` returned early on an empty buffer without clearing `#bufferExecutionId`, so the id survived and `ctx.executionId || this.#bufferExecutionId` reused it after a restart.

Tied the id to the batch instead: it is set only when an event is actually buffered, and cleared when the buffer empties. `setValidationTrackingExecution` now also ignores adoption with nothing held.

New test covers start -> adopt with empty buffer -> restart -> error is held for the new execution, not sent to the old one. Verified it fails against the previous code.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+            if (retriesLeft > 0) {
+              this.#scheduleRetry(batch, retriesLeft);
+            }
+          });
+      }
+
+      #scheduleRetry(batch: ValidationBatch, retriesLeft: number) {
+        // A request already in flight when tracking is switched off would
+        // otherwise keep retrying past the switch.
+        if (!this.#enabled) return;
```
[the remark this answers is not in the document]
nirgur: Fixed in c983675c. You were right on both counts, including that the existing disconnect test resolved the rejection before removal and so did not cover this.

Replaced the `#enabled` check with a generation token: captured when a send starts, bumped on disable and on teardown, and checked both before scheduling a retry and again inside the timer callback.

Two new tests, both using a deferred promise so the rejection lands *after* the event: one rejecting after teardown, one rejecting after an off -> on flow switch. Both fail against the previous code.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,350 @@
...
+        // Attach flush triggers on first real capture (nothing to flush before
+        // there is a buffered event, so there's no reason to attach earlier).
+        this.#attachListeners();
+        // Before the flow starts this stays unset and the buffer just holds.
+        this.#bufferExecutionId = ctx.executionId || this.#bufferExecutionId;
+
+        inputs.forEach((input) => {
+          // The cap also bounds a buffer that is holding because the flow never
+          // started, so an abandoned start screen cannot grow without bound.
+          if (this.#buffer.length >= MAX_BATCH_SIZE) return;
```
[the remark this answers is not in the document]
nirgur: Fixed in c983675c. A deliverable batch now flushes as soon as it hits the cap and the rest of the submit keeps collecting into a fresh batch, so nothing is lost. The hard cap stays in place while there is no execution or sender, so an abandoned start screen still cannot grow without bound.

Two tests as you suggested: 25 distinct invalid fields with a live execution now produce two batches (20 + 5) with all 25 fields present, and the same 25 with no execution still hold exactly 20.

packages/sdks/core-js-sdk/src/httpClient/types.ts
```diff
@@ -5,6 +5,14 @@ type HttpClientReqConfig = {
...
   queryParams?: { [key: string]: string };
   token?: string;
+  /** Let the request outlive the page - for best-effort sends on unload. */
+  keepalive?: boolean;
+  /**
+   * Skip the shared transient-error retry in createFetchLogger. For callers
+   * that do their own bounded retry - stacking the two multiplies the number
+   * of requests a single failure produces.
+   */
+  disableRetry?: boolean;
```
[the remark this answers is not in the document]
nirgur: Fixed in 2ccf2417. Verified: `get`/`patch`/`put`/`delete` all accept the shared request config, so they advertised `disableRetry` while only `post` forwarded it to `sendRequest` - the flag was silently doing nothing on those verbs.

Went with forwarding rather than restricting the option to POST, since a caller may reasonably want to skip the shared retry on any verb.

Worth noting the same defect applied to `keepalive`, which I also added in this PR and also only wired into `post`. Fixed alongside it.

Four tests, one per verb, each confirmed to fail before the fix.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -696,7 +712,13 @@ class DescopeWc extends BaseDescopeWc {
     const { outboundAppId } = this;
     const { outboundAppScopes } = this;
```
tomerlichtash: Not related to changes, but this can be a single line; are we sure these always exist? Perhaps it's safer to use this.* ?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -119,6 +120,18 @@ class DescopeWc extends BaseDescopeWc {
...
+
+    // The validation-tracking mixin captures and batches; delivery is ours, so
+    // validation events go out over the same SDK that runs flow.start and
+    // flow.next. `this.sdk` is read when a batch is actually sent, by which
+    // point it exists. A 4xx will never turn into a 2xx, so only 5xx and 429
+    // are worth another attempt.
+    this.setValidationTrackingSender(
+      async ({ executionId, events }, { keepalive }) => {
+        const res = await this.sdk.flow.event(executionId, events, keepalive);
+        return { ok: res.ok, retryable: res.code >= 500 || res.code === 429 };
```
tomerlichtash, line 132: Perhaps it'll be cleaner to move the retryable errors to an array.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
```diff
@@ -0,0 +1,700 @@
...
+    // type/pattern/badInput all collapse into a single `format` bucket
+    expect(deriveValidationRule({ typeMismatch: true } as ValidityState)).toBe(
+      'format',
+    );
+    expect(
+      deriveValidationRule({ patternMismatch: true } as ValidityState),
+    ).toBe('format');
+    expect(deriveValidationRule({ badInput: true } as ValidityState)).toBe(
+      'format',
+    );
```
tomerlichtash: Is this intentional? Why don't we keep the difference between type/pattern mismatch and bad input?

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
```diff
@@ -0,0 +1,700 @@
...
+    el.setValidationTrackingEnabled(false);
+    el.setValidationTrackingEnabled(true);
+
+    el.trackValidationErrors(
+      [makeInput('email', { valueMissing: true })],
+      el.currentFlowContext,
+    );
+    el.dispatchEvent(new CustomEvent('screen-updated', { detail: {} }));
+
+    expect(senderMock).toHaveBeenCalledTimes(1);
```
tomerlichtash: Just to make test foolproof consider adding another track call before turning it on again.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
```diff
@@ -0,0 +1,700 @@
...
+  const el = document.createElement('vt-test-host') as TestHost;
+  document.body.appendChild(el);
+  el.setValidationTrackingSender(senderMock);
+  return el;
+};
+
+// Mounts a host with tracking already switched on, which is what DescopeWc does
+// for a flow whose config enables it. The mixin itself starts off - see the
+// "captures nothing until it is switched on" test, which never calls the setter.
+const mount = (): TestHost => {
```
tomerlichtash: Minor, but naming is confusing here; mount and mountRaw might be confusing to read. I'd expect `mount` to be called something like `mountAndTrack` or `mountAndEnable` so test cases are easier to read.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
```diff
@@ -0,0 +1,700 @@
...
+      [makeInput('email', { valueMissing: true })],
+      el.currentFlowContext,
+    );
+    el.dispatchEvent(new CustomEvent('screen-updated', { detail: {} }));
+    expect(senderMock).toHaveBeenCalledTimes(1);
+
+    // Flow switch: off, then on for the next flow. The in-flight request
+    // belongs to the old flow and must not be retried under the new one.
+    el.setValidationTrackingEnabled(false);
+    el.setValidationTrackingEnabled(true);
```
tomerlichtash, line 705: Same as before, for foolproofing, consider adding another rejection before switching it on.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,387 @@
...
+  if (validity.valueMissing) return 'required';
+  if (validity.typeMismatch || validity.patternMismatch || validity.badInput) {
+    return 'format';
+  }
+  if (validity.tooShort) return 'too-short';
+  if (validity.tooLong) return 'too-long';
+  if (
+    validity.rangeUnderflow ||
+    validity.rangeOverflow ||
+    validity.stepMismatch
```
tomerlichtash, line 107: why not keep rules mapping 1:1?

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
```diff
@@ -0,0 +1,387 @@
...
+          this.#collect(inputs, context);
+        } catch (e) {
+          this.logger?.debug?.('Failed to track validation errors', String(e));
+        }
+      }
+
+      #collect(inputs: HTMLInputElement[], context: FlowContext) {
+        // Not enabled for this flow - capture nothing and attach no listeners.
+        if (!this.#enabled) return;
+        if (!inputs?.length) return;
```
tomerlichtash: Can be single line

tomerlichtash (dismissed):
Looks great. Added minor comments and few questions.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Merged into one line in 25f78b8c.

On the second half: both are getters on `BaseDescopeWc`, so they always resolve - a missing attribute just gives `undefined`, same as reading `this.*`. Your instinct is right that there is a difference though, just not existence: they are destructured at the top of `onFlowChange` and used ~100 and ~500 lines later, across several awaits, so the locals are a snapshot. Reading `this.*` at the point of use would pick up an attribute changed mid-flight.

Left that part alone here since it is pre-existing behaviour in a 600-line method and changing read semantics deserves its own change rather than riding along in this PR.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Left as is, happy to change if you feel strongly.

The condition mixes two kinds of test: an open-ended range (`>= 500`, any server error including ones that do not exist yet) and one specific code (`429`). An array only captures the second, so it ends up as a constant plus the range check still inline - more indirection without making the thing a reader has to understand any simpler. And the list cannot really grow: 5xx is already the catch-all and 429 is the one 4xx worth retrying.

The reason is on the comment directly above it, which is the part that was actually missing.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
nirgur: Done in 25f78b8c - good catch, and it needed one more thing than you suggested to actually bite.

Adding the `track` call while off is invisible to `toHaveBeenCalledTimes(1)`: a leaked event rides along inside the same batch, so the count stays 1 and the test passes. Added the call plus assertions on the batch contents (`events` length 1, field `email`).

Mutation-checked: removing the `#enabled` guard in `#collect` now fails it with `Expected length: 1 / Received length: 2`. Worth noting it guards the collect-time guard rather than discard-on-disable - a capture while off never reaches the buffer at all, so `#discard` is not what this exercises.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
nirgur: Renamed in 25f78b8c: `mountRaw` -> `mount` (the plain thing gets the plain name) and `mount` -> `mountAndEnable`.

Went with `mountAndEnable` over `mountAndTrack` since the helper only flips the switch - "AndTrack" reads like it also starts capturing. Agreed on the underlying problem: the shorter, more default-looking name was the one carrying extra behaviour, and tracking-on/off is the single most important piece of state in these tests.

packages/sdks/web-component/src/lib/mixins/__tests__/validationTrackingMixin.test.ts
nirgur: Took the spirit of this in 25f78b8c, with a change of angle - worth explaining why.

A promise only rejects once, so "another rejection" means moving it into the off-window. That interleaving passes on both the fixed and the previously-broken implementation: the old `#enabled` check also returned early while disabled, so it would not have caught anything. The current ordering (reject after re-enabling) is what catches the real bug.

What was genuinely missing is the other half: nothing asserted that killing the stale chain does not kill the component. Added a fresh capture after the dead rejection, asserting it still reaches the sender. Mutation-checked - a disable that also drops the sender fails it with `Expected 2 / Received 1`.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Changed to 1:1 in 25f78b8c - you were right, and the original plan actually specified distinct rules. The `format` bucket was an undocumented change I made during implementation.

Two things fell out of doing it properly:

1. `typeMismatch` only says the value is the wrong shape, not for what - it fires for `email` and `url` only (`number`/`date` produce `badInput` instead). So the rule is derived from the input `type`: `email`, `url`, else `type-mismatch`. Hardcoding `email` would mislabel a url field on a custom screen.
2. `badInput` is now checked before `valueMissing`. A control that cannot parse what was typed reports an empty value, so a required field sets BOTH flags - verified in Chromium, `type="number"` + `5e` gives `badInput: true, valueMissing: true`. The old order reported `required`, claiming the user left it blank when they had typed something.

Final mapping: `bad-input`, `required`, `email`/`url`/`type-mismatch`, `pattern`, `too-short`, `too-long`, `range`, `custom`. The browser e2e asserts `email` from a real `type="email"` input across chromium/firefox/webkit.

packages/sdks/web-component/src/lib/mixins/validationTrackingMixin.ts
nirgur: Done in 25f78b8c.

tomerlichtash (approved):

PR #1483, opened 2026-09-09
feat(core-js-sdk): support enchanted link over SMS

eliran-descope:
Adds Enchanted Link over SMS to core-js-sdk.

- Four new phone methods on the enchantedLink module.
- Additive only. Existing email methods unchanged.
- Failing e2e widget tests are pre-existing flake, unrelated to this diff.

Part of descope/etc#18315

### DIFF
--- a/packages/sdks/core-js-sdk/README.md
+++ b/packages/sdks/core-js-sdk/README.md
@@ -24,6 +24,22 @@ const loginId = 'loginId';
 sdk.otp.signIn.email(loginId);
 ```
 
+## Enchanted Link over SMS
+
+The enchanted link can be delivered by SMS instead of email, using `signInWithPhone`,
+`signUpWithPhone`, `signUpOrInWithPhone` and `update.phone.sms`. The login ID is a phone number, and
+the response is a `PhoneEnchantedLinkResponse`, which carries `maskedPhone` instead of `maskedEmail`.
+Unlike the email message, the text message contains only the correct link, so the user has nothing
+to choose. Polling with `waitForSession` and verification are unchanged.
+
+```js
+const { data } = await sdk.enchantedLink.signUpOrInWithPhone(
+  '+11234567890',
+  'https://myapp.com/verify-enchanted-link',
+);
+const jwt = await sdk.enchantedLink.waitForSession(data.pendingRef);
+```
+
 ## Outbound Connect
 
 Connect to outbound applications (e.g., Google, Microsoft) to access user data from external services.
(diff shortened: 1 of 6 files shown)


eliran-descope:
@shuni review

packages/sdks/core-js-sdk/src/sdk/enchantedLink/index.ts
```diff
@@ -87,72 +82,125 @@ const withEnchantedLink = (httpClient: HttpClient) => ({
...
+      postSignUpOrIn<PhoneEnchantedLinkResponse>(DeliveryPhone.sms),
+    ),
+
+    signUp: withSignValidations(
+      postSignUp<EnchantedLinkResponse>(DeliveryMethods.email),
+    ),
+
+    /** Send an enchanted link over SMS to sign a new user up */
+    signUpWithPhone: withSignValidations(
+      postSignUp<PhoneEnchantedLinkResponse>(DeliveryPhone.sms),
```
nirgur, line 128: go-sdk `SignUpWithPhone` fills `user.phone` when it is empty
Should we do the same here? WDYT

packages/sdks/core-js-sdk/README.md
```diff
@@ -24,6 +24,22 @@ const loginId = 'loginId';
 sdk.otp.signIn.email(loginId);
 ```
 
+## Enchanted Link over SMS
+
+The enchanted link can be delivered by SMS instead of email, using `signInWithPhone`,
+`signUpWithPhone`, `signUpOrInWithPhone` and `update.phone.sms`. The login ID is a phone number, and
+the response is a `PhoneEnchantedLinkResponse`, which carries `maskedPhone` instead of `maskedEmail`.
+Unlike the email message, the text message contains only the correct link, so the user has nothing
```
nirgur, line 32: Here we say the user has nothing to choose, but the `linkId` doc says the user should click on that link
Which one is right?

nirgur (dismissed):

packages/sdks/core-js-sdk/src/sdk/enchantedLink/index.ts
eliran-descope: backend already does it - UnmarshalUser fills user.phone from the loginId when it's
empty and the loginId parses as a phone. so the user gets the phone
either way.

go-sdk has it because SignUpWithPhone takes phone as a separate arg. here there's
only loginId to copy from.

packages/sdks/core-js-sdk/README.md
eliran-descope: README is right, the jsdoc was wrong. fixed the jsdoc

nirgur (approved):

PR #1484, opened 2026-09-09
fix(angular-sdk): set attributes before descope-wc connects RELEASE

nirgur:
## Summary

- The second and later `<descope>` mounts in a page session never started: endless
  loading, no `ready` and no `error` event. Reported by a customer (Zyte) after a
  version bump from 0.15.17 to 0.27.5, and still broken on 0.28.2.
- Angular appends an element to the DOM in its creation pass but applies `[attr.*]`
  bindings in the later update pass, so `<descope-wc>` connected before it had a
  `project-id` and the custom element failed permanently.
- The template now sits in an `<ng-template>` built by hand, so the bindings run
  while the nodes are still detached and the element always carries `project-id`
  and `flow-id` when it connects.
- Widgets were never affected — `BaseLazyWidgetComponent` already sets attributes
  before appending.

## Linked

- Ticket: https://github.com/descope/etc/issues/18415
- Related PRs: none (single-repo change)

## Changes

**Why only mounts after the first broke.** 0.27.0 (#1293, Angular SSR) made the
web component a dynamic `import()`, and importing that module is what calls
`customElements.define('descope-wc')`. So the first mount inserts an element that
is still undefined and upgrades late, after the attributes land. Every later mount
finds the class defined and upgrades on insertion, measured ~2ms before
`project-id` arrives. `init()` then bails at "Cannot get config file" and `#init`
stays `false`, so `attributeChangedCallback` drops the attribute that arrives
right after — the element is dead for good.

The declarative template itself predates 0.27; it arrived in #1181 (~0.18.3),
which moved away from `new DescopeWebComponent()` in a field initializer because
that broke on Angular 19. This fix keeps the template and all 20-odd bindings
byte-for-byte, so nothing can drift.

- `descope.component.ts` — template wrapped in `<ng-template #wcTpl>`; built in
  `ngOnInit` via `createEmbeddedView` + `detectChanges` while detached, then
  connected with `Renderer2.appendChild`. `@ViewChild`/`ngAfterViewInit` replaced,
  `ngOnDestroy` added, plus `ngSkipHydration` since the subtree is now built
  imperatively.
- The import is awaited *before* the element is built, on purpose:
  `sdkConfigOverrides` must be set first, or a flow can start without
  `persistTokens: false` and the `beforeRequest` hook.
- `descope.component.spec.ts` — new regression tests; existing tests now await the
  async insertion.
- `setup-jest.ts` — `Response` stub for jsdom (see Risk below).

## Manual test plan

Against a local backend, with `descopeBaseStaticUrl` set to
`https://static.local.descope.org/pages`:

- Load the demo app, click LOGIN — the flow renders.
- Navigate back, click LOGIN again — the second flow renders too (this is what
  used to hang forever).
- Enter an email on that second mount and press Continue — it advances to the
  verification screen, with `POST /v1/flow/start` and `/v1/flow/next` going out.
- Console shows no `project-id cannot be empty` and no `Cannot get config file`.
- A custom screen (children of `<descope>`) still renders.

Verified before/after:

| | mount 1 | mount 2 |
|---|---|---|
| before | `ready`, 14 components | `loading` forever, 0 components, 4 errors |
| after | `ready`, 14 components | `ready`, 14 components, 0 errors |

## Risk / review focus

- **SSR is not verified.** The subtree is now built imperatively, so
  `ngSkipHydration` is set — but there is no SSR harness in this repo, so that is
  reasoned, not tested. Worth a look from anyone who runs Angular SSR. The view is
  still built on the server so the rendered markup keeps containing
  `<descope-wc>`, and `nodeType` is compared to a literal rather than
  `Node.ELEMENT_NODE` because that global is not guaranteed server-side.
- **The await ordering in `ngOnInit` is load-bearing**, not stylistic — see
  Changes. Easy to "simplify" away by mistake.
- **Two spec changes look unrelated but are not.** `should emit success` was green
  while asserting nothing: jsdom has no `Response`, the handler threw, and the
  error was swallowed. This change surfaced it, so the stub moved to
  `setup-jest.ts` and the `afterRequest` mock now returns a promise. The repaired
  assertion was mutation-tested.
- The new tests were checked against the old component: 5 of them fail there, so
  they actually catch the bug.

### DIFF
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
@@ -249,7 +249,7 @@ describe('DescopeComponent', () => {
 
     // A stand-in for descope-wc that records what it can see when connected.
     class ProbeElement extends HTMLElement {
-      override connectedCallback() {
+      connectedCallback() {
         connects.push({
           projectId: this.getAttribute('project-id'),
           flowId: this.getAttribute('flow-id'),
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
@@ -5,7 +5,6 @@ import createSdk from '@descope/web-js-sdk';
 import { DescopeAuthConfig } from '../../types/types';
 import { DescopeAuthService } from '../../services/descope-auth.service';
 import {
-  ApplicationRef,
   Component,
   CUSTOM_ELEMENTS_SCHEMA,
   EventEmitter,
@@ -122,7 +121,7 @@ describe('DescopeComponent', () => {
     ).toEqual('true');
   });
 
-  it('should emit success when web component emits success', () => {
+  it('should emit success when web component emits success', async () => {
     const html: HTMLElement = fixture.nativeElement;
     const webComponentHtml = html.querySelector('descope-wc')!;
 
@@ -134,6 +133,9 @@ describe('DescopeComponent', () => {
       emitted.push(e);
     });
     webComponentHtml.dispatchEvent(new CustomEvent('success', event));
+    // The handler pipes the afterRequest promise through rxjs before emitting,
+    // so don't assume the output has arrived by the time dispatch returns.
+    await flush();
 
     expect(afterRequestHooksSpy).toHaveBeenCalled();
     expect(emitted).toHaveLength(1);
@@ -317,10 +319,10 @@ describe('DescopeComponent', () => {
 
       f.destroy();
 
-      expect(f.nativeElement.querySelector('descope-wc')).toBeNull();
       // isConnected going false is what makes the browser run the web
       // component's disconnectedCallback so it can clean up after itself.
       expect(element.isConnected).toBe(false);
+      expect(f.nativeElement.querySelector('descope-wc')).toBeNull();
     });
 
     it('projects content into the element (custom screens)', async () => {
@@ -380,17 +382,15 @@ describe('DescopeComponent', () => {
       });
       const f = TestBed.createComponent(DescopeComponent);
       f.componentInstance.flowId = 'sign-in';
-      const appRef = TestBed.inject(ApplicationRef);
-      const attachedBefore = appRef.viewCount;
 
       f.detectChanges(); // ngOnInit starts and hits the await
       f.destroy(); // destroyed before the continuation runs
       await flush();
 
-      // ngOnDestroy already ran and had no view to clean up, so the continuation
-      // must not build one - it would stay attached to ApplicationRef for good.
-      expect(appRef.viewCount).toBe(attachedBefore);
+      // The continuation must not build the view: inserting into a destroyed
+      // ViewContainerRef throws, and nothing would ever clean it up.
       expect(connects).toHaveLength(0);
+      expect(f.nativeElement.querySelector('descope-wc')).toBeNull();
     });
 
     it('still renders the element when loading the web component fails', async () => {
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
@@ -249,7 +249,7 @@ describe('DescopeComponent', () => {
 
     // A stand-in for descope-wc that records what it can see when connected.
     class ProbeElement extends HTMLElement {
-      connectedCallback() {
+      override connectedCallback() {
         connects.push({
           projectId: this.getAttribute('project-id'),
           flowId: this.getAttribute('flow-id'),
--- a/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
+++ b/packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
@@ -3,7 +3,10 @@ import { default as DescopeWC } from '@descope/web-component';
 import { DescopeComponent } from './descope.component';
 import createSdk from '@descope/web-js-sdk';
 import { DescopeAuthConfig } from '../../types/types';
+import { DescopeAuthService } from '../../services/descope-auth.service';
 import {
+  ApplicationRef,
+  Component,
   CUSTOM_ELEMENTS_SCHEMA,
   EventEmitter,
   PLATFORM_ID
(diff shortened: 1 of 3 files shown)


packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.spec.ts
```diff
@@ -117,11 +129,15 @@ describe('DescopeComponent', () => {
...
     component.success.subscribe((e) => {
-      expect(afterRequestHooksSpy).toHaveBeenCalled();
-      expect(e.detail).toHaveBeenCalledWith(event.detail);
+      emitted.push(e);
     });
     webComponentHtml.dispatchEvent(new CustomEvent('success', event));
+
+    expect(afterRequestHooksSpy).toHaveBeenCalled();
+    expect(emitted).toHaveLength(1);
+    expect(emitted[0].detail).toEqual(event.detail);
```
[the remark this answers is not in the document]
nirgur: Applied — the test is now async and awaits the emission before asserting.

One note for the record: the described failure did not reproduce. The test passed locally and in CI on both Node versions, and a mutation check (`toHaveLength(999)`) failed, which proves `emitted` held exactly one element at assertion time — zone.js drains the promise microtask inside the dispatch here. Changed anyway, since depending on that timing is fragile (it would break under zoneless change detection). (auto via /pr-watch)

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.ts
```diff
@@ -148,12 +174,58 @@ export class DescopeComponent implements OnInit, OnChanges, AfterViewInit {
...
+    // the time we get here (a route change while the chunk is downloading).
+    // ngOnDestroy has run by then and had no view to clean up, so building one
+    // now would leave it attached to ApplicationRef for good.
+    if (this.isDestroyed || this.wcView) return;
+
+    const view = this.wcTpl.createEmbeddedView(undefined);
+    // Apply the [attr.*] bindings while the nodes are detached.
+    view.detectChanges();
+    // Keep the view change-detected so later input changes reach the attributes.
+    this.appRef.attachView(view);
```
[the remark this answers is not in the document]
nirgur: Applied — the view is now inserted through a `ViewContainerRef` anchored inside the component template, so it lives in the component's change-detection tree. `ApplicationRef`, `Renderer2`, the `attachView`/`detachView` pair and the `ngOnChanges` refresh are all gone. Attributes are still applied while the nodes are detached, so the element keeps carrying `project-id` and `flow-id` when it connects.

Two notes. The staleness itself did not reproduce — `ComponentFixture.detectChanges()` routes through `appRef.tick()`, and the existing direct-assignment tests passed before this change. Made the change because the design is simpler, not because bindings were stale. Also, injecting `ViewContainerRef` directly anchors beside the host, which would move `descope-wc` outside `<descope>`; it needs an explicit `<ng-container>` anchor in the template to keep the DOM shape.

`ngOnDestroy` still destroys the view explicitly: Angular skips removing individual nodes when a parent view is going away, so without it the element never disconnects and the web component keeps its window listeners. (auto via /pr-watch)

nirgur:

nirgur:
Reverted 56e7c0ee in 948e1058.

That commit added `override` to `ProbeElement.connectedCallback()` in the spec, on the grounds that `noImplicitOverride` was making the suite fail to compile. The opposite is true: `noImplicitOverride` only requires `override` for members that exist on the base class, and `connectedCallback` is a custom-element lifecycle hook that TypeScript does not declare on `HTMLElement`. So the modifier is invalid there:

```
error TS4113: This member cannot have an 'override' modifier because it is not
              declared in the base class 'HTMLElement'.
```

CI backs that up — green on 33939ff8, red on 56e7c0ee, and reproducible locally: `Test Suites: 1 failed`, `Tests: 0 total`. The change caused the exact failure mode it described. After the revert: 24 suites, 81 tests passing.

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.ts
```diff
@@ -195,12 +269,18 @@ export class DescopeComponent implements OnInit, OnChanges, AfterViewInit {
...
 
-    this.webComponent = this.descopeWc.nativeElement;
-    this.setupNonAttributeProperties();
-    this.setupEventListeners();
+    // wcAnchor owns the view, so Angular would tear it down anyway - but when a
+    // parent view is going away Angular skips removing the individual nodes, and
+    // then the element never disconnects. Destroying it here takes the element
+    // out of the DOM, so the web component's disconnectedCallback runs and drops
+    // its window listeners and flow-state subscriptions.
+    this.wcView?.destroy();
```
asafshen, line 283: just making sure that this is not async

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.ts
```diff
@@ -148,12 +177,57 @@ export class DescopeComponent implements OnInit, OnChanges, AfterViewInit {
...
+    const view = this.wcTpl.createEmbeddedView(undefined);
+    // Apply the [attr.*] bindings while the nodes are still detached.
+    view.detectChanges();
+    this.wcView = view;
+
+    // rootNodes can include whitespace text nodes, so find the element. The
+    // nodeType is compared to a literal rather than Node.ELEMENT_NODE because
+    // this also runs during SSR, where the Node global may not exist.
+    this.webComponent = view.rootNodes.find(
+      (node: { nodeType: number }) => node.nodeType === ELEMENT_NODE
```
asafshen, line 222: I didn't quite follow the ELEMENT_NODE usage, mind to elaborate? if there are two - will they be different?

asafshen (dismissed):
nice, one question

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.ts
nirgur: No, it's all synchronous. `ngOnDestroy` returns `void`, `ViewRef.destroy()` is a synchronous call, removing the nodes from the DOM is synchronous, and the browser runs the element's `disconnectedCallback` as part of that removal. So by the time `ngOnDestroy` returns, the web component has already dropped its `window` listeners and flow-state subscriptions.

The destroy test asserts `element.isConnected === false` on the line right after `f.destroy()`, which only passes if nothing here defers.

packages/sdks/angular-sdk/projects/angular-sdk/src/lib/components/descope/descope.component.ts
nirgur: `rootNodes` is the top-level nodes of the instantiated template, and the `find` picks the element out of it.

On "if there are two" — there's only ever one per `<descope>`. The template holds a single `<descope-wc>`, and `rootNodes` belongs to that instance's own view, so it can't see another instance's element. I used `find` rather than `rootNodes[0]` so a later template-formatting change can't make index 0 a whitespace text node, and a literal rather than `Node.ELEMENT_NODE` because this also runs during SSR. Reworded the comment to say that directly.

Added tests in 4a5b69f9 covering multiple instances on one page, so this is verified rather than assumed: two different flows side by side (both get their attributes before connecting), the same flow twice (timing still holds — that shape stays unsupported by the web component, which warns about it), and destroying one instance while the other stays connected.

asafshen (approved):

PR #1488, opened 2026-09-10
feat(web-component): add lazy-start to hold a flow's start call

tomerlichtash:
Adds a `lazy-start` attribute to `descope-wc`. While set, the element preloads its flow (config, components, first screen) but holds the `flow.start` call, so no flow execution is created until the host calls `start()` on the element.

The hold wraps `sdk.flow.start` once in `#createSdk`, outside the existing retry / session-JWT wrappers, so every start call site is covered and the session token is read when the request is really sent. Calling `start()` before the flow reaches the held call is supported; the gate latches open afterwards so a flow restart is not left waiting.

Used by the user-profile widget's edit-email modal, which builds its flow ahead of being opened and starts it on open. Replaces a monkey-patch of `sdk.flow.start` from the widget side, which depended on `descope-wc` internals and could not deliver the response to the awaiting caller.

## Related Issues

Fixes https://github.com/descope/etc/issues/17399

## Related PRs

| branch       | PR         |
| ------------ | ---------- |
| service a PR | Link to PR |
| service b PR | Link to PR |

## Description

A few sentences describing the overall goals of the pull request's commits.

## Must

- [ ] Tests

### DIFF
--- a/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
+++ b/packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
@@ -402,6 +402,18 @@ class BaseDescopeWc extends BaseClass {
         }
       };
     });
+
+    // While `lazy-start` is set, hold every flow start call until start() is
+    // called. Wrapping the sdk fn holds all call sites at once, and sits outside
+    // the wrapper above so the session JWT is read when the call is really sent.
+    const startFn = this.sdk.flow.start;
+    this.sdk.flow.start = async (
+      ...args: Parameters<typeof startFn>
+    ): ReturnType<typeof startFn> => {
+      await this.#waitForStart();
+
+      return startFn.apply(this.sdk.flow, args);
+    };
   }
 
   // adds the current session JWT to a flow request options (opt-in via the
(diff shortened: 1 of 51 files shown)


packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
```diff
@@ -472,6 +484,42 @@ class BaseDescopeWc extends BaseClass {
...
+  }
+
+  // Start a flow rendered with `lazy-start`. Safe to call before the flow
+  // reached the held start call - it then goes through on arrival. The gate is
+  // latched open rather than re-armed, so a flow restart (e.g. restart-on-error)
+  // is not left waiting for another call.
+  start() {
+    this.#releaseStartGate?.();
+    this.#startGate = Promise.resolve();
+    this.#releaseStartGate = undefined;
```
nirgur: Is this needed?

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initEmailUserAttrMixin.ts
nirgur, line 1: This is only for email
What about other flows in this widget?

what about other widgets?

packages/sdks/web-component/src/lib/descope-wc/BaseDescopeWc.ts
tomerlichtash: Removed.

packages/widgets/user-profile-widget/src/lib/widget/mixins/initMixin/initComponentsMixins/initEmailUserAttrMixin.ts
tomerlichtash: Moved into `modalMixin` so it covers every modal flow in every widget: the modal driver sets `lazy-start` in `setContent` and releases it in `open()`. It's only a default — a flow that should start on render opts out with `createFlowTemplate({ lazyStart: false })`.

packages/libs/sdk-mixins/src/mixins/modalMixin/modalMixin.ts
```diff
@@ -20,8 +20,22 @@ export const modalMixin = createSingletonMixin(
         return class ModalDriverWrapper extends ModalDriver {
           setContent(template: HTMLTemplateElement) {
             loadDescopeUiComponents(template);
+            // A modal builds its flow ahead of being opened, so hold the flow's
+            // start call by default - otherwise every preloaded modal creates a
+            // flow execution the user may never open. Released in open() below.
+            // A caller that wants its flow to start on render opts out with
+            // `lazyStart: false`, which is already on the template by now.
+            const flowEle = template.content.querySelector('descope-wc');
```
nirgur: Why do you think this is the right place for it?
What if the modal does not contain a flow?

tomerlichtash: Then we can override it with false. I've put it there since most modals currently contain flows, and I thought it would be a cleaner change.

nirgur: I think we're mixing unrelated stuff here
Modal and flow are separate components, and I don't think we should have one's logic mixed into the other's mixin
Can we put it somewhere else?

tomerlichtash: Do you think we should define it per modal?

tomerlichtash: Or create an alternative fn for `setContentLazy`?

tomerlichtash: Moved it to `flowModalMixin`.

nirgur (approved):

PR #1491, opened 2026-09-15
fix(web-component): recover when a passkey prompt is never answered RELEASE

nirgur:
## Summary

- A passkey step could leave the user on a dead screen for over two minutes. Password manager extensions can leave `navigator.credentials.get/create` pending indefinitely after their prompt is dismissed, and we kept awaiting it.
- The wait is now bounded at 90 seconds. On timeout we abort, report the failure to the server, and the flow renders its normal "cancelled, please try again" screen.
- That is the whole change. One contained block in the webauthn branch, `+36/-3` in `DescopeWc.ts`.

## Linked

- Ticket: descope/etc#18538 (customer: Datavant)
- Related PRs: none

## Root cause

Bitwarden's `fido2-page-script.js` does not end the WebAuthn call when its prompt is dismissed. It requests a fallback and then calls `waitForFocus()`, which waits for a `focus` event on `window.top` **for up to five minutes** before re-running the request natively:

```js
catch (error) {
  if (fallbackSupported && error.fallbackRequested) {
    yield waitForFocus();            // races a focus event against a 5 minute timer
    return yield browserCredentials.get(options);
  }
```

Whether the user sees the hang depends on whether the page happens to hold focus at that moment. Focused, the fallback runs at once and everything recovers in ~3s. Not focused, the promise stays pending. The customer's HAR shows 136 seconds.

That also explains why this looked unreproducible: anyone with a passkey saved gets the normal picker, which rejects correctly. The bug needs a vault with *no* matching passkey, which is what the customer's video shows ("No passkeys found" + Close).

## What we decided NOT to do

The first version of this PR also **kept the other buttons on the screen active while the passkey prompt was open**, so the user could pick another method immediately instead of waiting out the timer. That half has been removed.

It ran on every passkey flow, including the ones that work today, and it forced the component to handle a passkey answer arriving after the user already moved on. Review found five separate problems in it across three rounds, the worst being a `completed` response getting dropped: `withPersistTokens` writes the tokens inside `flow.next`, so the user was signed in while the host app was never told.

Each one was fixed and then another appeared. That is a signal the change is riskier than the problem it solves, so it was pulled rather than patched further. If we later decide it is a must, it should be its own task with its own plan, built on top of a shipped and stable timeout. The removed code and its tests are in this branch's history (`7526c133`..`82e13aed`).

## Changes

| File | Change |
|---|---|
| `DescopeWc.ts` | race the browser call against `timeoutPromise`; on timeout abort and report `AbortError` / `aborted` |
| `constants/index.ts` | `WEBAUTHN_TIMEOUT = 90000` |
| `web-js-sdk/src/sdk/webauthn.ts` | optional `abort?: AbortController` on `get` / `create` (additive; mirrors the existing `conditional(options, abort)`) |

`web-js-sdk` gains an additive signature change, so it needs a version bump on release.

## Manual test plan

- Passkey sign-in, dismiss the prompt -> recovers in ~3s with `NotAllowedError`. The 90s bound must **not** fire.
- Passkey sign-in, complete it -> signs in normally, no `failure` reported.
- With Bitwarden and an empty vault, dismiss "No passkeys found" and wait -> the error screen appears at 90s instead of hanging.
- Same for the sign-up (`webauthnCreate`) flow.

## Risk / review focus

- **The timeout only fires after 90 seconds of nothing happening**, which today is a hang. It cannot change the outcome of a flow that works.
- **Late answers are harmless.** `timeoutPromise` discards the losing promise, so a browser answer arriving after the timer does nothing. Covered by a test.
- **A timeout that fires after the user navigated away** calls `flow.next` against a run that has moved, and the backend rejects it. That is exactly what `main` does today when a ceremony resolves late, so it is not new behaviour.
- **Three assertions in `descope-wc.webauthnRedirect.test.ts`** now wait for the reply instead of asserting immediately, because `timeoutPromise` defers it by a microtask. The asserted payloads are unchanged.
- **Abort is best-effort, not the mechanism.** Bitwarden honours it during its messenger phase but ignores it inside `waitForFocus`; 1Password ignores it entirely. The timer is what guarantees recovery.
- **90s is deliberately long.** Since we cannot cancel, a shorter bound risks discarding a genuine authentication if the user completes the native fallback afterwards.


nirgur:
Both findings addressed in 0bbaa7bb.

**Finding 1 (Major) — applied.** You were right, and the race only exists because this PR re-enables the alternatives. The guard ran before `flow.next` but the reply could land after the user switched methods, and it was handed to `#handleSdkResponse` unconditionally. Added a second check after the await. Covered by a new test that fails without the guard.

**Finding 2 (Minor) — applied.** This was also the webkit CI failure. The stub now sets a flag and the test waits for the browser call before `fastForward`, so the 90s timer is always registered before the clock jumps. Verified across chromium, firefox and webkit.

nirgur:
Addressed the two open threads in a055e59c.

**P2 (invalidate on non-submit transitions) — fixed, and generalised.** You were right that the counter only advanced on submit and SDK responses. Rather than hunt for every other caller, staleness is now derived from the flow state itself: the ceremony captures `executionId|stepId` and compares it when the browser answers. A popstate, an attribute-driven restart, or any future path that moves the flow invalidates a pending ceremony without anyone having to remember. The counter stays only for the two moments where no state change happens but the ceremony is dead: a submit (which runs before its response) and disconnect. Covered by a test that fails if the position check is removed.

**P1 (block method switches while the reply is in flight) — not fixing, and I think the premise is off.** `withFlowNonce` does not serialize concurrent `flow.next` calls; it attaches a nonce and keeps a high-water sequence, and its own comment says it is "Unlocked by design". There is a real but smaller risk: two overlapping requests can make the server reject one with E108201 "Request sequence validation failed". The window is a single HTTP round trip rather than the whole ceremony, and the outcome is a retryable error. Blocking the controls for that window would partly undo the escape hatch this PR is for, so I would rather leave it. Happy to reconsider if you think the nonce sequencing behaves differently than I read it.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1078,6 +1142,16 @@ class DescopeWc extends BaseDescopeWc {
...
+        // The user moved on while we were waiting. Reporting now would answer
+        // for a step the flow has already left.
+        this.loggerWrapper.debug(
+          'Ignoring a webauthn result for a step the flow already left',
+        );
+        return;
+      }
+
       // Call next with the transactionId and the response or failure
       const sdkResp = await this.sdk.flow.next(
```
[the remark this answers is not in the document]
nirgur: You are right and I was wrong. I had only read the high-water comment in `afterRequest` and concluded the enhancer did not queue; I missed the FIFO chain at lines 85-102. My earlier reply on this thread was incorrect, sorry.

Fixed in b7f306d7. The screen is still handed back while the browser dialog is up - that is the long wait, and no request is in flight there - but once the reply request starts, the controls go back under the normal in-flight lock. So nothing can queue behind the passkey reply and then run against an already-advanced flow.

Covered by a test that observes both states: usable while the ceremony is pending, disabled once the reply is in flight. Verified it fails without the change.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1078,6 +1142,16 @@ class DescopeWc extends BaseDescopeWc {
         failureReason = e.reason;
         failureMessage = e.message;
       }
+
+      if (this.#flowGeneration !== ceremonyGeneration) {
```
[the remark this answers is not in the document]
nirgur: Both points taken, fixed in b7f306d7.

1. Response-driven invalidation restored. You are right that dropping it was a regression: polling and completion responses do not move executionId/stepId, so the position check alone cannot see them. Both mechanisms are in place now.

2. popstate now invalidates synchronously, via a listener on this component rather than relying on BaseDescopeWc forwarding its state through a setTimeout subscriber. That closes the gap you described.

The new test uses a real `PopStateEvent` rather than poking `flowState` directly, which was a fair criticism of the previous one. Verified it fails without the listener.

One thing I did not add: a test for a response that does not move the step. Staging a real polling response mid-ceremony needs more scaffolding than the case warrants, so that path is covered by the code change but not directly by a test. Flagging it rather than claiming coverage I do not have.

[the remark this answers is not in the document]
nirgur: Right again, fixed in 7dc28e22.

Changing flow-id or project-id now clears a pending passkey in attributeChangedCallback, before BaseDescopeWc updates its own state. Added a test that changes flow-id mid-passkey and checks the old answer is not sent; it fails without the change.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -502,10 +540,25 @@ class DescopeWc extends BaseDescopeWc {
...
+  attributeChangedCallback(
+    attrName: string,
+    oldValue: string,
+    newValue: string,
+  ) {
+    // A new flow-id or project-id restarts the flow. BaseDescopeWc updates its
+    // own state right away but this component only hears about it on a later
+    // timer, so invalidate here or a passkey settling in between still reports.
+    if (oldValue !== newValue) {
+      this.#invalidatePendingWebauthn();
```
nirgur: This will run on any attribute change, not only attributes that reset the flow run, no?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -101,6 +108,33 @@ class DescopeWc extends BaseDescopeWc {
...
+  // single one covers everything.
+  //
+  // #flowPosition is read from the flow state, so *any* move invalidates a
+  // ceremony without a code path having to remember: an SDK response, a
+  // popstate from the back button, an attribute-driven restart.
+  //
+  // #flowGeneration covers the two moments where no state change happens but
+  // the ceremony is still dead: the user submitting something else (which is
+  // before its response arrives), and the component being disconnected.
+  #flowGeneration = 0;
```
nirgur: This name is too generic and does not give a clue what it affects on

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1060,11 +1117,53 @@ class DescopeWc extends BaseDescopeWc {
...
       try {
-        response =
+        const ceremony =
           action === RESPONSE_ACTIONS.webauthnCreate
-            ? await this.sdk.webauthn.helpers.create(webauthnOptions)
-            : await this.sdk.webauthn.helpers.get(webauthnOptions);
+            ? this.sdk.webauthn.helpers.create(webauthnOptions, abortController)
+            : this.sdk.webauthn.helpers.get(webauthnOptions, abortController);
+
+        const outcome = await timeoutPromise(
```
nirgur: Name is too generic

nirgur:

packages/sdks/web-component/src/lib/constants/index.ts
```diff
@@ -63,6 +63,14 @@ export const ELEMENTS_TO_IGNORE_ENTER_KEY_ON = [
 
 export const SDK_SCRIPTS_LOAD_TIMEOUT = 5000;
 
+// Upper bound on a passkey ceremony. Password manager extensions can leave
+// navigator.credentials.get/create pending for minutes after their prompt is
+// dismissed (Bitwarden waits up to 5 minutes for a window focus event before
+// falling back), and there is no signal we can observe. Without this the flow
+// waits forever. Kept long so a slow but genuine ceremony is not cut off - the
+// user's fast way out is the rest of the screen staying usable.
```
tomerlichtash: Comment might be too elaborate for this line, which explains itself, IMO.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -79,6 +80,12 @@ import {
 } from '../types';
 import BaseDescopeWc from './BaseDescopeWc';
 
+// Marks a passkey ceremony that ran out of time. timeoutPromise can only reject
+// with a plain Error, which is exactly what some password manager extensions
+// throw, so we take its resolve-with-a-fallback form instead. A real ceremony
+// resolves with encoded JSON, so this value cannot collide with one.
```
tomerlichtash: Same.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1060,11 +1067,36 @@ class DescopeWc extends BaseDescopeWc {
...
+            `WebAuthn operation timed out after ${WEBAUTHN_TIMEOUT}ms`,
+          );
+          // Report it as an abort, which is what it is from the flow's side and
+          // a value the SDK already produces (see identifyWebauthnError).
+          failure = 'AbortError';
+          failureReason = 'aborted';
+          failureMessage = 'Passkey operation timed out';
+        } else {
+          response = outcome;
+        }
```
tomerlichtash: Perhaps consider extracting this to a fn?

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  });
+
+  it('reports a timeout for the create (sign-up) ceremony too', async () => {
+    await renderPasskeyScreen();
+    respondWithCeremony('webauthnCreate');
+    nextMock.mockReturnValueOnce(generateSdkResponse({ screenId: '1' }));
+    sdk.webauthn.helpers.create.mockReturnValue(new Promise(() => {}));
+
+    await startCeremony();
+    await jest.advanceTimersByTimeAsync(WEBAUTHN_TIMEOUT + 1000);
```
tomerlichtash: Why +1000?
Instead of magic number consider extrdacting it to a const, or even create a dedicated const for the computed value which repeats in more tests.

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    nextMock.mockReturnValueOnce(generateSdkResponse({ screenId: '1' }));
+    sdk.webauthn.helpers.get.mockReturnValue(new Promise(() => {}));
+
+    await startCeremony();
+    await jest.advanceTimersByTimeAsync(WEBAUTHN_TIMEOUT + 1000);
+
+    await waitFor(() => expect(nextMock).toHaveBeenCalledTimes(2), {
+      timeout: WAIT_TIMEOUT,
+    });
+    expect(nextMock.mock.calls[1][5]).toEqual(
```
tomerlichtash: While pattern of `[x][y]` is common, the positions 1 and 5 give the impression that this generated mock may change in the future and break the test. Can we at least extract this to a getter, or think of a way to ensure this is safe?

packages/sdks/web-component/e2e/webauthnTimeout.spec.ts
```diff
@@ -0,0 +1,158 @@
+import { expect } from '@playwright/test';
```
nirgur: Do we really need this e2e file? The unit test file already covers the timeout, the create flow and the late answer. This one is bigger than our whole existing e2e suite and needs a navigator.credentials shim for WebKit.

packages/sdks/web-js-sdk/src/sdk/webauthn.ts
```diff
@@ -130,8 +130,14 @@ const createWebAuthn = (sdk: CoreSdk) => ({
 
 // Helpers functions
 
-async function create(options: string): Promise<string> {
+async function create(
+  options: string,
+  abort?: AbortController,
+): Promise<string> {
   const createOptions = decodeCreateOptions(options);
+  if (abort) {
```
nirgur, line 138: Do we need the abort param here? The PR says extensions ignore it and the timer is what actually recovers, so we are adding a public API param to the SDK for best effort cleanup. WDYT?

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
```diff
@@ -1060,11 +1067,36 @@ class DescopeWc extends BaseDescopeWc {
...
+          // We gave up, the browser did not. Asking it to stop is honoured by
+          // some extensions and ignored by others, so it is cleanup rather than
+          // the mechanism.
+          abortController.abort();
+          this.loggerWrapper.warn(
+            `WebAuthn operation timed out after ${WEBAUTHN_TIMEOUT}ms`,
+          );
+          // Report it as an abort, which is what it is from the flow's side and
+          // a value the SDK already produces (see identifyWebauthnError).
+          failure = 'AbortError';
```
nirgur: What if the run already expired on the server after 90s? Do we still get the cancelled screen, or a generic error?

packages/sdks/web-component/src/lib/constants/index.ts
nirgur: This last line is not true anymore. The screen staying usable was removed from the PR, so there is no fast way out now.

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
```diff
@@ -0,0 +1,205 @@
...
+describe('webauthn ceremony timeout', () => {
+  beforeEach(() => {
+    setupWebComponentTestEnv();
+  });
+
+  afterEach(() => {
+    teardownWebComponentTestEnv();
+  });
+
+  it('reports a timeout for the create (sign-up) ceremony too', async () => {
```
nirgur: This is the first test in the file, so `too` reads odd. Can we move it after the get one?

packages/sdks/web-component/src/lib/constants/index.ts
nirgur: Cut to 3 lines in 35af7ade. Good catch beyond the style point: the last sentence still said the fast way out is the rest of the screen staying usable, which was the screen-unlock part we removed. It was stale.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Trimmed to 2 lines in 35af7ade. Kept the reason we resolve with a sentinel instead of rejecting, since that is the part that is not obvious: timeoutPromise rejects with a plain Error, which is exactly what some extensions throw, so the two would be indistinguishable.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Extracted to #runWebauthnCeremony in 35af7ade. It returns what flow.next needs instead of assigning four outer variables. One note: the payload now spreads only the keys the outcome has, so on success there is no failure key. The request on the wire is the same because JSON.stringify drops undefined either way.

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
nirgur: Named in 35af7ade: PAST_WEBAUTHN_TIMEOUT for the 4 places that jump past the budget, and WITHIN_WEBAUTHN_TIMEOUT for the one that stays inside it.

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
nirgur: Added passkeyReplyPayload() in 35af7ade, with PASSKEY_REPLY_CALL and NEXT_PAYLOAD_ARG named next to it. All 5 call sites go through it, so the positions live in one place now.

packages/sdks/web-component/e2e/webauthnTimeout.spec.ts
nirgur: Agreed, deleted in 3a9c26f5. It was 158 lines against 77 for the whole existing e2e suite, and after the screen-unlock removal only one case was left. The jest suite already covers the timeout, the create flow and the late answer, so all the spec added was a WebKit navigator.credentials shim to maintain.

packages/sdks/web-component/src/lib/constants/index.ts
nirgur: Already cut in 35af7ade, which landed just after you wrote this. Tomer raised the same thing. The comment is 3 lines now and the stale sentence is gone.

packages/sdks/web-component/test/descope-wc.webauthnTimeout.test.ts
nirgur: Moved in 3a9c26f5, the get test comes first now.

packages/sdks/web-js-sdk/src/sdk/webauthn.ts
nirgur: I would keep it, but my reason has changed after looking closer.

Abort is not just cleanup. It does nothing for the Bitwarden case, since by 90s that call is already inside waitForFocus, which ignores the signal. But for a user on the native browser dialog or a hardware key, without abort we leave a live passkey dialog open for a flow that has already moved on, and completing it does nothing. AbortSignal closes it.

On the public API point: conditional(options, abort) in the same file already takes exactly this parameter, so we are matching an existing signature rather than inventing one, and both new params are optional.

Happy to drop it if you prefer a single-package PR, it would just cost that dialog cleanup.

packages/sdks/web-component/src/lib/descope-wc/DescopeWc.ts
nirgur: Checked this in the backend rather than guessing. We still get the cancelled screen, and the run is nowhere near expiry.

The timeout reply never touches the webauthn transaction. RunFinishSignIn short-circuits on input.Failure != "" and returns before any onetimeservice call, so there is no Redis lookup and no TTL dependency. Same in the sign-up, sign-up-or-in and add-device handlers. The result is HTTP 200, the flow walks its Failure branch, or the last screen is re-rendered with the error text.

The numbers around it:
- flow run TTL is FLOW_RUN_TIMEOUT_MINUTES, default 180 minutes, refreshed on every step
- webauthn transaction TTL is WEBAUTHN_TRANSACTION_EXPIRATION, default 120 seconds

Two things worth knowing that came out of this. The server already sends publicKey.timeout = 120s to the browser, the same constant, so the browser is told to give up at 120s and Bitwarden ignores it. Our 90s is the client side backstop for that. And 90 under 120 matters for a reason I had not thought of: at 150s a genuine late response would land on an expired transaction, at 90s it always lands on a live one.

Also: failureReason is logging only, nothing branches on it.

tomerlichtash (approved):

PR #1492, opened 2026-09-17
fix(web-component): build SAML IdP form with DOM APIs instead of innerHTML RELEASE

tomerlichtash:
Fixes descope/etc#18632

## Problem

`injectSamlIdpForm` built the SAML auto-submit form by interpolating `samlResponse` and `relayState` into `formEle.innerHTML`:

```ts
formEle.innerHTML = `
<input type="hidden" role="saml-relay-state" name="RelayState" value="${relayState}" />
...`;
```

`RelayState` is attacker-controllable — it originates from whoever initiates SSO and the backend correctly returns it unchanged, as SAML requires. A double quote in the value closes the `value` attribute, and the remainder is parsed as markup in the origin hosting `<descope-wc>`. With `">` the relay-state input ends up empty and a real `img` element is appended to the form, so an inline handler runs in the authentication origin.

Secondary issue on the same line: an injected control named `submit` clobbers `form.submit`, so `submitForm` throws and the SAML finish never completes.

## Fix

Create the inputs with `document.createElement` and assign `.value` as a property. Property assignment is not a parsing context, so nothing is reinterpreted as HTML and the values are preserved byte for byte — required for `RelayState` round-tripping and `SAMLResponse` signature validation at the SP. Deliberately no escaping or sanitization: it would corrupt values that legitimately contain `&` or `"`.

This mirrors `injectWsFedIdpForm`, which already used DOM APIs. Both helpers now share one module-level `createHiddenInput`; the WS-Fed call sites are unchanged.

## Test

Added a regression test with markup-shaped `RelayState` and `SAMLResponse` that asserts no foreign DOM nodes are created, the full values survive, and `form.submit` is still a function. Verified it fails on the old code:

```
expect(received).toBeNull()
Received: <img onerror="alert(1)" src="x" />
```

SAML, WS-Fed, and misc suites pass with the fix.

## Not in this PR

Definition-of-Done items that are not code in this repo: publishing the patched bundle, updating the advisory's affected range, and the CSP hardening review.


omercnet (approved):

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -759,11 +772,20 @@ export const injectSamlIdpForm = (
...
-  `;
+
+  formEle.appendChild(
+    createHiddenInput('SAMLResponse', samlResponse, 'saml-response'),
+  );
+  formEle.appendChild(
+    createHiddenInput('RelayState', relayState, 'saml-relay-state'),
+  );
+
+  const submitBtn = document.createElement('input');
```
nirgur, line 783: The submit button here is the same as in `injectWsFedIdpForm`, only the id differs. Now that `createHiddenInput` is shared, can we merge the two into one `injectIdpForm(url, fields, submitBtnId, cb)`? WDYT

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -750,6 +750,19 @@ export const showFirstScreenOnExecutionInit = (
   !thirdPartyAppStateId &&
   !applicationScopes;
 
+// Use DOM APIs to set values safely - these are server-provided values (raw XML,
```
nirgur, line 753: Can we mention HTML injection in the comment and not only values breaking the markup? `relayState` is attacker controlled, and "server-provided" reads like it is trusted. Without it someone might move this back to a template string later.

packages/sdks/web-component/test/descope-wc.saml.test.ts
```diff
@@ -93,6 +93,64 @@ describe('web-component', () => {
       );
     });
 
+    it('should not parse markup-shaped saml idp values as HTML', async () => {
```
nirgur, line 96: This is almost a copy of the test above. Can we extract the setup into a small helper so only the assertions differ?

nirgur (approved):
Left a few questions below, nothing blocking.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Fixed.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Done

packages/sdks/web-component/test/descope-wc.saml.test.ts
tomerlichtash: Done

PR #1493, opened 2026-09-17
refactor(web-component): address review on SAML IdP form fix

tomerlichtash:
Follow-up to #1492. Non-blocking review comments from @nirgur, split out so they did not hold up the security fix.

### Comment: comment wording

`relayState` is attacker-controlled, and "server-provided" read like it was trusted. Reworded to name HTML injection and the no-escaping rule directly, so nobody moves this back to a template string later:

```ts
// DOM APIs only, never innerHTML: these values are attacker-controllable
// (SAML RelayState), so interpolating them into markup is HTML injection.
// Do not escape either - they must round-trip byte for byte.
```

### Comment: merge the two injectors

Extracted `createSubmitButton(id)`, which removes the duplicated block between `injectSamlIdpForm` and `injectWsFedIdpForm` (six identical lines, only the id differed).

Stopped short of the full `injectIdpForm(url, fields, submitBtnId, cb)` merge. With `createHiddenInput` and `createSubmitButton` both shared, each injector is eight lines of straight-line code with no shared logic left, so a generic would add a `{name, value, role?}[]` field type and an indirection layer for no gain. Worth revisiting if a third IdP form appears. Happy to do the merge instead if you disagree.

### Comment: test duplication

Extracted `renderSamlIdpForm(samlResponse, relayState)` returning `{ form, mockSubmitForm }`. Only the assertions differ between the two tests now, and the file is 13 lines shorter.

No behavior change. SAML, WS-Fed, and misc suites pass (21 tests).


omercnet (dismissed):

packages/sdks/web-component/src/lib/helpers/helpers.ts
```diff
@@ -763,6 +764,15 @@ const createHiddenInput = (name: string, value: string, role?: string) => {
   return input;
 };
 
+const createSubmitButton = (id: string) => {
```
nirgur: Do we need the hidden submit button at all? We always call `form.submit()`, and nothing in the repo uses `SAMLSubmitButton` or `WSFedSubmitButton`.

packages/sdks/web-component/test/descope-wc.saml.test.ts
```diff
@@ -67,6 +71,15 @@ describe('web-component', () => {
...
+      return { form, mockSubmitForm };
+    };
+
+    it('should validate handling of saml idp response', async () => {
+      const { form, mockSubmitForm } = await renderSamlIdpForm(
+        'saml-response-dummy-value',
+        'saml-relay-state-dummy-value',
+      );
+
       expect(form).toBeInTheDocument();
```
nirgur, line 83: Now that we get `form` back, can the asserts below use `form.querySelector` like the second test does?

nirgur (dismissed):
About the injector merge, the reasoning makes sense. Let's keep the two as they are and revisit if a third IdP form shows up. Two small questions below.

packages/sdks/web-component/src/lib/helpers/helpers.ts
tomerlichtash: Removed.

packages/sdks/web-component/test/descope-wc.saml.test.ts
tomerlichtash: Changed.

